PR #107, opened 2023-04-17
add dynamic wait until gateway is ready

mohamedamer453:
### Description

Added a dynamic wait for the gateway test to keep checking for the gateway until it's ready and handle failures in a better way.

> Workflow runs might still fail due to an issue discovered with gateway nodes on devnet https://github.com/threefoldtech/tf_operations/issues/1598

### DIFF
--- a/packages/grid_client/tests/modules/gateways.test.ts
+++ b/packages/grid_client/tests/modules/gateways.test.ts
@@ -194,28 +194,28 @@ test("TC1237 - Gateways: Expose a VM Over Gateway", async () => {
   const domain = "https://" + gatewayResult[0].domain;
   let reachable = false;
 
-  for (let i = 0; i < 15; i++) {
+  for (let i = 0; i < 30; i++) {
     fetch(domain, { mode: "no-cors" })
-      .then(r => {
+      .then(res => {
         log("gateway is reachable");
         reachable = true;
-        i = 15;
       })
-      .catch(e => {
+      .catch(err => {
         log("gateway is not reachable");
       });
-    const wait = await setTimeout(10000, "Waiting for gateway to be ready");
+    if (reachable) {
+      break;
+    }
+    const wait = await setTimeout(5000, "Waiting for gateway to be ready");
     log(wait);
   }
 
   if (reachable) {
-    //Verify that the domain points to VM and lists the directories available in the VM
-    exec("curl " + domain, function (_, stdout) {
-      log(stdout);
-      expect(stdout).toContain("Directory listing for /");
-      expect(stdout).toContain("bin/");
-      expect(stdout).toContain("dev/");
-      expect(stdout).toContain("etc/");
+    fetch(domain, { mode: "no-cors" }).then(res => {
+      log(res.status);
+      log(res.statusText);
+      expect(res.status).toBe(200);
+      expect(res.statusText).toBe("OK");
     });
   } else {
     throw new Error("Gateway is unreachable after multiple retries");
--- a/packages/grid_client/tests/modules/gateways.test.ts
+++ b/packages/grid_client/tests/modules/gateways.test.ts
@@ -1,3 +1,5 @@
+import { setTimeout } from "timers/promises";
+
 import { FilterOptions, GatewayNameModel, generateString, GridClient, MachinesModel, randomChoice } from "../../src";
 import { config, getClient } from "../client_loader";
 import { generateInt, log, splitIP } from "../utils";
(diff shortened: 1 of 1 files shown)


packages/grid_client/tests/modules/gateways.test.ts
```diff
@@ -190,15 +192,34 @@ test("TC1237 - Gateways: Expose a VM Over Gateway", async () => {
...
 
   const domain = "https://" + gatewayResult[0].domain;
+  let reachable = false;
+
+  for (let i = 0; i < 15; i++) {
+    fetch(domain, { mode: "no-cors" })
+      .then(r => {
+        log("gateway is reachable");
+        reachable = true;
+        i = 15;
```
AhmedHanafy725, line 202: why setting `i = 15`? Is it to break the loop?

packages/grid_client/tests/modules/gateways.test.ts
```diff
@@ -190,15 +192,34 @@ test("TC1237 - Gateways: Expose a VM Over Gateway", async () => {
...
-  exec("curl " + domain, function (_, stdout) {
-    log(stdout);
-    expect(stdout).toContain("Directory listing for /");
-    expect(stdout).toContain("bin/");
-    expect(stdout).toContain("dev/");
-    expect(stdout).toContain("etc/");
-  });
+  if (reachable) {
+    //Verify that the domain points to VM and lists the directories available in the VM
+    exec("curl " + domain, function (_, stdout) {
```
AhmedHanafy725, line 213: why using `exec` to `curl` http request? can't we use `fetch` or `axios` to do it?

packages/grid_client/tests/modules/gateways.test.ts
```diff
@@ -190,15 +192,34 @@ test("TC1237 - Gateways: Expose a VM Over Gateway", async () => {
...
+    fetch(domain, { mode: "no-cors" })
+      .then(r => {
+        log("gateway is reachable");
+        reachable = true;
+        i = 15;
+      })
+      .catch(e => {
+        log("gateway is not reachable");
+      });
+    const wait = await setTimeout(10000, "Waiting for gateway to be ready");
```
AhmedHanafy725, line 207: I think 10 sec is too much to wait for each iteration

AhmedHanafy725 (changes requested):

packages/grid_client/tests/modules/gateways.test.ts
mohamedamer453: i think `fetch` also works but i encountered a case were `fetch` returned a false positive.
The case happened with node 11 as sometimes `fetch` will return that it's reachable but the gateway will still give `404 not found` for a couple of minuets so the extra `curl` command confirms that the gateway is actually working and asserts that the desired result appeared.

packages/grid_client/tests/modules/gateways.test.ts
mohamedamer453: yes it's for breaking the loop. but i'll check if there's a cleaner way to do it

packages/grid_client/tests/modules/gateways.test.ts
mohamedamer453: sure will decrease it to 5 seconds and increase the iterations to 30

packages/grid_client/tests/modules/gateways.test.ts
AhmedHanafy725: you can use `break`

mohamedamer453: done in https://github.com/threefoldtech/tfgrid-sdk-ts/pull/107/commits/de767e12e94a87712fd099f88828ca329928ff9d i was unable to use `break` inside the fetch method so that what made me use the `i=15` at first but i found a better way of using `break` but outside the fetch method with a condition using the `reachable` variable.

packages/grid_client/tests/modules/gateways.test.ts
mohamedamer453: done in https://github.com/threefoldtech/tfgrid-sdk-ts/pull/107/commits/de767e12e94a87712fd099f88828ca329928ff9d instead of checking with `curl` now the check is done with `fetch` and the assertion is done for the status and status code.

packages/grid_client/tests/modules/gateways.test.ts
mohamedamer453: done in https://github.com/threefoldtech/tfgrid-sdk-ts/pull/107/commits/de767e12e94a87712fd099f88828ca329928ff9d Now it's 30 iterations for 5 sec each.

packages/grid_client/tests/modules/gateways.test.ts
AhmedHanafy725: we can still check the content of the response

mohamedamer453: Re-added in https://github.com/threefoldtech/tfgrid-sdk-ts/pull/107/commits/67c62b4a0910f8c9e12f83a048e13bac6b637f79 using fetch

AhmedHanafy725 (approved):

Issue #77, opened 2023-04-06
UI Taiga and Owncloud Email Address label need to be changed


A-Harby:
### Description

In Taiga and Owncloud, What `From Email Address` suppose to mean?

![image](https://user-images.githubusercontent.com/109473873/230379482-4610c1f1-d4f4-4971-b391-908134f37a0a.png)

mohamedamer453:
Verified on devnet.

The email address label was updated for the two solutions.

![image](https://user-images.githubusercontent.com/101194226/233073184-2b1d8d32-f485-4ae7-9df3-82548c5b6145.png)

![image](https://user-images.githubusercontent.com/101194226/233073269-571c2a4e-ef68-467c-b9b2-b976091d9a07.png)

Weblets Version: `ad87c97`

Issue #75, opened 2023-04-06
QR code of newly created twin isn't displayed.


A-Harby:
### Description

Devnet, 01de9dc.
QR code of newly created twin isn't displayed.

### Steps to reproduce

Just click on `Create account` buuton. 

### Logs/Alerts

![image](https://user-images.githubusercontent.com/109473873/230385587-aaa80db9-b1e1-4bc6-b56d-1e6ad2d3217d.png)

mohamedamer453:
Verified on devnet.

QR code now appears after creating a new twin.

[Screencast from 19-04-23 14:23:35.webm](https://user-images.githubusercontent.com/101194226/233074174-be2bd2bf-0d4f-49e4-9084-00d45f72ed66.webm)

Weblets Version: `ad87c97`

Issue #72, opened 2023-04-11
searching for node in presearch does not keeps searching


khaledyoussef24:
### Description

while searching for node in presearch when it does not find nodes  it does not time out

### Steps to reproduce

1. go to playground
2. go to presearch tab
3. use capacity filter to find a node
4. it will not time out
![Screenshot from 2023-04-11 10-38-12](https://user-images.githubusercontent.com/110984055/231109377-84eb6220-c8da-42b4-a1f9-6ccdbdcf5da1.png)

mohamedamer453:
Verified on devnet.

Now filtering for nodes will timeout when there are no available nodes to deploy on. and it returns that there are `0` nodes found.

[Screencast from 19-04-23 14:28:58.webm](https://user-images.githubusercontent.com/101194226/233075303-65695409-63b6-4f3c-960d-d17467bbcd79.webm)

Weblets Version: `ad87c97`

PR #109, opened 2023-04-18
bug fixes for playground solutions and profile manager 

alichaddad:
### Description

Fixes small bug issues in playground

### Changes

- Change `From email` in mail setup to `Support email`
- Fix capacity filter in presearch
- Fix showing QR code after creating new account in profile manager

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/75
- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/77
- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/72

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/packages/weblets/src/elements/presearch/Presearch.wc.svelte
+++ b/packages/weblets/src/elements/presearch/Presearch.wc.svelte
@@ -166,7 +166,7 @@
           {profile}
           on:fetch={({ detail }) => (data.selection.nodes = detail)}
           nodes={data.selection.nodes}
-          exclusiveFor="research"
+          exclusiveFor="presearch"
         />
       </section>
       <section style={display(active, "restore")}>
--- a/packages/weblets/src/elements/presearch/Presearch.wc.svelte
+++ b/packages/weblets/src/elements/presearch/Presearch.wc.svelte
@@ -166,7 +166,7 @@
           {profile}
           on:fetch={({ detail }) => (data.selection.nodes = detail)}
           nodes={data.selection.nodes}
-          exclusiveFor="presearch"
+          exclusiveFor="research"
         />
       </section>
       <section style={display(active, "restore")}>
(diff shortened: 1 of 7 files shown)


packages/weblets/src/elements/profiles/Profiles.wc.svelte
```diff
@@ -74,7 +74,7 @@
     try {
       const relay = grid.getDefaultUrls(window.env.NETWORK).relay.slice(6);
       const account = await grid.tfchain.createAccount(relay, null);
-      mnemonics.setValue(account.mnemonic, { error: "Please fund your wallet then refresh." });
+      mnemonics.setValue(account.mnemonic);
```
AhmedHanafy725, line 77: I think we still need the error message somehow instead of saying that the balance is not enough. maybe we can remove it from here and update the other error message.

AhmedHanafy725 (changes requested):

packages/weblets/src/elements/profiles/Profiles.wc.svelte
alichaddad: Why is this case different than the other one? In the case of new account or using an existing, both will have to fund their accounts, right?

AhmedHanafy725: yes that's what I mean

packages/weblets/src/elements/presearch/Presearch.wc.svelte
```diff
@@ -166,7 +166,7 @@
           {profile}
           on:fetch={({ detail }) => (data.selection.nodes = detail)}
           nodes={data.selection.nodes}
-          exclusiveFor="research"
+          exclusiveFor="presearch"
```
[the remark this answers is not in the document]
alichaddad: I see but for whatever reason `research` doesn't work. It causes the filter request to hangs without any timeout. Is that something that should be fixed from the server?

AhmedHanafy725 (approved):

PR #207, opened 2023-05-03
Add client

AhmedHanafy725:


packages/tfchain_client/src/client.ts
```diff
@@ -0,0 +1,244 @@
...
+        }
+        if (status.isInBlock) {
+          events.forEach(({ phase, event: { data, method, section } }) => {
+            console.log(`phase: ${phase}, section: ${section}, method: ${method}, data: ${data}`);
+            if (section === "system" && method === "ExtrinsicFailed") {
+              const errorIndex = parseInt(data.toJSON()[0].module.error.replace(/0+$/g, ""));
+              reject(errorIndex);
+            } else if (resultSections.includes(section)) {
+              resultData.push(data.toPrimitive()[0]);
+            } else if (section === "system" && method === "ExtrinsicSuccess") {
```
Mahmoud-Emad, line 190: The hard coded of the `ExtrinsicSuccess` here, can we put it inside an enum just like.

```suggestion
enum TFChainMethods {
  ExtrinsicFailed = 'ExtrinsicFailed',
  ExtrinsicSuccess = 'ExtrinsicSuccess',
}
} else if (section === "system" && method === TFChainMethods.ExtrinsicSuccess) {
```
if we changed it in the feature for any reason, we don't have to search for it in the whole project.
it's just a question, if you see something different please go ahead.

Mahmoud-Emad (approved):
LGTM, Good job ya @AhmedHanafy725

Issue #375, opened 2023-05-28
TFChain integration tests randomly fail with different errors


mohamedamer453:
### Description

The integrations tests for tfchain have been randomly failing for a while now and after some investigation these are the issues that i found.

I ran each test multiple times while debugging this issue but i wasn't able to detect a pattern or cause for these errors.

The only test that didn't produce any of these errors is `TC1261 - TFChain: Create Account` other than that the following errors appear randomly in the rest of the tests 

- I think this one is not an issue but it's reported as an error but it mentions that it's `Normal connection closure` from the chain and it doesn't cause the tests to fail.

```bash
  console.error
    2023-05-28 10:56:00          API-WS: Error: disconnected from wss://tfchain.dev.grid.tf/ws: 1000:: Normal connection closure
        at W3CWebSocket.value (/home/mohamed/Desktop/Threefold Repos/sdkts_temp/tfgrid-sdk-ts/node_modules/@polkadot/rpc-provider/cjs/ws/index.js:221:23)
        at W3CWebSocket._dispatchEvent [as dispatchEvent] (/home/mohamed/Desktop/Threefold Repos/sdkts_temp/tfgrid-sdk-ts/node_modules/yaeti/lib/EventTarget.js:107:17)
        at W3CWebSocket.onClose (/home/mohamed/Desktop/Threefold Repos/sdkts_temp/tfgrid-sdk-ts/node_modules/websocket/lib/W3CWebSocket.js:228:10)
        at WebSocketConnection.<anonymous> (/home/mohamed/Desktop/Threefold Repos/sdkts_temp/tfgrid-sdk-ts/node_modules/websocket/lib/W3CWebSocket.js:201:17)
        at WebSocketConnection.emit (node:events:513:28)
        at WebSocketConnection.Object.<anonymous>.WebSocketConnection.handleSocketClose (/home/mohamed/Desktop/Threefold Repos/sdkts_temp/tfgrid-sdk-ts/node_modules/websocket/lib/WebSocketConnection.js:389:14)
        at TLSSocket.emit (node:events:525:35)
        at node:net:322:12
        at TCP.done (node:_tls_wrap:588:7)

      at apply (../../node_modules/@polkadot/util/cjs/logger.js:62:22)
      at Object.error (../../node_modules/@polkadot/util/cjs/logger.js:118:14)
      at ../../node_modules/@polkadot/rpc-provider/cjs/ws/index.js:390:15
          at async Promise.all (index 0)
[...]

PR #379, opened 2023-05-29
Don't disconnect after each connection got opened on the tfhchain module

AhmedHanafy725:
### Description

- Don't disconnect after each connection got opened on the tfhchain module and do it once after the tests finish
- Use import for loading localstorage backend instead of require
- Use while for wait on account activation instead of backoff package

### Related Issues

- #375 
- https://github.com/threefoldtech/www-mastodon/pull/171

### Checklist

- [x] Tests included
- [x] Build pass
- [ ] Documentation

### DIFF
--- a/packages/grid_client/scripts/create_account.ts
+++ b/packages/grid_client/scripts/create_account.ts
@@ -1,3 +1,4 @@
+import { QueryClient } from "@threefold/tfchain_client";
 import { env } from "process";
 import { default as urlParser } from "url-parse";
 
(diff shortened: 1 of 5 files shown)


packages/grid_client/scripts/create_account.ts
```diff
@@ -1,3 +1,4 @@
+import { QueryClient } from "@threefold/tfchain_client";
```
zaelgohary, line 1: Can we remove this as it's not used?

zaelgohary (changes requested):

zaelgohary (approved):

PR #390, opened 2023-05-29
improve timeouts

mohamedamer453:
### Description

Improving timeouts in the grid client integration tests.

### Changes

- Added a timeout for the disconnect method, so that the test will not be stuck forever if it fails.
- Added a dynamic timeout for the k8s tests instead of having a static one, the timeout checks every 5 seconds if the master and worker/s are returned or not.
- Increased the default timeout of the k8s tests because sometimes the add worker test might exceed the current timeout while waiting for the added worker.

### DIFF
--- a/packages/grid_client/tests/modules/kubernetes.test.ts
+++ b/packages/grid_client/tests/modules/kubernetes.test.ts
@@ -1,5 +1,3 @@
-import { setTimeout } from "timers/promises";
-
 import {
   AddWorkerModel,
   DeleteWorkerModel,
@@ -10,7 +8,7 @@ import {
   randomChoice,
 } from "../../src";
 import { config, getClient } from "../client_loader";
-import { bytesToGB, generateInt, log, RemoteRun, splitIP } from "../utils";
+import { bytesToGB, generateInt, k8sWait, log, RemoteRun, splitIP } from "../utils";
 
 jest.setTimeout(500000);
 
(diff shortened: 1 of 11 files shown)


packages/grid_client/tests/modules/kubernetes.test.ts
```diff
@@ -217,13 +217,25 @@ test("TC1231 - Kubernetes: Deploy a Kubernetes Cluster", async () => {
...
+      if (res.includes(masterName.toLowerCase()) && res.includes(workerName.toLowerCase())) {
+        reachable = true;
+      }
+    });
+    if (reachable) {
+      break;
+    }
+    const wait = await setTimeout(5000, "Waiting for cluster to be ready");
+    log(wait);
+  }
```
AhmedHanafy725, line 237: can we make it a method and reuse it?

AhmedHanafy725 (changes requested):

packages/grid_client/tests/modules/kubernetes.test.ts
mohamedamer453: added in https://github.com/threefoldtech/tfgrid-sdk-ts/pull/390/commits/a1068c537c5469c02465d9e7c7ec6034d2154162

mohamedamer453: forgot to add the utils file to the last commit. added in https://github.com/threefoldtech/tfgrid-sdk-ts/pull/390/commits/3b5ca9871fe7817734eaee2a46c0da7c1861ab29

packages/grid_client/tests/utils.ts
```diff
@@ -54,4 +55,29 @@ async function returnRelay() {
   return relay;
 }
 
-export { log, generateHash, generateInt, splitIP, bytesToGB, RemoteRun, returnRelay };
+async function k8sWait(masterIP, masterName, workerName, waitTime, newWorkerName?) {
```
AhmedHanafy725, line 58: can we pick a better name instead of `masterIp`? cause it's an object from the ssh client

AhmedHanafy725 (changes requested):

packages/grid_client/tests/utils.ts
mohamedamer453: Names were updated in https://github.com/threefoldtech/tfgrid-sdk-ts/pull/390/commits/25e34483bdfa54279036a1d157ae4b50480f66b3

AhmedHanafy725 (approved):

Issue #419, opened 2023-05-31
Cannot read properties of undefined (reading 'query')  


Mahmoud-Emad:
### Description

After refreshing the nodes page there is an error appears in the console, while, if refreshed the main page there are no errors

### Steps to reproduce

- Go to [dashboard](https://dashboard.dev.grid.tf/explorer/nodes)
- open the console
- reload the page

### Logs/Alerts
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/57001890/630538e6-ac05-4b3e-af60-d445fa72aa47)

Mahmoud-Emad:
Should be fixed in https://github.com/threefoldtech/tfgrid-sdk-ts/pull/420

PR #420, opened 2023-05-31
Displaying the tft-price swap one second after mounting the page

Mahmoud-Emad:
### Description

After refreshing the nodes page there is an error appears in the console, while, if refreshed the main page there are no errors
implemented a small fix to wait on the API one second after mounting the page

### Related Issues
- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/419

### Checklist

- [ ] Tests included
- [ ] Build pass
- [x] Documentation

### DIFF
--- a/packages/dashboard/src/Dashboard.vue
+++ b/packages/dashboard/src/Dashboard.vue
@@ -284,11 +284,6 @@ export default class Dashboard extends Vue {
 
   async updated() {
     this.accounts = this.$store.state.portal.accounts;
-    if (this.$api && this.$route.path == "/") {
-      this.loadingAPI = false;
-    } else if (this.$route.path !== "/") {
-      this.loadingAPI = false;
-    }
   }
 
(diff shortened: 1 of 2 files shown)


xmonader (approved):

packages/dashboard/src/Dashboard.vue
```diff
@@ -284,11 +284,6 @@ export default class Dashboard extends Vue {
 
   async updated() {
     this.accounts = this.$store.state.portal.accounts;
-    if (this.$api && this.$route.path == "/") {
-      this.loadingAPI = false;
-    } else if (this.$route.path !== "/") {
-      this.loadingAPI = false;
-    }
```
Mahmoud-Emad, line 291: Removed this part because it doesn't make any sense, it sets the API Loading variable to false in all routes on the updated lifecycle, so the API read as undefined depends on its actual value of it, which it break the request

PR #442, opened 2023-05-31
Add support for config.js file

MohamedElmdary:


Mahmoud-Emad (approved):

packages/playground/src/clients/index.ts
```diff
@@ -1,7 +1,9 @@
 import GridProxyClient, { Network } from "tf_gridproxy_client";
 import TFGridGqlClient, { Networks } from "tfgrid-gql";
 
-const gqlClient = new TFGridGqlClient(process.env.NETWORK as Networks);
-const gridProxyClient = new GridProxyClient(process.env.NETWORK as Network);
+const network = process.env.NETWORK || window.env.NETWORK;
+
+const gqlClient = new TFGridGqlClient(network as Networks);
+const gridProxyClient = new GridProxyClient(network as Network);
```
AhmedHanafy725, line 7: these clients are better to work with links in case their services are deployed on different domains. or support it beside the `network` env

AhmedHanafy725 (approved):

Issue #320, opened 2023-05-14
Public SSH Key Validation is not working


mohamedamer453:
### Description

Public SSH Key Validation is not working and it can be updated with any value.

### Adding a new ssh key with an invalid value

[Untitled_ May 14, 2023 3_45 PM.webm](https://github.com/threefoldtech/playground/assets/101194226/7d72d7e3-9af7-41a5-bbaf-c572c0e106e9)

### Updating and already existing key with an invalid value.

[Untitled_ May 14, 2023 3_40 PM.webm](https://github.com/threefoldtech/playground/assets/101194226/06f5cbb4-67e2-4385-a672-fd78d3da8a3a)

0oM4R:
PR #455 ready for review

0oM4R:
after some discussions, will make sure the user provides a ssh key, if no the user will not be able to complete the deployment.
whoever the pr now is ready for  review :D

PR #455, opened 2023-06-01
add ssh validator 

0oM4R:
### Description

making ssh key required, if the user tried to deploy without providing ssh key will stop the deployment process and show an error to the user

### Changes

- make ssh-key env var mandatory can't be removed
    image

- expose a ssh validator and use it in the weblet 
    image

### Related Issues

- #320 
### Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation

### DIFF
--- a/packages/playground/src/weblets/profile_manager.vue
+++ b/packages/playground/src/weblets/profile_manager.vue
@@ -197,7 +197,7 @@
                     : SSHKeyHint
                 "
                 :persistent-hint="updatingSSH || generatingSSH || !!SSHKeyHint"
-                :rules="[validateSSH]"
+                :rules="[value => !!value || 'SSH key is required']"
               />
             </CopyInputWrapper>
           </template>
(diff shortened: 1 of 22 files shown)


packages/playground/src/weblets/profile_manager.vue
```diff
@@ -413,6 +414,14 @@ function validatePassword(value: string) {
...
   }
 }
+const SSH_REGEX =
+  /^(sk-)?(ssh-rsa AAAAB3NzaC1yc2|ecdsa-sha2-nistp256 AAAAE2VjZHNhLXNoYTItbmlzdHAyNT|ecdsa-sha2-nistp384 AAAAE2VjZHNhLXNoYTItbmlzdHAzODQAAAAIbmlzdHAzOD|ecdsa-sha2-nistp521 AAAAE2VjZHNhLXNoYTItbmlzdHA1MjEAAAAIbmlzdHA1Mj|ssh-ed25519 AAAAC3NzaC1lZDI1NTE5|ssh-dss AAAAB3NzaC1kc3)[0-9A-Za-z+/]+[=]{0,3}( .*)?$/;
+
+function validateSSH(value: string) {
+  if (value === "") return "SSH Value is required";
+  if (!SSH_REGEX.test(value)) return "Please enter a valid SSH key";
+  return true;
+}
```
Mahmoud-Emad, line 424: We should validate that the input doesn't contain any spaces or is not empty. However, we should not validate the SSH key format. The reason for this is that even if the SSH key format is incorrect, the deployment process should still continue. Since the SSH key is an environment variable, it is the responsibility of the user to enter the correct SSH key in order to access their machine. Validating that the input doesn't contain spaces or is not empty helps ensure the integrity and accuracy of the provided data without imposing unnecessary restrictions on the SSH key format.

Mahmoud-Emad: If necessary, I would prefer to use the `sshpk` package for several reasons:

1. The `sshpk` package provides a more comprehensive and reliable solution compared to using a regular expression. By utilizing the package, we can be confident that it covers all SSH key types and is regularly updated to include any new types that may arise. This saves us from the manual effort of updating the regular expression to match new types.

2. The `sshpk` package has a small footprint, meaning it doesn't introduce significant overhead or increase the overall size of our project unnecessarily.

3. With the `sshpk` package, we can leverage its extensive coverage of SSH key types, ensuring that all types are supported without any gaps.

Considering these factors, I believe using the `sshpk` package would be a more robust and future-proof approach for validating SSH keys. It provides better maintenance, reliability, and comprehensive support for various key types.

what do you think?

0oM4R: As I discussed with @AhmedHanafy725 ; will just trim the entred key for now

Mahmoud-Emad (approved):
LGTM, Good job :)

Issue #393, opened 2023-05-29
Update explorer links to direct to the correct network.


A-Harby:
### Description

`Explore ThreeFold Grid Capacity` button directs to the mainnet explorer not to the devnet. 
Also, it should reflect the chosen network.

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/905bf354-23e2-4827-912e-dd7bb762de63)

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/e01a4302-df51-4729-9513-0476639e81f9)

mohamedamer453:
Verified on devnet.

The explorer link now redirects to the correct network.

[Screencast from 12-06-23 12:50:42.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/101194226/f578fe07-a53c-45d5-bb9f-3ab5d11627b8)

PR #472, opened 2023-06-04
Fix capacity url

zaelgohary:
### Description

Fix capacity url

### Changes

Update capacity url to be dynamic depending on the network

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/393

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/packages/playground/src/components/connect_wallet_landing.vue
+++ b/packages/playground/src/components/connect_wallet_landing.vue
@@ -90,6 +90,7 @@
 </template>
 
 <script lang="ts">
+import { NetworkEnv } from "@threefold/grid_client";
 import { useTheme } from "vuetify";
 
 export default {
@@ -99,8 +100,8 @@ export default {
     const baseUrl = import.meta.env.BASE_URL;
     const network = process.env.NETWORK || (window as any).env.NETWORK;
 
-    const getCapacityURL = (network: string) => {
-      if (network === "main") {
+    const getCapacityURL = (network: NetworkEnv) => {
+      if (network === NetworkEnv.main) {
         return "https://dashboard.grid.tf/explorer/statistics";
       }
       return `https://dashboard.${network}.grid.tf/explorer/statistics`;
--- a/packages/playground/src/components/connect_wallet_landing.vue
+++ b/packages/playground/src/components/connect_wallet_landing.vue
@@ -64,9 +64,7 @@
         </div>
 
         <div class="d-flex justify-center mt-16">
-          <v-btn color="primary" target="_blank" href="https://dashboard.grid.tf/explorer/statistics">
-            Explore ThreeFold Grid Capacity
-          </v-btn>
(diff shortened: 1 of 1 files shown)


packages/playground/src/components/connect_wallet_landing.vue
```diff
@@ -99,6 +97,19 @@ export default {
   setup() {
     const theme = useTheme();
     const baseUrl = import.meta.env.BASE_URL;
+    const network = process.env.NETWORK || (window as any).env.NETWORK;
+
+    const getCapacityHref = () => {
```
xmonader, line 102: rename to getCapacityURL(network), your function shouldn't depend on enclosed variables e.g the network const, but make it as a param

packages/playground/src/components/connect_wallet_landing.vue
```diff
@@ -99,6 +97,19 @@ export default {
...
+      let href = `https://dashboard.${network}.grid.tf/explorer/statistics`;
+      if (network === "main") {
+        href = "https://dashboard.grid.tf/explorer/statistics";
```
xmonader, line 105: extract to a const `mainnetCapacityURL`

packages/playground/src/components/connect_wallet_landing.vue
xmonader: Better to move into its own file, so it would be easily testable too

packages/playground/src/components/connect_wallet_landing.vue
```diff
@@ -99,6 +97,19 @@ export default {
...
+        return href;
+      }
+
+      return href;
+    };
+
+    const capacityURL = getCapacityHref();
```
Mahmoud-Emad: what if we did something like that.

```suggestion

const getCapacityHref = (network: string) => {
  if (network === "main"){
    network = ""
  }
  const = statisticLink = `https://dashboard.${network}.grid.tf/explorer/statistics`;
  return statisticLink;
};

const capacityURL = getCapacityHref();
```

in this example we change the value of the network if the network is equal to "main", then return the URL, also we have an enum called NETWORK maybe we can use it instead of a hard-coded string.

what do you think?

zaelgohary: URL will be corrupted as in mainnet case will be "https://dashboard..grid.tf/explorer/statistics"

xmonader: It will have two dots given that network is empty

Mahmoud-Emad: indeed, then we can correct it to set the default value of the statisticLink to be the one for the main net, then return it.

Mahmoud-Emad (approved):

xmonader (approved):

Mahmoud-Emad:
Good job 👏

PR #473, opened 2023-06-04
Move capacity url fn to a separate file

zaelgohary:
### Description

- Move capacity url fn to a separate file

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/393

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/packages/playground/src/utils/getCapacityUrl.ts
+++ b/packages/playground/src/utils/getCapacityUrl.ts
@@ -1,0 +1,14 @@
+import { NetworkEnv } from "@threefold/grid_client";
+
+const network = process.env.NETWORK || (window as any).env.NETWORK;
+
+const getCapacityURL = (network: NetworkEnv) => {
+  if (network === NetworkEnv.main) {
+    return "https://dashboard.grid.tf/explorer/statistics";
+  }
+  return `https://dashboard.${network}.grid.tf/explorer/statistics`;
+};
+
+const capacityURL = getCapacityURL(network);
+
+export default capacityURL;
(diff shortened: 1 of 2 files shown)


packages/playground/src/utils/getCapacityUrl.ts
```diff
@@ -0,0 +1,14 @@
+import { NetworkEnv } from "@threefold/grid_client";
+
+const network = process.env.NETWORK || (window as any).env.NETWORK;
```
xmonader, line 3: this line isn't needed, it can be passed from the calling function, the file should be only the getCapacityURL function parameterized in the a network var, the usage of that function should be in the right context

packages/playground/src/utils/getCapacityUrl.ts
```diff
@@ -0,0 +1,14 @@
...
+
+const getCapacityURL = (network: NetworkEnv) => {
+  if (network === NetworkEnv.main) {
+    return "https://dashboard.grid.tf/explorer/statistics";
+  }
+  return `https://dashboard.${network}.grid.tf/explorer/statistics`;
+};
+
+const capacityURL = getCapacityURL(network);
```
xmonader, line 12: Shouldn't be called here

packages/playground/src/utils/getCapacityUrl.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export default capacityURL;
```
xmonader, line 14: the exported should be the getCapacityURL function, this variable capacityURL shouldn't exist, it needs to  only exist at the caller frame

Mahmoud-Emad (approved):

xmonader:
Awesome!

Issue #12, opened 2023-04-02
Add method to calculate amount needed to unlock contract from grace


DylanVerstraete:
Currently, it's not possible to see how much tokens are required to "unlock" a graced out contract. This info is will be added on tfchain in 2.4.0. 

A client can get the the lock attached to a contract by querying: `smartContractModule.contractLock(contract_id)`

This will return an object like: 

```
{
  amountLocked: 123123123123
  lockUpdated: 1,677,534,012
  cycles: 12
}
``` 

`amountLocked` specifies the outstanding amount "due" for the contract. We can expose this function to add this in the weblets so a user can know how much he needs to recharge his account with if he wants to unlock his contract.

mohamedamer453:
Verified on devnet.

The method was added and it's currently used in the playground to display the amount of tfts needed to unlock a certain contract.

[Screencast from 12-06-23 14:18:29.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/101194226/df45e580-819a-4019-ab34-c6456fcece44)

PR #479, opened 2023-06-05
Implemented the locked contract feature, added inside the playground.

Mahmoud-Emad:
### Description

The locked contract amount feature has been implemented within the chain client. It has been called and exposed in the grid client, allowing it to be used in the playground. When the dialog is opened, the grid client is requested to retrieve the locked contract amount and display it to the end user. This enhancement enables users to easily view and interact with the locked contract amount information within the playground interface.

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/12

### Screenshots
- image

on click

- image
- image

### Checklist

- [ ] Tests included
- [x] Build pass
- [x] Documentation

### DIFF
--- a/packages/grid_client/src/modules/contracts.ts
+++ b/packages/grid_client/src/modules/contracts.ts
@@ -13,6 +13,7 @@ import {
   ContractConsumption,
   ContractGetByNodeIdAndHashModel,
   ContractGetModel,
+  ContractLockModel,
   ContractsByAddress,
   ContractsByTwinId,
   ContractState,
@@ -100,7 +101,7 @@ class Contracts {
 
   @expose
   @validateInput
-  async contractLock(options: ContractConsumption) {
+  async contractLock(options: ContractLockModel) {
     return this.client.contracts.contractLock(options);
   }
 
--- a/packages/grid_client/src/modules/contracts.ts
+++ b/packages/grid_client/src/modules/contracts.ts
@@ -100,6 +100,12 @@ class Contracts {
 
   @expose
   @validateInput
+  async contractLock(options: ContractConsumption) {
+    return this.client.contracts.contractLock(options);
+  }
+
+  @expose
+  @validateInput
   @checkBalance
   async update_node(options: NodeContractUpdateModel) {
     return (await this.client.contracts.updateNode(options)).apply();
--- a/packages/playground/src/weblets/tf_contracts_list.vue
+++ b/packages/playground/src/weblets/tf_contracts_list.vue
@@ -34,7 +34,7 @@
       <template #[`item.state`]="{ item }">
         <v-tooltip
           v-if="item && item.value.state === ContractStates.GracePeriod"
-          :text="'Click here to view the amount you need to charge in order to unlock the contract.'"
+          :text="'the number of tokens required to remove your contract from the grace period and restore functionality to your workloads.'"
           location="top center"
         >
           <template #activator="{ props }">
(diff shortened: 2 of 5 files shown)


packages/playground/src/weblets/tf_contracts_list.vue
```diff
@@ -32,7 +32,22 @@
       </template>
 
       <template #[`item.state`]="{ item }">
-        <v-chip :color="getStateColor(item.value.state)">
+        <v-tooltip
+          v-if="item && item.value.state === ContractStates.GracePeriod"
+          :text="'Click here to view the amount you need to charge in order to unlock the contract.'"
```
xmonader, line 37: the amount of tokens needed to get your contract out of the grace period, to have your workloads functional again.

packages/tfchain_client/src/contracts.ts
```diff
@@ -12,6 +12,16 @@ interface NameContract {
   name: string;
 }
 
+export interface ContractLockOptions {
+  id: number;
+}
+
+export interface ContractLock {
+  amountLocked: number;
+  lockUpdated: number;
```
xmonader, line 21: what does this field do?

packages/playground/src/weblets/tf_contracts_list.vue
```diff
@@ -148,7 +189,23 @@ function getStateColor(state: ContractStates): string {
...
+  contractStateDialog.value = true;
+  loading.value = true;
+  const grid = await getGrid(profileManager.profile!);
+  await grid?.contracts
+    .contractLock({ id: contractId })
+    .then((data: ContractLock) => {
+      contractLocked.value = data;
+    })
+    .catch(err => {
+      layout.value.setStatus("failed", normalizeError(err, `Failed to fetch the contract lock details.`));
```
xmonader, line 202: add the contract Id to the error message please

packages/playground/src/weblets/tf_contracts_list.vue
```diff
@@ -82,6 +97,30 @@
...
+      <v-card-text>
+        <p v-if="loading" class="text-center">
+          <strong>Loading The Locked Amount...</strong>
+        </p>
+        <p v-else class="text-center">
+          Amount Locked <strong>{{ contractLocked?.amountLocked }}</strong>
+        </p>
+        <br />
+        <v-alert type="info" variant="tonal">
+          To unlock your contract from the grace period, please ensure that you fund your account with the displayed
```
xmonader, line 113: The contract is in a `GracePeriod` condition, which means that your workloads are suspended but not deleted; in order to resume your workloads and restore their functionality, you must pay your account with the necessary tokens.

packages/playground/src/weblets/tf_contracts_list.vue
MohamedElmdary: Note: 
Using
```ts
layout.value.setStatus("failed", normalizeError(err, `Failed to fetch the contract lock details.`));
```
won't close the dialog also it will show the error in the entire contracts list view so it will confuse the user

packages/tfchain_client/src/contracts.ts
Mahmoud-Emad: all these fields come from the chain, I just wrapped them

for the lock update, this is a timestamp field, I didn't use it in any place

packages/playground/src/weblets/tf_contracts_list.vue
Mahmoud-Emad: @MohamedElmdary 
in case we closed the dialog when there is an error, will the error still appear?

packages/grid_client/src/modules/contracts.ts
```diff
@@ -98,6 +98,12 @@ class Contracts {
     return this.client.contracts.getContractIdByActiveRentForNode(options);
   }
 
+  @expose
+  @validateInput
+  async contractLock(options: ContractConsumption) {
```
AhmedHanafy725, line 103: please implement the model for it
```suggestion
  async contractLock(options: ContractLockModel) {
```

Mahmoud-Emad: updated

zaelgohary (approved):

Issue #478, opened 2023-06-05
GrafanaStatistics not working correctly in playground


MohamedElmdary:
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/31689104/1b81df45-2847-4da8-a7e5-9c89820eaa5d)

Mahmoud-Emad:
Should be fixed in #480

A-Harby:
Verified,
Error removed from console and grafana stats page is working now.
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/f329faf1-f4df-4eee-82ee-89906b8c3ba1)

PR #480, opened 2023-06-05
Fix metrics url issue.

Mahmoud-Emad:
### Description

The issue was related to an error in retrieving the node ID from the deployment data. The expectation was that the data would be sent with the node ID, but in the contracts list, the `name gateway` did not have a node ID associated with it, resulting in the raised error.

### Changes

To address this issue, a condition was added to check if the deployment has any length. If the deployment data exists, the code proceeds with retrieving the account ID and farm ID using the node ID from the first index of the deployment list.

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/478

### Screenshots

- image
- image
- image
- image

### Checklist

- [ ] Tests included
- [x] Build pass
- [x] Documentation

### DIFF
--- a/packages/playground/src/utils/getMetricsUrl.ts
+++ b/packages/playground/src/utils/getMetricsUrl.ts
@@ -1,5 +1,4 @@
 import type { GridClient } from "@threefold/grid_client";
-import type { Node } from "@threefold/tfchain_client";
 
 export interface IGrafanaArgs {
   orgID: number;


packages/playground/src/utils/getMetricsUrl.ts
```diff
@@ -24,15 +24,19 @@ export class GrafanaStatistics implements IGrafanaArgs {
...
-      this.accountID = res.accountId;
-    });
+    if (this.deployment.length) {
+      await this.grid.twins.get({ id: this.deployment[0].nodeId }).then(res => {
+        this.accountID = res.accountId;
+      });
+    }
   }
 
   async setfarmID() {
```
xmonader, line 34: i noticed it's not in the correct camel case, should be `setFarmID` instead of `setfarmID`

Mahmoud-Emad: updated

AhmedHanafy725 (changes requested):

AhmedHanafy725 (approved):

xmonader (approved):

Issue #463, opened 2023-06-01
Deleting a contract that is in `GracPeriod` is not handled correctly


mohamedamer453:
### Description

Deleting a contract that is in `GracPeriod` is not handled correctly.

An error appears in the console mentioning that the fees to delete the contract is not enough but in the ui there's no error and it just keeps loading forever.

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/101194226/4692b3bb-2efe-4b26-b1dd-06d10e1b935f)

### Expected Behavior

The contract shouldn't be deleted, and a clear error should be displayed to the user.

### Steps to reproduce

- Navigate to the contracts list.
- Delete any contract in `GracePeriod`

0oM4R:
this error is happening while deleting the keys the client stucks on that error which leads to the btn to be loading forever,
tried to investigate and I think the issue occurred on 
https://github.com/threefoldtech/tfgrid-sdk-ts/blob/8c14e7b5fb03bf86ecb61330657b1176b62fe28d/packages/tfchain_client/src/client.ts#L317
the callback is never called, and there is some logs 
````txt
yarn run v1.22.19
$ /home/kassem/tfgrid-sdk-ts/node_modules/.bin/ts-node --project tsconfig-node.json scripts/name.ts
@polkadot/util has multiple versions, ensure that there is only one installed.
Either remove and explicitly install matching versions or dedupe using your package manager.
The following conflicting packages were found:
	cjs 10.4.2	node_modules/@polkadot/util/cjs
	cjs 9.7.2 	node_modules/@polkadot/rpc-provider/node_modules/@polkadot/util/cjs
Credentials not all found in env variables. Loading all credentials from default config.json...
2023-06-05 16:31:06        API/INIT: RPC methods not decorated: transaction_unstabale_submitAndWatch, transaction_unstable_unwatch
2023-06-05 16:31:06        API/INIT: Not decorating runtime apis without matching versions: TransactionPaymentApi/2 (1 known)
Validation succeed
Lock acquired
2023-06-05 16:31:08        RPC-CORE: submitAndWatchExtrinsic(extrinsic: Extrinsic): ExtrinsicStatus:: 1010: Invalid Transaction: Inability to pay some fees , e.g. account balance too low
2023-06-05 16:31:08             DRR: 1010: Invalid Transaction: Inability to pay some fees , e.g. account balance too low
[...]

PR #487, opened 2023-06-06
Enhance deleting grace Period contracts UX 

0oM4R:
### Description

In case the user only has 0.01 tft <deleting contract fee> the contract will go on with the deleting process, but will stuck while deleting its keys, and then it needs to reload the page, now this case is handled if it fails while deleting some keys will show an error message.

### Changes
- await `signAndSend` function in `_applyExtrinsic`
- add an error notification to inform the user there are some keys not deleted.
- add an alert that deleting contracts may take a while 

### screen record 

[Screencast from 2023-06-06 16-41-08.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/62248851/eb1fc885-9bd7-445b-bfae-21829c4de58d)

### Related Issues

-#463

### Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation

### DIFF
--- a/packages/playground/src/weblets/tf_contracts_list.vue
+++ b/packages/playground/src/weblets/tf_contracts_list.vue
@@ -81,10 +81,12 @@
       </v-btn>
     </template>
   </weblet-layout>
-
   <v-dialog width="70%" v-model="deletingDialog">
     <v-card>
-      <v-card-title class="text-h5"> Are you sure you want to delete the following contracts? </v-card-title>
+      <v-card-title class="text-h5 mt-2"> Are you sure you want to delete the following contracts? </v-card-title>
+      <v-alert class="ma-4" density="compact" type="warning" variant="tonal"
+        >Deleting contracts may take a while to complete.</v-alert
+      >
       <v-card-text>
         <v-chip class="ma-1" color="primary" label v-for="c in selectedContracts" :key="c.contractId">
           {{ c.contractId }}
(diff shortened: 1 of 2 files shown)


packages/playground/src/weblets/tf_contracts_list.vue
```diff
@@ -81,10 +81,10 @@
       </v-btn>
     </template>
   </weblet-layout>
-
   <v-dialog width="70%" v-model="deletingDialog">
     <v-card>
-      <v-card-title class="text-h5"> Are you sure you want to delete the following contracts? </v-card-title>
+      <v-card-title class="text-h5 mt-2"> Are you sure you want to delete the following contracts? </v-card-title>
```
Mahmoud-Emad, line 86: Please confirm if you wish to proceed with the deletion of the following contracts.

packages/playground/src/weblets/tf_contracts_list.vue
```diff
@@ -223,7 +226,13 @@ async function onDelete() {
     contracts.value = contracts.value!.filter(c => !selectedContracts.value.includes(c));
     selectedContracts.value = [];
   } catch (e) {
-    layout.value.setStatus("failed", normalizeError(e, `Failed to delete some of the selected contracts.`));
+    if ((e as Error).message.includes("Inability to pay some fees")) {
```
Mahmoud-Emad, line 229: i think if we changed or improved this message from the client we'll have a big here also no?

packages/playground/src/weblets/tf_contracts_list.vue
0oM4R: I think we need to open an issue for this one, the change I made here is just to add some space before it.

packages/playground/src/weblets/tf_contracts_list.vue
0oM4R: also, this case is handled in the else, the error whatever it will be will appear to the user normally, if you have any idea how to make it better please tell me

Mahmoud-Emad (approved):

Issue #46, opened 2023-03-30
Inconsistency between the explorer and the dedicated nodes tables


mohamedamer453:
### Description

Inconsistency between the explorer and the dedicated nodes tables.

![image](https://user-images.githubusercontent.com/101194226/228827559-fbb046d2-c55b-4c2a-a07d-51ef0faea736.png)

As it can be seen above the data is inconsistent between the two tables for the same nodes.

Both tables are technically correct and there is no problem with the data itself, however the way of displaying the data should be unified. this should include the units, approximations etc.. and if the same data is displayed anywhere else in the dashboard or in some other place throughout the grid then it also should use the same convention.

Also the location of the columns should be consistent. for example in the explorer the columns are sorted like this: `HRU` -> `SRU` -> `MRU` -> `CRU` and in the dedicated nodes page its: `CRU` -> `HRU` -> `MRU` -> `SRU` this section should be consistent in both tables. 

and lastly the location of the unit. in the explorer its next to each value.

![image](https://user-images.githubusercontent.com/101194226/228830572-7b0a05f6-8087-4057-a8d8-b784d68c80e6.png)

While in the dedicated nodes table it's next to the header.

![image](https://user-images.githubusercontent.com/101194226/228830689-63b2f123-0a9d-45d5-9d7e-6422d763774e.png)

The `HRU` in the explorer is in `TB` while it's in `GB` in the dedicated nodes table.

0oM4R:
Should I change the dedicated nodes table to be like the explorer, or vice versa? @mohamedamer453  @AhmedHanafy725

AhmedHanafy725:
the dedicated nodes table to be like the explorer one

0oM4R:
status: PR #503  ready to review

mohamedamer453:
Verified on qanet dashboard `2.0.0-rc14`

The data is now consistent between the two tables. the order is now `CRU` -> `MRU` -> `SRU` -> `HRU` for both tables.

Units are now next to the value in each column and the approximations are consistent in both tables.

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/101194226/8f2f3e80-85a2-4f86-9c08-156a54d29b17)

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/101194226/ee7c861f-7c88-47e2-aad6-f367c00f216a)

PR #503, opened 2023-06-08
Sync display data on dedicated nodes table with node explorer 

0oM4R:
### Description

Make the dedicated nodes table display the data as the explorer table.
- now dedicated nodes table 
    image
- the explorer 
   image

### Changes

- change the order to cru, mru, sru, hru in both tables
- set the values to be Fixed(2)
- set HRU to be in TB

### Related Issues

- #46 

### Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation

### DIFF
--- a/packages/dashboard/src/portal/components/NodesTable.vue
+++ b/packages/dashboard/src/portal/components/NodesTable.vue
@@ -109,10 +109,10 @@ export default class NodesTable extends Vue {
   headers = [
     { text: "Node ID", value: "nodeId", align: "center" },
     { text: "Location", value: "location.country", align: "center" },
-    { text: "HRU (TB)", value: "resources.hru", align: "center" },
-    { text: "SRU (GB)", value: "resources.sru", align: "center" },
-    { text: "MRU (GB)", value: "resources.mru", align: "center" },
     { text: "CRU", value: "resources.cru", align: "center" },
+    { text: "MRU (GB)", value: "resources.mru", align: "center" },
+    { text: "SRU (GB)", value: "resources.sru", align: "center" },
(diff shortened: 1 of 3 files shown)


packages/dashboard/src/portal/components/NodesTable.vue
```diff
@@ -184,6 +184,10 @@ export default class NodesTable extends Vue {
   byteToGB(capacity: number) {
     return byteToGB(capacity);
   }
+
+  byteToTB(capacity: number) {
```
AhmedHanafy725, line 188: please use this method for conversion to always has the same units for the both tables https://github.com/threefoldtech/tfgrid-sdk-ts/blob/development/packages/dashboard/src/explorer/filters/toTeraOrGigaOrPeta.ts

AhmedHanafy725 (changes requested):
the order of the resources should be cru, mru, sru, hru in both tables

packages/dashboard/src/portal/lib/nodes.ts
```diff
@@ -224,7 +224,11 @@ export function generateReceipt(doc: jsPDF, node: nodeInterface) {
   return doc;
 }
 export function byteToGB(capacity: number) {
-  return (capacity / 1024 / 1024 / 1024).toFixed(0);
+  return (capacity / 1024 / 1024 / 1024).toFixed(2);
+}
+
+export function byteToTB(capacity: number) {
+  return (capacity / (1024 * 1024 * 1024 * 1024)).toFixed(2);
```
AhmedHanafy725, line 231: is it used?

AhmedHanafy725 (approved):

Issue #488, opened 2023-06-07
change the default view for the login manager


xmonader:
Yesterday in the demo when sam was trying the playground for the first time he was blocked by the fact the first tab was login and he was asked for a password, I checked the flow in foldit.tf, what they do is if there's a local storage item they show the login option

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/64129/24d0efa4-9f1f-4cb1-9d86-987209eafefc)

and if not (indication it's the first time to use on that device they show the create account form)

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/64129/62e486c2-a981-45b3-929f-54c1dd00ecda)

A-Harby:
Verified Devnet 5e5c622,
First time using the playground will only display the connect your wallet tab
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/d831bbc6-0f8e-4396-b6a5-2abd5c946951)
After that both login and connect tabs will be shown.
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/b61e1373-41ce-4e3a-bb09-64397b42abb3)

A-Harby:
Just missing a small fix is needed,
First time connecting your wallet, the password field will give an error once its open.

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/7282e0fd-8b0d-4c6b-9a38-d7617c40ae2d)

PR #509, opened 2023-06-11
Enhancing the profile manager to selectively show the password tab, contingent upon the presence of securely hashed credentials stored in the local storage.

Mahmoud-Emad:
### Description

It enhances the profile manager to selectively show the password tab, contingent upon the presence of securely hashed credentials stored in the local storage.

### Changes
- created a new interface named Credentials that includes `passwordHash` and `mnemonicHash` fields to represent securely hashed credentials.
- added new helper functions to manage the `Credentials` interface like `getCredentials`, `setCredentials` and `isStoredCredentials` to facilitate the handling and retrieval of stored credentials.
- Implemented a key-down function to enable `login` or `connect` functions when the Enter key is pressed on the password input, provided there are no validation errors.
- Change the `Store and login` button text to `Connect`

### Related Issues
- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/488
- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/519

### Screenshots
- image
- image

*Update*
- image

### Checklist

- [ ] Tests included
- [x] Build pass
- [x] Documentation

### DIFF
--- a/packages/playground/src/weblets/profile_manager.vue
+++ b/packages/playground/src/weblets/profile_manager.vue
@@ -51,103 +51,104 @@
         "
       >
         <VContainer>
-          <FormValidator v-model="isValidForm">
-            <VTooltip
-              v-if="activeTab === 1"
-              text="Mnemonic are your private key. They are used to represent you on the ThreeFold Grid. You can paste existing mnemonic or click the 'Create Account' button to create an account and generate mnemonic."
-              location="bottom"
-              max-width="700px"
-            >
-              <template #activator="{ props: tooltipProps }">
-                <PasswordInputWrapper #="{ props: passwordInputProps }">
-                  <InputValidator
-                    :value="mnemonic"
-                    :rules="[
-                      validators.required('Mnemonic is required.'),
-                      v => (validateMnemonic(v) ? undefined : { message: `Mnemonic doesn't seem to be valid.` }),
-                    ]"
-                    :async-rules="[validateMnInput]"
-                    valid-message="Mnemonic is valid."
-                    #="{ props: validationProps }"
-                  >
-                    <div v-bind="tooltipProps" v-show="!profileManager.profile" class="d-flex">
-                      <VTextField
-                        class="mb-2"
-                        label="Mnemonic"
-                        placeholder="Please insert your mnemonic"
-                        autofocus
-                        v-model="mnemonic"
-                        v-bind="{ ...passwordInputProps, ...validationProps }"
-                        :disabled="creatingAccount || activating"
-                      />
-                      <VBtn
-                        class="mt-2 ml-3"
-                        color="secondary"
-                        variant="tonal"
-                        :disabled="isValidForm"
-                        :loading="creatingAccount"
-                        @click="createNewAccount"
-                      >
-                        generate account
-                      </VBtn>
-                    </div>
-                  </InputValidator>
-                </PasswordInputWrapper>
-              </template>
-            </VTooltip>
-
-            <v-alert type="error" variant="tonal" class="mb-4" v-if="createAccountError && activeTab === 1">
-              {{ createAccountError }}
-            </v-alert>
-
-            <PasswordInputWrapper #="{ props: passwordInputProps }">
-              <InputValidator
-                :value="password"
-                :rules="[
-                  validators.required('Password is required.'),
-                  validators.minLength('Password must be at least 6 characters.', 6),
-                  validatePassword,
-                ]"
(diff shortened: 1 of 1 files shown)


MohamedElmdary:
Enter doesn't respect **Mnemonic** value

https://github.com/threefoldtech/tfgrid-sdk-ts/assets/31689104/16e9ddfb-39a6-42fd-9e36-65fa7c31c380

Using keydown.enter will cause some unexpected cases; It's better to use <[form](https://www.w3schools.com/html/html_forms.asp)> with *onsubmit* event as it won't fire event if the button is disabled

packages/playground/src/weblets/profile_manager.vue
```diff
@@ -284,13 +287,93 @@ defineProps({
...
+    if (m) {
+      if (mountedTimeout) {
+        clearTimeout(mountedTimeout);
+      }
+      mountedTimeout = setTimeout(() => {
+        mounted();
+      });
+    }
+  },
+);
```
MohamedElmdary, line 303: Please don't use `setTimeout` api to wait as it might lead to unexpected results like old playground

packages/playground/src/weblets/profile_manager.vue
```diff
@@ -284,13 +287,93 @@ defineProps({
...
+    }
+  } else {
+    activeTab.value = 1;
+    return;
+  }
+}
+
+function onKeyDown() {
+  if (password.value.length >= 6) {
+    if (activeTab.value === 0) [login()];
```
MohamedElmdary, line 326: Why do we wrap login call with array [login()]?

packages/playground/src/weblets/profile_manager.vue
```diff
@@ -284,13 +287,93 @@ defineProps({
...
+
+function getCredentials() {
+  const getCredentials = localStorage.getItem("wallet");
+  let credentials: Credentials = {};
+
+  if (getCredentials) {
+    credentials = JSON.parse(getCredentials);
+  }
+  return credentials;
+}
```
MohamedElmdary, line 341:
```suggestion

const version = 1; // for example
const WALLET_KEY = "wallet.v" + version
function getCredentials() {
  const getCredentials = localStorage.getItem(WALLET_KEY);
  let credentials: Credentials = {};

  if (getCredentials) {
    credentials = JSON.parse(getCredentials);
  }
  return credentials;
}
```

Support versioning for wallet as we might change how we manage password and mnemonic from time to time.
let's imagine if we changed the storing way when the user open his browser it will be broken as we will be expecting another way of parsing data and here versioning comes into play it won't respect any wallet beside the new one

packages/playground/src/weblets/profile_manager.vue
Mahmoud-Emad: Updated

packages/playground/src/weblets/profile_manager.vue
Mahmoud-Emad: Used `requestAnimationFrame` instead

packages/playground/src/weblets/profile_manager.vue
Mahmoud-Emad: let's assume that we changed the version of storing the wallet, in that case, we have to update the `version`, which is the way we're going to update the `wallet` key, in both cases we have to change the key, what is the difference?

MohamedElmdary: Updating the version should be done whenever we change the behaviour of the wallet so it will break the user browser
@AhmedHanafy725

packages/playground/src/weblets/profile_manager.vue
MohamedElmdary: Don't use `requestAnimationFrame` either.
Please check `nextTick`

```ts
import { nextTick } from "vue"

nextTicket().then(mounte)
```

packages/playground/src/weblets/profile_manager.vue
AhmedHanafy725: the version will make it easier to migrate from the old key to the new one. if the same key is used, you will have to check which one is used first and then decide whether to migrate to the new one or not.

packages/playground/src/weblets/profile_manager.vue
MohamedElmdary: @Mahmoud-Emad any updates?

Mahmoud-Emad: should be done

MohamedElmdary (approved):

Issue #16, opened 2023-03-09
Use url join instead of joining url parts with string formating


AhmedHanafy725:
for example https://github.com/threefoldtech/grid3_client_ts/blob/2528c0b1e9ac67e9e7d3520ecf553a3e225f163d/src/primitives/nodes.ts#L135

PR #515, opened 2023-06-11
Use join url

zaelgohary:
### Description

Replace normal string format in get request with join-url package

### Changes

change url format in nodes file

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/16

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/packages/grid_client/src/modules/algorand.ts
+++ b/packages/grid_client/src/modules/algorand.ts
@@ -1,6 +1,7 @@
 import { default as AlgoSdk } from "algosdk";
 import axios from "axios";
 import * as PATH from "path";
+import urlJoin from "url-join";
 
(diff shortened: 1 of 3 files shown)


xmonader (dismissed):
why not use `.join` with `/` ?  Also, please don't use path.sep (it will endup with the wrong behavior if anyone runs the project on windows)

zaelgohary:
> why not use `.join` with `/` ? Also, please don't use path.sep (it will endup with the wrong behavior if anyone runs the project on windows)

In the case of using .join, it may add unwanted "/" in the URL while urlJoin prevents such behavior.

image

Mahmoud-Emad (approved):

Issue #322, opened 2023-05-14
The `Generate ssh key` is not disabled for an account that doesn't have enough tfts


mohamedamer453:

mohamedamer453:
So after checking the twin that originally had the issue i found that it was one of my twins that was used in other tests and had `0.00005 TFT` which why it wasn't able to pay for the transaction.

[Screencast from 12-06-23 15:25:45.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/101194226/7a04a5c3-8fb1-43cd-8960-108e3158c41d)

but i still think that this scenario should be taken into consideration since the user might want to generate a key without realizing that there are no funds in the account. so instead of being stuck forever since there are no tokens it either should fail and mention that there are no tokens in this account or it should check from the start if the balance is `<0.01 TFT` then the `Generate SSH Key` button should be disabled.

mohamedamer453:
The `Generate SSH Keys` button is now disabled when the balance is not enough. however while verifying this i realized that the same issue happens with the `Update Public  SSH Key` button.

[Screencast from 14-06-23 16:01:00.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/101194226/29c56891-d5df-4199-9bd7-39cbf419005f)

khaledyoussef24:
verified now on devnet.
you can not now update the ssh key or generate ssh key without balance.
![Screenshot from 2023-06-26 11-07-51](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/c2cd97e8-5d96-4b0c-a7c7-c68d1f1c744a)
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/6f534ee3-43bd-4b8c-a799-8201aba186cd)

PR #551, opened 2023-06-12
Disable generate ssh on low balance

zaelgohary:
### Description

Disable generate ssh if balance is less than 0.01

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/322

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/packages/playground/src/weblets/profile_manager.vue
+++ b/packages/playground/src/weblets/profile_manager.vue
@@ -273,7 +273,7 @@ import { generateKeyPair } from "web-ssh-keygen";
 
 import { useProfileManager } from "../stores";
 import { type Balance, createAccount, getGrid, loadBalance, loadProfile, storeSSH } from "../utils/grid";
-import { normalizeError } from "../utils/helpers";
+import { isEnoughBalance, normalizeError } from "../utils/helpers";
 import { downloadAsFile, normalizeBalance } from "../utils/helpers";
 
 interface Credentials {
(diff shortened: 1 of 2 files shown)


packages/playground/src/weblets/profile_manager.vue
```diff
@@ -221,7 +221,7 @@
           <VBtn
             color="secondary"
             variant="text"
-            :disabled="!!ssh || updatingSSH || generatingSSH"
+            :disabled="!!ssh || updatingSSH || generatingSSH || (balance && balance.free < 0.01)"
```
Mahmoud-Emad, line 224: I believe that creating a helper function to check the balance and determine if it is sufficient or not would be beneficial in terms of reusability.

```ts

const isEnoughBalance = (balance: number) => {
    // Check the free balance and return true if it's enough for making transactions or not
    return (balance && balance.free < 0.01) ? true : false
}
```
```suggestion

            :disabled="!!ssh || updatingSSH || generatingSSH || !isEnoughBalance()"
```

AhmedHanafy725, line 224: is it 0.01 or 0.001?

zaelgohary: Edited.

AhmedHanafy725 (approved):

Mahmoud-Emad (approved):
LGTM, Good job ya zizi :)

Issue #524, opened 2023-06-12
Number of locked TFTs doesn't appear if it's 0


mohamedamer453:
### Description

Number of locked tfts doesn't appear if it's 0.

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/101194226/edb7853e-e8f8-4a20-9300-cc7354b030a7)

### Expected result

It should display `0 TFT`.

A-Harby:
Verified, Devnet c765569.

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/fabd4d43-fdd3-41cd-b206-2ffc4589047a)

PR #560, opened 2023-06-13
Fix locked balance

zaelgohary:
### Description

When the locked balance is undefined, set it to 0

image

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/524

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/packages/playground/src/weblets/profile_manager.vue
+++ b/packages/playground/src/weblets/profile_manager.vue
@@ -21,7 +21,8 @@
               Balance: <strong :style="{ color: '#76e2c8' }">{{ normalizeBalance(balance.free, true) }} TFT</strong>
             </p>
             <p>
-              Locked: <strong :style="{ color: '#76e2c8' }">{{ normalizeBalance(balance.locked, true) }} TFT</strong>
+              Locked:
+              <strong :style="{ color: '#76e2c8' }">{{ normalizeBalance(balance.locked, true) || 0 }} TFT</strong>
             </p>
           </template>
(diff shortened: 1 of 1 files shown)


Mahmoud-Emad (approved):

packages/playground/src/weblets/profile_manager.vue
```diff
@@ -21,7 +21,8 @@
               Balance: <strong :style="{ color: '#76e2c8' }">{{ normalizeBalance(balance.free, true) }} TFT</strong>
             </p>
             <p>
-              Locked: <strong :style="{ color: '#76e2c8' }">{{ normalizeBalance(balance.locked, true) }} TFT</strong>
+              Locked:
+              <strong :style="{ color: '#76e2c8' }">{{ normalizeBalance(balance.locked, true) || 0 }} TFT</strong>
```
Mahmoud-Emad, line 25: Is it possible to do the same with the balance instead of a boolean value?

Mahmoud-Emad (changes requested):

packages/playground/src/weblets/profile_manager.vue
zaelgohary: The free balance will always be a number and even if not found, it'll return 0. However, this is not the case in free balance which returns undefined in case of not found.

image

Mahmoud-Emad (approved):

Issue #563, opened 2023-06-13
support info markdown pages


MohamedElmdary:
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/31689104/62793ba0-3076-4d6e-8d83-d26c48c7524a)

khaledyoussef24:
verified and works fine for all solution.
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/142710be-776e-435b-9b3e-5c14a4e3ad26)

PR #566, opened 2023-06-13
Add a dynamic way to add info for required pages only

MohamedElmdary:
## Issues
- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/563

### Updates
- Allow to add markdown pages indicates info about the page
- Allow to change tooltip if needed (default: Show Details)
- Allow to use {title, subtitle} as metadata in markdown (optional)
- Accepting .md and .html (as markded.js *js markdown parser* supports both)

#### Preview
image
image
image
image

`Note: An example is included for preview and should be replaced with the real data before merge PR`


AhmedHanafy725:
all workflows are failing

AhmedHanafy725 (approved):

PR #642, opened 2023-06-18
add tft price test case

A-Harby:
### Description

Add new test case for tft price 

image
image

### DIFF
--- a/packages/dashboard/tests/frontend_selenium/tests/test_homepage.py
+++ b/packages/dashboard/tests/frontend_selenium/tests/test_homepage.py
@@ -206,3 +206,19 @@ def test_search_by_invalid_name_address(browser, cases):
     dashboard_page.search_accounts(cases.lower())
     assert address not in address_list and address not in dashboard_page.accounts_list()
     assert len(account_list) == 0 and len(dashboard_page.accounts_list()) == 0
+
+
+def test_tft_price(browser):
+    """
+      Test Case: TC1674 - TFT price
+      Steps:
+          - Navigate to dashboard.
+          - Authenticate polkadot.
+          - Click on TFT swap icon.
+      Result: TFT in USD and vice versa.
+    """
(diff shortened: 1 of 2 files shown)


packages/dashboard/tests/frontend_selenium/tests/test_homepage.py
```diff
@@ -206,3 +206,19 @@ def test_search_by_invalid_name_address(browser, cases):
...
+          - Navigate to dashboard.
+          - Authenticate polkadot.
+          - Click on TFT swap icon.
+      Result: TFT in USD and vice versa.
+    """
+    dashboard_page, _ = before_test_setup(browser)
+    tft_in_usd = float(dashboard_page.tft_price_result()[:-4])
+    dashboard_page.tft_price_swap()
+    usd_in_tft = float(dashboard_page.tft_price_result()[:-4])
+    assert 0.99 < tft_in_usd * usd_in_tft < 1.1
```
mohamedamer453, line 224: I'm not really sure if this assertion does anything except making sure that the conversion is correct. i think the best scenario in this case is to get the tft price from the same source/api and assert that the values are equal. this assertion will make sure that the value is correct and it wasn't affected during the process of retrieving it from the source and displaying it to the user. however the current one just makes sure that the conversion of the two values is correct and whatever these values are if the ratio between them is the same then it will pass. 

for example if the the current `tft_in_usd` is 90.909 and `usd_in_tft` is 0.011 so the calculation will be 0.999999 and the assertion will pass. but for the example if we modified the values to 900.909 and 0.0011 it will also be 0.9909999 and the test will pass so making sure that the conversion is correct isn't really be meaningful in this case. but it can still be checked as a secondary assertion after asserting that the values are retrieved correctly.

mohamedamer453 (approved):

xmonader (approved):

PR #684, opened 2023-06-21
Disable updateSSH on low balance

zaelgohary:
### Description

Disable updateSSH on low balance

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/322

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/packages/playground/src/weblets/profile_manager.vue
+++ b/packages/playground/src/weblets/profile_manager.vue
@@ -271,7 +271,7 @@
             color="primary"
             variant="text"
             @click="updateSSH"
-            :disabled="!ssh || profileManager.profile.ssh === ssh || updatingSSH"
+            :disabled="!ssh || profileManager.profile.ssh === ssh || updatingSSH || !isEnoughBalance(balance)"
             :loading="updatingSSH"
           >
             Update Public SSH Key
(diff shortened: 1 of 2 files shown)


packages/playground/src/weblets/profile_manager.vue
```diff
@@ -271,7 +271,7 @@
             color="primary"
             variant="text"
             @click="updateSSH"
-            :disabled="!ssh || profileManager.profile.ssh === ssh || updatingSSH"
+            :disabled="!ssh || profileManager.profile.ssh === ssh || updatingSSH || !isEnoughBalance(balance)"
```
MohamedElmdary, line 274:
```suggestion
            :disabled="!ssh || profileManager.profile.ssh === ssh || updatingSSH || !isEnoughBalance(balance, 0.001)"
```

Please update isEnoughBalance to  take the balance and the amount to check for because right now no idea what is the enough balance without digging into isEnoughBalance ...

zaelgohary: Wouldn't that leave room for error if the second param does not equal the lowest possible balance? The minimum balance should be the same in all cases and the suggested fix won't ensure that.

zaelgohary: Added a minimum amount by default in isEnoughBalance.

MohamedElmdary (approved):

AhmedHanafy725 (approved):

Issue #499, opened 2023-06-08
GPU: Allow a farmer to set a custom "extra" price on a node in the dashboard


DylanVerstraete:
To complete https://github.com/threefoldtech/home/issues/1392 we need on chain a way to set an extra price. But we also need a flow in the dashboard for a farmer to set a custom price.

I would suggest adding the new "button" on the farmer's node overview page where a modal opens so he can view/set an extra price. 

A call to tfchain `smartContractModule.setDedicatedNodeExtraFee` must be mode with (node_id: u32, extra_fee: u64).

The `extra_fee` is expressed in mUSD (milli usd => USD * 1000)

Errors will be thrown when:

- the node has active deployments
- the node is rented already

AhmedHanafy725:
for these errors, does it mean that the farmer can't set the extra fee if the node has node/rent contracts?

DylanVerstraete:
> he farmer can't set the extra fee if the node has node/rent contracts

Yes indeed

renauter:
See some specs from TFChain side here:
https://github.com/threefoldtech/tfchain/issues/725

A-Harby:
Verified.
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/323fc4df-7eb4-4a1e-89ea-5fb0d145b0df)
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/3a77e7ff-51f0-459e-9d46-d76c5470f116)
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/72367b64-427d-44c3-a5b9-df8103fbb45f)

PR #705, opened 2023-06-22
Added extra fees dialogue

zaelgohary:
### Description

- Added a new action in your farm nodes section
- Added extra fees dialogue

image

image

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/499

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/packages/dashboard/src/portal/components/FarmNodesTable.vue
+++ b/packages/dashboard/src/portal/components/FarmNodesTable.vue
@@ -337,7 +337,7 @@
                 required
                 outlined
                 dense
-                hint="Extra fees are in USD"
+                hint="Extra fees are in USD/month"
                 persistent-hint
                 type="number"
                 :error-messages="extraFeeErrorMessage"
@@ -411,6 +411,7 @@
   </div>
 </template>
 <script lang="ts">
+import { QueryClient } from "@threefold/tfchain_client";
 import jsPDF from "jspdf";
 import { default as PrivateIp } from "private-ip";
 import { Component, Prop, Vue } from "vue-property-decorator";
@@ -420,7 +421,7 @@ import { addNodePublicConfig, deleteNode, nodeInterface } from "@/portal/lib/far
 import { byteToGB, generateNodeSummary, generateReceipt, getNodeUptimePercentage } from "@/portal/lib/nodes";
 import { hex2a } from "@/portal/lib/util";
 
-import { getDedicatedNodeExtraFee, setDedicatedNodeExtraFee } from "../lib/nodes";
+import { setDedicatedNodeExtraFee } from "../lib/nodes";
 import ReceiptsCalendar from "./ReceiptsCalendar.vue";
 
 @Component({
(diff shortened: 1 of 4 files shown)


AhmedHanafy725 (changes requested):

packages/dashboard/src/portal/components/FarmNodesTable.vue
```diff
@@ -314,6 +323,45 @@
...
+          <v-card-text class="text">
+            <v-form v-model="isValidExtraFee">
+              <v-text-field
+                class="mt-4"
+                label="Extra Fees"
+                v-model="extraFee"
+                required
+                outlined
+                dense
+                hint="Extra fees are in USD"
```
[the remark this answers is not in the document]
xmonader: In general the GPU is already very expensive to be expressed in milli USD, it's already expressed in general with USD/hr on other cloud providers. Let's put it USD/Month instead for the farmer

xmonader:
I suggest 
- we use `Set Fees` or `Set Additional Fees`  as a title (tooltip/explanation) most useful to set the price for GPU or special hardware added to the node (or something like that)
the unit should be on the right so I guest

- field row to be `[  10    ] USD/month (tooltip)` about the payment going to happen in TFT nevertheless but with price anchored to USD, or in the field label itself we say it's USD/month

packages/dashboard/src/portal/components/FarmNodesTable.vue
```diff
@@ -629,6 +686,27 @@ export default class FarmNodesTable extends Vue {
     }
     this.openPublicConfigDialog = true;
   }
+
+  async openExtraFee(node: nodeInterface) {
+    this.nodeToEdit = node;
+    // convert fees from USD to mili USD while getting
+    this.extraFee =
+      (await this.queryClient.contracts.getDedicatedNodeExtraFee({ nodeId: this.nodeToEdit.nodeId })) / 1000;
```
AhmedHanafy725, line 694: the value should be multiplied by 1000 not divided

packages/dashboard/src/portal/components/FarmNodesTable.vue
```diff
@@ -629,6 +686,27 @@ export default class FarmNodesTable extends Vue {
...
+  }
+
+  async saveExtraFee(fee: number) {
+    this.loadingExtraFee = true;
+    // convert fees from mili USD to USD while setting
+    setDedicatedNodeExtraFee(this.$store.state.credentials.account.address, this.nodeToEdit.nodeId, fee * 1000).then(
+      () => {
+        this.$toasted.show(`Transaction succeeded: Fee is added to node ${this.nodeToEdit.nodeId}`);
+        this.loadingExtraFee = false;
+        this.openExtraFeeDialogue = false;
```
AhmedHanafy725, line 705: can we also handle the failure case?
it keeps loading forever
image

AhmedHanafy725 (changes requested):

packages/dashboard/src/portal/components/FarmNodesTable.vue
AhmedHanafy725: thought it was on setting not the getting, so this should be ignored

zaelgohary: It's divided while getting & multiplied while setting.

packages/dashboard/src/portal/components/FarmNodesTable.vue
zaelgohary: handled.

PR #802, opened 2023-07-11
Add build files for ARM

scottyeager:
### Description

Adds a Dockerfile and Github workflow to build and release a `grid_http_server container` for the ARM architecture. See https://github.com/threefoldtech/farmerbot/issues/11 for motivation and discussion.

### Notes

* All changes are adjacent to the existing workflows, to prevent any issues or complications with the mainline builds
* The container images are tagged with a suffix `-arm`, for example `latest-arm` . While different architecture can coexist under a single tag, it's only possible if all are built under a single workflow run (avoided to preserve the point above)
* This workflow is triggered by the completion of the main release workflow to ensure the desired release can be obtained with `yarn`.  This complicated tagging the image, thus the included `github-script` step
* Any suggestions regarding how to organize the new files, in terms of directory structure or naming conventions, are welcome

### DIFF
--- a/.github/workflows/publish_arm.yml
+++ b/.github/workflows/publish_arm.yml
@@ -1,0 +1,67 @@
+name: Publish ARM
+
+on:
+  workflow_run:
+    workflows: ["Full Clients Publish"]
+    types:
+      - completed
+
+jobs:
+  docker-image-grid-http-server:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+    steps:
+      - name: 'Get tag'
+        id: tag
+        uses: actions/github-script@v6
+        with:
+          result-encoding: string
+          script: |
+            let run = await github.rest.actions.getWorkflowRun({
+               owner: context.repo.owner,
+               repo: context.repo.repo,
+               run_id: context.payload.workflow_run.id,
+            });
+            return run.data.head_branch;
+
+      - name: Checkout the repo
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ steps.tag.outputs.result }}
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+
+      - name: Log in to the Container registry
+        uses: docker/login-action@v2.1.0
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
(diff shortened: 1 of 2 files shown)


.github/workflows/publish_arm.yml
```diff
@@ -0,0 +1,67 @@
+name: Publish ARM
+
+on:
+  workflow_run:
+    workflows: ["Full Clients Publish"]
+    types:
+      - completed
+
+jobs:
```
brandonpille, line 9: This needs to run after the build from [here](https://github.com/threefoldtech/tfgrid-sdk-ts/blob/12a427059196245379a567536f3edaa238daadad/.github/workflows/publish.yml#L12). I'm not sure it is the case. It's important because the docker image installs the specified GRID_VERSION (env argument) which might not yet be published when this job starts.

scottyeager: These lines trigger the arm build workflow after each completion of the main publish workflow, not before. Below in the 'Get tag' step, we extract the tag that triggered the main publish workflow so we can use it as needed.

.github/workflows/publish_arm.yml
```diff
@@ -0,0 +1,67 @@
...
+      - name: Checkout the repo
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ steps.tag.outputs.result }}
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
```
AhmedHanafy725, line 38: why do we use `qemu` and `buildx` here?

.github/workflows/publish_arm.yml
AhmedHanafy725: for the current flow, it will run after the `publish` workflow is completed regardless of its success or not. 
so can we please check if the publish workflow is completed successfully like this [example](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-a-workflow-based-on-the-conclusion-of-another-workflow)?

scottyeager: I have added the condition of success for the main build workflow.

.github/workflows/publish_arm.yml
scottyeager: Some libraries that we depend on use native binaries that are not prepackaged for ARM and thus need to be compiled. Building under emulation with QEMU is a convenient way to do this without needing to bother with cross compiling.

AhmedHanafy725 (approved):

Issue #65, opened 2022-11-02
(Mainnet verification) Node download receipts pdf formatting


amiraabouhadid:
as of now the pdf only shows the first 8 receipts of a node, need to make it show all receipts of a node in the farm nodes receipts summary

Omarabdul3ziz:
update: wip in this pr #814

Omarabdul3ziz:
update: fixed in this pr #814

khaledyoussef24:
Verified on mainnet
Downloading minting pdf files for the receipts working fine
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/8121e0ad-4f94-49ad-ace6-e856c02ec37a)

Issue #61, opened 2023-01-10
Minting details in Dashboard


sabrinasadik:

AhmedHanafy725:
> ### 1. Current farm report
> Showing uptime for the current minting period 
  > - [ ] farm id
  > - [ ] uptime % (as explained here: https://github.com/threefoldtech/tfgrid_dashboard/issues/366#issuecomment-1304436906)
  > - [ ] payout address 

showing the uptime percentage only on the PDF report?

I think the button to download the PDF report should be on the farm instead of the nodes table as the node might belong to another farm. so each farm should have its own report. what do you think?


> ### 2. Historical farm report since creation date 
>  - [ ] farm id
>  - [ ] scrolling list of all months (not calendar, just select month in dropdown, include 'all' and then January, February,...)
>  - for each month
>    - [ ] date of payout **(needed for tax reasons)**

`date of the payout` how to get that? we can get the minting start/end date, but can't find the payout date in the minting endpoint

>    - [ ] node TFT amount
>    - [ ] detail of calculation CU, SU, NU, IP

found that on the endpoint, is it what is required to be added?
```
"minted_cloud_units": {
          "cu": 3.631928,
          "su": 336.174283,
          "nu": 0.544711
        },
```
also if the node went in fixup this month should be shown or it will be dropped? (this is now shown in the calendar)

AhmedHanafy725:
@sabrinasadik can you check please?

sabrinasadik:
> > ### 1. Current farm report
> > Showing uptime for the current minting period
> > 
> > * [ ]  farm id
> > * [ ]  uptime % (as explained here: [Node got over 100% uptime  tfgrid_dashboard#366 (comment)](https://github.com/threefoldtech/tfgrid_dashboard/issues/366#issuecomment-1304436906))
> > * [ ]  payout address
> 
> showing the uptime percentage only on the PDF report?
> 
> I think the button to download the PDF report should be on the farm instead of the nodes table as the node might belong to another farm. so each farm should have its own report. what do you think?
> OK
> > ### 2. Historical farm report since creation date
> > 
> > * [ ]  farm id
> > * [ ]  scrolling list of all months (not calendar, just select month in dropdown, include 'all' and then January, February,...)
> > * for each month
> >   
> >   * [ ]  date of payout **(needed for tax reasons)**
> 
> `date of the payout` how to get that? we can get the minting start/end date, but can't find the payout date in the minting endpoint
[...]

PR #814, opened 2023-07-16
Minting details in Dashboard 

Mahmoud-Emad:
### Description
Enhance minting receipts in dashboard

### Changes
- reorganize node details in nodes table
- replace ReceiptsCalender with NodeMinitingDetails component
- get extra info on the receipt call
- view all the receipts in the PDF not just the first eight.
- add extra info for the receipts in the PDF
- remove `Download All receipts` button and add `Download Farm receipts`

### Related Issues
- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/61
- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/65

### UI Changes Preview 
image


AhmedHanafy725 (changes requested):
can we hide the download receipts button when any network is used except mainnet?
image

packages/dashboard/src/portal/lib/farms.ts
```diff
@@ -84,6 +85,12 @@ export async function getFarmPayoutV2Address(
   const address = await api.query.tfgridModule.farmPayoutV2AddressByFarmID(id);
   return address.toJSON();
 }
+
+export async function getNodesCount(api: { query: any }, farmID: number) {
+  const nodes = await axios.get(`${config.gridproxyUrl}/nodes?farm_ids=${farmID}`);
```
AhmedHanafy725, line 90: I think we should get the nodes' count from the ret_count header instead as the farm may has more than 50 nodes

Issue #738, opened 2023-07-02
The disclaimer text in the stats website overlaps with the map at certain window sizes


mohamedamer453:
### Description

The disclaimer text in the stats website for `dev`, `qa`, `test` networks is overlapping with the map when the window size is decreased.

[Threefold Capacity.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/101194226/42dcf723-e2be-459f-859d-677deba87a68)

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/101194226/8aaaa9da-a7a0-42b2-9bf1-3868dbe4bdbd)

### Steps to reproduce

- Navigate to https://stats.dev.grid.tf (or qa, test)
- Decrease the size of the window.

ramezsaeed:
we can also please add "Net" the end of the ine.

mohamedamer453:
Verified on devnet.

Now the text is above the map and it doesn't overlap with it under any conditions.

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/101194226/431a07d7-fa21-4de5-8f2e-6593e2a251b0)

PR #843, opened 2023-07-24
update disclamer

ehab-hassan:
[fix The disclaimer text #738](https://github.com/threefoldtech/tfgrid-sdk-ts/issues/738#event-9875043607)

image
image

### DIFF
--- a/packages/stats/src/App.svelte
+++ b/packages/stats/src/App.svelte
@@ -32,7 +32,7 @@
     const networks = Object.entries($status)
       .filter(x => x[1])
       .map(x => x[0]);
-    return `Please note that the current node distribution is for ${networks.join(", ")} net.`;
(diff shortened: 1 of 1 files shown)


packages/stats/src/App.svelte
```diff
@@ -32,7 +32,7 @@
     const networks = Object.entries($status)
       .filter(x => x[1])
       .map(x => x[0]);
-    return `Please note that the current node distribution is for ${networks.join(", ")}.`;
+    return `Please note that the current node distribution is for ${networks.join(", ")} net.`;
```
AhmedHanafy725, line 35:
```suggestion
    return `Please note that the current node distribution is for ${networks.join(", ")} net(s).`;
```

ehab-hassan: fixed
image

Issue #712, opened 2023-06-23
add playground notifications in case of graceperiod contracts


xmonader:
maybe every 15mins show a notification to the user that they have contracts in graceperiod (allow the user to snooze the during the session if they want)

A-Harby:
Verified, Devnet 73edbcd.

Notification is added and if user has a grace period contracts, he will be notified once logged in and every 15 minute.
 
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/44bca3d4-4bd6-41cb-ae69-0e1ae327fc57)

PR #846, opened 2023-07-25
Notify user when a contract is in grace period

maayarosama:
### Description

Notify user when a contract is in grace period every 15 minutes

### Changes
- Adding a notification component
### Related Issues

#712
### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/packages/playground/src/components/tf_notification.vue
+++ b/packages/playground/src/components/tf_notification.vue
@@ -1,0 +1,47 @@
+<template>
+  <div></div>
+</template>
+
+<script lang="ts" setup>
+import "mosha-vue-toastify/dist/style.css";
+
+import { ContractStates } from "@threefold/grid_client";
+import { createToast } from "mosha-vue-toastify";
+import { onMounted } from "vue";
+import { ref } from "vue";
+
+import { useProfileManager } from "../stores";
+import { getUserContracts, type NormalizedContract } from "../utils/contracts";
+import { getGrid } from "../utils/grid";
+
+const profileManager = useProfileManager();
+const contracts = ref<NormalizedContract[]>([]);
+
+onMounted(async () => {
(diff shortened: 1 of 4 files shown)


packages/playground/src/components/tf_notification.vue
```diff
@@ -0,0 +1,47 @@
...
+          position: "top-right",
+          hideProgressBar: true,
+          toastBackgroundColor: "red",
+          timeout: 5000,
+        });
+      }
+    }
+    await new Promise(resolve => setTimeout(resolve, 15 * 60 * 1000));
+  }
+});
```
MohamedElmdary, line 40: Using while (true) without a break condition will force the code to continue even if `profileManager.profile` isn't no longer exists
so let's say a *USER* login into the tfchain wallet then logout then login again ...etc
this code will keep running as many times as the *USER* login

packages/playground/src/components/tf_notification.vue
```diff
@@ -0,0 +1,46 @@
...
+
+const profileManager = useProfileManager();
+const contracts = ref<NormalizedContract[]>([]);
+
+onMounted(async () => {
+  contracts.value = [];
+
+  while (profileManager.profile) {
+    const grid = await getGrid(profileManager.profile!);
+    contracts.value = await getUserContracts(grid!);
```
AhmedHanafy725, line 25: I think it's easier to get the contract in the grace period instead of getting them, then start filtering them.
you can do that by 
`gridClient.contracts.listMyContracts({state: [ContractStates.GracePeriod]})`

AhmedHanafy725 (changes requested):

AhmedHanafy725 (approved):

Issue #721, opened 2023-06-26
Add search capabilities in dedicated nodes page


AhmedHanafy725:
should support search with:

- CPU
- memory
- SSD disk
- HDD disk
- GPU's vendor
- GPU's device

require: https://github.com/threefoldtech/tfgrid-sdk-go/issues/209

Mahmoud-Emad:
Currently working on handling the old implementation and re-implementing the feature from scratch to match the nodes table inside the explorer.

Mahmoud-Emad:
WIP: The search keys you requested are now supported on the 'portal/dedicated nodes' page, alongside the table

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/57001890/42326c3a-2aec-42a7-adb2-f5fb319f747a)


I am currently awaiting updates regarding the [necessary issue](https://github.com/threefoldtech/tfgrid-sdk-go/issues/209) in order to support the newly added keys and provide an accurate outcome.

Mahmoud-Emad:
Working on it, the proxy indexer were added

Mahmoud-Emad:
The [mentioned pull request](https://github.com/threefoldtech/tfgrid-sdk-ts/pull/730) is ready for review, will move this issue into verification column once it got merged

mohamedamer453:
### CPU filter

The cpu filter is not working with any value.

[TF DASHBOARD (3).webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/101194226/7f937c27-2527-443c-98c8-a2567c649c37)

### SRU, MRU, HRU filters

These filters are working fine but the decimal values are not allowed even though there values in the table are in decimal.

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/101194226/72db6517-0a79-4e69-b21a-06b0120d78d1)

and the error message is also not accurate when a decimal value is entered `This field must be a number larger than 0.`

### GPU

GPU filters are working but they're too strict, for example in the `GPU's vendor name` if i entered `amd` nothing will be returned and i must type `Advanced Micro Devices` for it to return something.

The same issue is present in the `GPU's device name` if i searched with `Radeon RX 7900 XT` it will not return anything even though that's the correct model of the gpu and i have to enter it the exact same way as written in the details for it to be returned.

[...]

PR #875, opened 2023-07-27
Send the CRU in VCores instead of bytes, update the validation error message

Mahmoud-Emad:
### Related issues
- see [this commenct](https://github.com/threefoldtech/tfgrid-sdk-ts/issues/721)

- Update the validation error message in case a not-valid number entered 
- - image
- Update the Sur, Hru, and Mru fields to accept decimal numbers
- - image
- Filter the Cru based on VCores instead of bytes 
- - image
- Search with `IContaines` supported from the grid proxy
- - image


Mahmoud-Emad:
WIP: The decimal point number should be working on the sru, hru and mru fields only

AhmedHanafy725 (approved):

Issue #757, opened 2023-07-05
list dedicated nodes in the capacity filter


xmonader:
as a user i may have rented a node on farm X, and while deploying on farm X, there's no way to specify if i want to deploy on a shared node or a dedicated node (from what i understood from devs, dedicated nodes get priority)

I don't think there's an option, but to restore the node selector or listing the dedicated nodes in a farm to choose from.

Mahmoud-Emad:
Searching with @ehab-hassan for the best flow

ehab-hassan:
What about adding a switch to show a shared node, but that'll be clickable when the user already had a shared node?
![Frame 60](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/10484295/0189aa36-0cc0-438a-baba-18ae3648d3ef)

xmonader:
the user reserves a dedicated node, not a shared node

zaelgohary:
Edited the filters in select farm to include dedicated but it's not working although I have rented a dedicated node. I tried to use farmerbot script from the client and it returns no available nodes.

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/40770501/ab42d9a7-aeff-48c9-9b68-f4609d387a52)


![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/40770501/5bc1a26f-50cd-4f2b-82dc-7bff09aae82a)

will continue working on it tomorrow.

zaelgohary:
In order for Farmerbot to be able to find a node, the node configuration file needs to be updated with the certified & dedicated nodes manually. An [issue](https://github.com/threefoldtech/farmerbot/issues/53) was opened to change this flow. The PR will be ready for review once the issue is closed.

zaelgohary:
Used Grid Proxy to list dedicated nodes in Playground instead of Farmerbot.

zaelgohary:
Dedicated nodes are now listed in Fullvm, Microvm & Freeflow. I'm working on adding them to the other solutions.

zaelgohary:
Refactored select node to include all nodes (dedicated & shared). Also added a chip beside each node so it would be easily labeled. Added the changes in fullvm & micro.

zaelgohary:
Select nodes are now implemented in all solutions except for Caprover & K8s.

zaelgohary:
All solutions have select node component.

mohamedamer453:
[...]

PR #876, opened 2023-07-30
Set timeout before loading nodes

zaelgohary:
### Description

- Set a timeout and cancel any previously scheduled function calls before loading nodes
- Set deep option to false in props watcher

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/757

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/packages/playground/src/components/select_node.vue
+++ b/packages/playground/src/components/select_node.vue
@@ -127,6 +127,8 @@ const cards: NodeGPUCardType[] = [];
 const emptyResult = ref(false);
 const validator = ref();
 const pingingNode = ref(false);
+const timeoutId = ref();
+
 watch(selectedCards, async () => {
   for (const card of nodeCards.value) {
     for (const selectedCard of selectedCards.value) {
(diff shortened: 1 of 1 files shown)


packages/playground/src/components/select_node.vue
```diff
@@ -206,7 +209,10 @@ watch([loadingNodes, shouldBeUpdated], async ([l, s]) => {
       availableNodes.value = [];
       return;
     }
-    loadNodes(farmId);
+    clearTimeout(timeoutId.value);
+    timeoutId.value = setTimeout(() => {
+      loadNodes(farmId);
+    }, 1000);
```
AhmedHanafy725, line 215: I think 1 second is a bit low, I don't think it will work on mainnet. can we increase the timeout? (maybe 5 seconds)

AhmedHanafy725: this is a delay to start loading not a timeout on the loading itself. so ignore this comment

AhmedHanafy725 (approved):

Issue #854, opened 2023-07-26
Private SSH Key error in Caprover


AlaaElattar:
I want trying to add my ssh private key to join as a worker node in caprover cluster, but it gives my error that `something bad went bad`. 
I did ssh to the leader node to get logs, that's what i found 

> July 26th 2023, 12:48:03.648 pm    Error: Cannot parse privateKey: Unsupported key format
> Error: Cannot parse privateKey: Unsupported key format
>     at Client.connect (/usr/src/app/node_modules/ssh2/lib/client.js:261:15)
>     at /usr/src/app/built/docker/DockerUtils.js:54:22
>     at new Promise (<anonymous>)
>     at /usr/src/app/built/docker/DockerUtils.js:16:20
>     at runMicrotasks (<anonymous>)
>     at processTicksAndRejections (internal/process/task_queues.js:95:5)
> POST /api/v2/user/system/nodes 500 14.746 ms - 21 

That's probably because my private key is OPENSSH not RSA private key.
I think maybe we should support more than one type of keys or just document for users to user only RSA Private Key.

zaelgohary:
A note will be added to the docs to make sure users only use SSH keys of RSA formats only. The issue should be verified once [this](https://github.com/threefoldtech/info_grid/issues/120) is closed.

zaelgohary:
A note is added to the manual for SSH format.

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/40770501/fe01a58a-f826-4907-a8fd-7b9d0787148f)

xmonader:
It there on the staging server https://www2.manual.grid.tf/playground/caprover.html

PR #892, opened 2023-08-01
Disable updating ssh if not in RSA format

zaelgohary:
### Description

- Enforce RSA format for SSH key

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/854

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation


xmonader (commented):
but some people use other formats, that works well on VMs, I don't think we should do that

Issue #886, opened 2023-07-31
(Mainnet Verification) Storage pool error : "not enough space left in pools" 


0oM4R:
#### using grid client v2.0.0 on main net

I tried to deploy vm with 2 disks  200GB and 700GB;  
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/62248851/7dceda94-5a82-4000-93e7-432f4680d64f)

this order failed but if i reverse the disks order  to be  [700,200]  the deployment success 
 

####  grid client logs 
```
➜  grid_client git:(v2.0.0) ✗ yarn run ts-node --project tsconfig-node.json scripts/single_vm.ts         
yarn run v1.22.19
$ /home/kassem/threefold/tfgrid-sdk-ts/node_modules/.bin/ts-node --project tsconfig-node.json scripts/single_vm.ts
Credentials not all found in env variables. Loading all credentials from default config.json...
2023-07-30 19:40:07        API/INIT: RPC methods not decorated: transaction_unstable_submitAndWatch, transaction_unstable_unwatch
2023-07-30 19:40:07        API/INIT: Not decorating runtime apis without matching versions: TransactionPaymentApi/2 (1 known)
Validation succeed
Start creating the machine deployment with name newVMS
Loading network wedtest
Creating a vm on node: 8, network: wedtest with private ip: 10.249.2.2
Merging workloads
Validation succeed
Deploying on node_id: 8
Lock acquired
2023-07-30 19:40:24        API/INIT: Not decorating runtime apis without matching versions: TransactionPaymentApi/2 (1 known)
phase: {"applyExtrinsic":10}, section: balances, method: Withdraw, data: ["5HHRg2nVfTXD9StGihrjHre5LhC2xVRZhBYvVqoorvsqo5F5",10198]
phase: {"applyExtrinsic":10}, section: smartContractModule, method: ContractCreated, data: [{"version":4,"state":{"created":null},"contractId":46199,"twinId":8574,"contractType":{"nodeContract":{"nodeId":8,"deploymentHash":"0x6137613830333834386531373830333930333139343031666263353762326133","deploymentData":"0x7b2274797065223a22766d222c226e616d65223a226e6577564d53222c2270726f6a6563744e616d65223a22227d","publicIps":0,"publicIpsList":[]}},"solutionProviderId":null}]
phase: {"applyExtrinsic":10}, section: utility, method: ItemCompleted, data: []
[...]

Issue #864, opened 2023-07-27
(Mainnet Verification) Check storage pools before deploying VMs


AhmedHanafy725:
- on the grid client, add the check before sending the deployment to the node
- on the playground, use it instead of the ping nodes to check if the disk size can fit on the storage pools

0oM4R:
__Update__: not that much progress now i can get the storage pool and the deploying disks to the checkStorage function,
now I'm trying to implement the search logic to fit all disks in side the available pool

0oM4R:
implemented the search logic; need to sync whit zos team to know how zos distribute the disks on the storage pool

0oM4R:
- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/886
 will sort disks before deployment

0oM4R:
__Update__:
update pr #897  to check hdd disks and check multiple rootfs disks <in case of multiple vms>

A-Harby:
@0oM4R Small comment; using `Checking if the disks will fit in the node's storage pools...`  maybe isn't the best statement to be used for the end user, Isn't clear what is being checked and will cause a confusion if the check fails (user would have no idea what caused the error or how to fix it).

0oM4R:
@A-Harby Can you suggest something please?

A-Harby:
> @A-Harby Can you suggest something please?

I think our main focus should be on improving the statement provided when the check fails, it's better to tell the user what to do rather than what failed, so giving him statements like lower the disks size or try other node (maybe if possible you can clarify which disk size was larger than node's disk and print sizes of them). 

And maybe for the check itself it could be 'Checking the compatibility between the disks and the collection of the disks within the node...' , But if we add the part above the current check would be also be good enough.

A-Harby:
Verified, mainnet 2.1.0.
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/621cbd16-d302-44ed-86ef-4b04f86126c1)

PR #897, opened 2023-08-01
add check storage pool to gridclient 

0oM4R:
### Description

add a function that Verifies the node's storage pool capacity to accommodate required disks and the root file system.
this function sort the passed disks and try to fit them inside the node's storage pool then do the same for root file system disk;
it throws errors if there is an error getting the node, or if required disks cannot fit in the storage pools. 

### Related Issues
- #864
- #886 

### Checklist

- [ ] Tests included
- [x] Build pass
- [x] Documentation


packages/grid_client/src/high_level/twinDeploymentHandler.ts
```diff
@@ -281,9 +281,15 @@ class TwinDeploymentHandler {
       let hru = 0;
       let sru = 0;
       let mru = 0;
-
+      let rootfs_size;
+      const disks: number[] = [];
       for (const workload of workloads) {
-        if (workload.type == WorkloadTypes.zmachine || workload.type == WorkloadTypes.zmount) {
+        if (workload.type == WorkloadTypes.zmachine) {
+          rootfs_size = workload.data["size"];
```
AhmedHanafy725, line 288: you may have more than one vm per deployment, so you can't replace it. rootfs needs to be list as well

packages/grid_client/src/high_level/twinDeploymentHandler.ts
```diff
@@ -281,9 +281,15 @@ class TwinDeploymentHandler {
       let hru = 0;
       let sru = 0;
       let mru = 0;
-
+      let rootfs_size;
+      const disks: number[] = [];
```
AhmedHanafy725, line 285:
```suggestion
      const ssdDisks: number[] = [];
```

AhmedHanafy725: we need also to take the HDD disks into account. It used for zdb deployments.

packages/grid_client/src/primitives/nodes.ts
```diff
@@ -373,6 +377,72 @@ class Nodes {
...
+    pool.sort((a, b) => b - a);
+    disks.sort((a, b) => b - a);
+
+    disks.forEach(disk => {
+      const index: number = this.findDiskIndex(pool, disk);
+      if (index === -1) {
+        throw new Error(
+          `Cannot fit the required disk with size ${
+            disk / 1024 ** 3
+          } to the disks pool of node ${nodeId}, please select another node`,
```
AhmedHanafy725, line 429: can we add the available disk pools in the error message?

packages/grid_client/src/primitives/nodes.ts
```diff
@@ -373,6 +377,72 @@ class Nodes {
...
+  /**
+   * Searches for a disk in the given DisksPool and returns the index of the first occurrence
+   * where the disk size is greater than or equal to the specified disk size.
+   *
+   * @param {number[]} DisksPool - An array containing the sizes of disks in the pool.
+   * @param {number} disk - The size of the disk to search for.
+   * @returns {number} - The index of the first disk in the DisksPool that has a size greater than or equal to the specified disk size.
+   *                    If no disk is found, it returns -1.
+   */
+  findDiskIndex(DisksPool: number[], disk: number) {
```
AhmedHanafy725, line 389:
```suggestion
  findDiskIndex(DiskPools: number[], disk: number) {
```

packages/grid_client/src/primitives/nodes.ts
```diff
@@ -373,6 +377,72 @@ class Nodes {
...
+    for (const index in DisksPool) {
+      if (DisksPool[index] >= disk) return +index;
+    }
+    return -1;
+  }
```
AhmedHanafy725, line 394: you don't have to keep track of the index, the `diskPools` are already sent by reference and you can do the subtraction here

AhmedHanafy725 (changes requested):

packages/grid_client/src/primitives/nodes.ts
```diff
@@ -373,6 +377,101 @@ class Nodes {
...
 
+  /**
+   * Finds the index of a disk with a size greater than or equal to the specified disk size in the DisksPool.
+   *
+   * @param diskPools - An array of disk sizes available in the pool.
+   * @param disk - The required disk size to fit into the pool.
+   * @returns {boolean} - True if the disk can be fitted into the pool, otherwise false.
+   */
+  findDiskIndex(diskPools: number[], disk: number): boolean {
+    for (const index in diskPools) {
```
AhmedHanafy725, line 388: why are we still looping with the index? you can use `of` instead of `in` to get the value

packages/grid_client/src/primitives/nodes.ts
```diff
@@ -373,6 +377,101 @@ class Nodes {
...
+
+  /**
+   * Fits disks into a disks pool based on their sizes.
+   *
+   * @param disks - An array of disk sizes required to be fitted into the pool.
+   * @param pool - An array representing the available disk space in the pool.
+   * @param type - The type of disks being fitted (SSD, HDD, or RootFileSystem).
+   * @throws {Error} - If the required disk cannot be fitted into the pool.
+   */
+  fitDisksInDisksPool(disks: number[], pool: number[], type: "SSD" | "HDD" | "RootFileSystem"): void {
```
AhmedHanafy725, line 414: can you make these types in an enum for ssd, hdd, and rootfs?

packages/grid_client/src/primitives/nodes.ts
```diff
@@ -74,7 +74,11 @@ interface RMBNodeCapacity {
   total: NodeResources;
   used: NodeResources;
 }
-
+export interface StoragePool {
+  type: string;
```
AhmedHanafy725, line 78: this `type` should be enum

AhmedHanafy725 (changes requested):

packages/grid_client/src/primitives/nodes.ts
0oM4R: to be able to edit the original disk pool value

AhmedHanafy725 (approved):

Issue #909, opened 2023-08-03
When user transfer TFTs to himself, the transaction fee is deducted


AlaaElattar:

khaledyoussef24:
verified on devnet now you can not transfer to yourself
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/f09093cf-4ee1-4cd8-b061-44a97f9e286e)
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/3e57b63f-3a3f-45f8-8f4f-a7fe51f97437)

PR #910, opened 2023-08-03
fix transferring to same twin or address

AlaaElattar:
Screenshot from 2023-08-03 15-13-40
Screenshot from 2023-08-03 15-13-49
https://github.com/threefoldtech/tfgrid-sdk-ts/issues/909

### DIFF
--- a/packages/dashboard/src/portal/views/Transfer.vue
+++ b/packages/dashboard/src/portal/views/Transfer.vue
@@ -36,7 +36,7 @@
                 <span class="fee">0.01 transaction fee will be deducted</span>
               </v-form>
               <TransferFormButtons
-                :isTransferValid="isTransferValidAddress"
+                :isTransferValid="isTransferValidAddress && amountByAddress > 0"
                 :loadingTransfer="loadingTransferAddress"
                 @submit="transferTFTWithAddress"
                 @clear="clearInputAddress"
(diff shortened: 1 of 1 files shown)


AhmedHanafy725 (changes requested):
it's working well, but the messages need to be updated when the reason is changed.
for example: 
- set the address to mine -> `You can't transfer to yourself`
- if the address is removed, the message will remain the same while it should be `This field is required`

AlaaElattar:
https://github.com/threefoldtech/tfgrid-sdk-ts/assets/54944491/f0afd57e-fe36-42df-b8e0-de448a61f410

packages/dashboard/src/portal/views/Transfer.vue
```diff
@@ -118,18 +122,30 @@ export default class TransferView extends Vue {
...
     this.transferTwinIdCheck();
+    this.targetErrorTwin = "";
   }
 
   async transferTwinIdCheck() {
     await this.$nextTick();
+    if (this.receptinTwinId.length === 0) {
+      this.targetErrorTwin = "This field is required";
+      return;
+    }
```
AhmedHanafy725, line 142: the rest error messages should be handled here too and removed from the form's rules?

AhmedHanafy725 (changes requested):

AhmedHanafy725 (approved):

PR #929, opened 2023-08-08
readme.md, adjusted run build command for lerna

Mik-TF:
readme.md, adjusted run build command for lerna

from [lerna's documentation](https://lerna.js.org/docs/features/run-tasks)

> npx lerna run build

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ yarn install
 ## Build
 
 ```bash
-npx lerna run build
+lerna run build
 ```
 
 ## Related Documentations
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -36,7 +36,7 @@ yarn install
 ## Build
 
 ```bash
-lerna run build
+npx lerna run build
```
AhmedHanafy725, line 39: can we use `yarn` instead of `npx` as it's the main package manager on the repo?
```suggestion
yarn lerna run build
```

Mik-TF:
Thanks @AhmedHanafy725 for the review.

I tested it and it works with yarn instead of npx.

Should be good to merge now.

AhmedHanafy725 (approved):
Thanks

Issue #491, opened 2023-06-07
Add filtering farms based on it's node


MohamedElmdary:
## Related Issues
- https://github.com/threefoldtech/tfgrid-sdk-go/issues/41

MohamedElmdary:
@xmonader currently we're using the `nodeFilters` so we might no need to update it right now

zaelgohary:
Blocked on [this](https://github.com/threefoldtech/tfgrid-sdk-go/issues/263).

mohamedamer453:
Verified on devnet.

The filtering is now done with the `/farms` endpoint according to its nodes. using the new queries from https://github.com/threefoldtech/tfgrid-sdk-go/issues/263#issuecomment-1681878077

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/101194226/2bfaa1f0-6c9f-499f-829d-d7b5efc73fd3)

PR #931, opened 2023-08-08
Farm filters playground

MohamedElmdary:
## Issues
- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/491


AhmedHanafy725 (changes requested):

packages/playground/src/utils/get_farms.ts
```diff
@@ -1,31 +1,34 @@
...
 
   let page = 1;
   while (page) {
-    const _nodes = await grid.capacity.filterNodes({ ...filters, page }).catch(() => []);
-    nodes.push(_nodes);
-    page = nodes[page - 1].length ? ++page : 0;
+    const farmsChunk = await grid.capacity.filterFarms({ ...filters, page }).catch(() => []);
+    farmsChunks.push(farmsChunk);
+    page = 0;
   }
```
AhmedHanafy725, line 22: no need for `while` loop if we gonna get only one page. Also in this case we will need to get the rest farms after finishing scrolling the first page

AhmedHanafy725 (changes requested):

AhmedHanafy725 (commented):
LGTM, but we should wait till this [issue](https://github.com/threefoldtech/tfgrid-sdk-go/issues/329) is done to make sure everything is working

AhmedHanafy725 (approved):

PR #938, opened 2023-08-08
Allow decimal values when filtering nodes in explorer

AlaaElattar:
- when converting to bytes if value still have fractions, ceil the value.
https://github.com/threefoldtech/tfgrid-sdk-ts/issues/935

### DIFF
--- a/packages/dashboard/src/explorer/store/actions.ts
+++ b/packages/dashboard/src/explorer/store/actions.ts
@@ -47,9 +47,7 @@ export default {
         key == "total_sru"
       ) {
         value *= 1024 * 1024 * 1024; // convert from gb to b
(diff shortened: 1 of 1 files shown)


packages/dashboard/src/explorer/store/actions.ts
```diff
@@ -47,8 +47,10 @@ export default {
         key == "total_sru"
       ) {
         value *= 1024 * 1024 * 1024; // convert from gb to b
+        console.log("before", value);
+        value = Math.ceil(value);
+        console.log("after", value);
```
AhmedHanafy725, line 52: please remove the debug logs

AhmedHanafy725 (approved):

Issue #928, opened 2023-08-07
Adding Public Config to Node - Unable to save.


TullysInc:
When adding a public config for a Node ID, the save button does not highlight until all fields have been entered:

![RJ Lamers](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/105203696/96517443-1b00-4f67-832e-79ccc50a0b96)

However, as per our current [manual](https://manual.grid.tf/farmers/3node_building/1_create_farm.html?highlight=Public%20Config#how-to-configure-public-ip-addresses-for-a-threefold-farm-or-a-3node) - the IPv6 address and the Domain are meant to be optional.

0oM4R:
the `domainCheck` rule is returning false if the  `domain` field is empty; 
not sure should we remove this or update the manual ? @AhmedHanafy725

A-Harby:
Verified, Devnet 2.1.0-rc8.

Based on docs, Only IPV4 is the necessary ip to be added. 
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/a287eef4-dc86-47ca-ba01-997c86afe234)
And it was updated to only request IPV4 to enable submit button
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/ae447e04-a170-4425-8e72-0b3362c082c5)

Ashishban:
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/91247662/ae77464c-ff4b-4916-a707-fdb8d88e8235)

Client continues to receive the same error.

Ashishban:
Client was on mainnet. Thanks for pointing that out.

PR #944, opened 2023-08-08
Dashboard: in Public configuration; remove required validation in domain field

0oM4R:
### Description

Domain is optional and if its value is empty string should return true to enable save button 

### Changes

return true if the domain value is empty string in `domainCheck` rule, 
now the user can add public ip without domain as mentioned in the manual 

[Screencast from 08 أغس, 2023 EEST 05:25:32 م.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/62248851/0209e300-efa1-4b9e-a800-ae63b777f16a)

### Related Issues

- #928 

### Checklist

- [ ] Tests included
- [x] Build pass
- [x] Documentation `updated with the manual`

### DIFF
--- a/packages/dashboard/src/portal/components/FarmNodesTable.vue
+++ b/packages/dashboard/src/portal/components/FarmNodesTable.vue
@@ -850,7 +850,7 @@ export default class FarmNodesTable extends Vue {
     }
   }
   domainCheck() {
-    if (this.domain !== "" && !this.validator.isURL(this.domain)) return "Invalid url format";
+    if (this.domain && !this.validator.isURL(this.domain)) return "Invalid url format";
     return true;
   }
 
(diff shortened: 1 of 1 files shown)


packages/dashboard/src/portal/components/FarmNodesTable.vue
```diff
@@ -850,7 +850,7 @@ export default class FarmNodesTable extends Vue {
     }
   }
   domainCheck() {
-    if (this.domain == "") return false;
+    if (this.domain == "") return true;
```
AhmedHanafy725, line 853: is there a point for having the condition? I think we can remove the whole line

0oM4R: I keep it to prevent url validation to return false; will test what will happen if we remove it

0oM4R: I tested it, removing this condition makes the field invalid

AhmedHanafy725 (approved):

Issue #895, opened 2023-08-01
Reload user's balance after successful transfer


AlaaElattar:
- When user transfer TFT, its balance may be updated immediately without refreshing the page.

A-Harby:
Verified Devnet, .

Balance amount is now updated after every transcation.

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/2a81451c-17fb-4dfd-87f4-a5d21d4558c0)
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/18bfdc9b-49ae-4dde-8ff9-7a6c3ebcd83f)

PR #958, opened 2023-08-09
fix updating balance issue 

AlaaElattar:
- When user performs any transaction, the balance should be updated without refreshing page.
https://github.com/threefoldtech/tfgrid-sdk-ts/issues/895

### DIFF
--- a/packages/dashboard/src/components/TfChainConnector.vue
+++ b/packages/dashboard/src/components/TfChainConnector.vue
@@ -225,7 +225,6 @@ import { Component, Vue, Watch } from "vue-property-decorator";
 import { generateKeyPair } from "web-ssh-keygen";
 
 import config from "@/portal/config";
-import { getBalance } from "@/portal/lib/balance";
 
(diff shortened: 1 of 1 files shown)


packages/dashboard/src/components/TfChainConnector.vue
```diff
@@ -225,6 +225,7 @@ import { Component, Vue, Watch } from "vue-property-decorator";
 import { generateKeyPair } from "web-ssh-keygen";
 
 import config from "@/portal/config";
+import { getBalance } from "@/portal/lib/balance";
```
Mahmoud-Emad, line 228: I see you imported this function and it's never used, so remove it please if not needed.

Mahmoud-Emad (approved):

Mahmoud-Emad (changes requested):
I see you imported the `getBalance` function and it's never used, so remove it please if not needed.

MohamedElmdary (approved):

Mahmoud-Emad (approved):

PR #930, opened 2023-08-08
Development nextcloud weblet

Mik-TF:
Added the nextcloud all-in-one weblet on the playground based on this nextcloud flist: https://github.com/threefoldtech/tf-images/tree/development_nextcloudaio/tfgrid3/nextcloud

* The nextcloud flist works on the current main net playground.
* This weblet works when I build it locally on the Playground dev net

Note: I haven't tested it with RMB + internet as I couldn't make RMB work so far on my MAC.

There might be things to improve. Please let me know.

PR #960, opened 2023-08-10
Development nextcloudv2

Mik-TF:
Continuation of #930 

The api docs were in the way so I made a new branch and set gitignore with docs/api/

I also updated the work with the new changes with disks.

@AhmedHanafy725 
@MohamedElmdary 

I tag you here since you did already help a lot in the previous PR, so you know the ongoing work.

Thanks.


Mik-TF:
This now works and all checks are passed.

Two things we might want to improve before publishing:

* Have the IPv4 address on by default, and that it cannot be set off
  * Reason: the deployment needs an IPv4 address
* Remove the micro VM drop down menu so users cannot change to "Other" and change the FList

Mik-TF:
@AhmedHanafy725 
@MohamedElmdary 

Now, the IPv4 is there by default and users can't change it. So it should always work.
Also the VM image can't be changed, it's always the Nextcloud AIO flist.

I might do some deployment testing and I invite you and others to test it if needed.

Thanks.

Mik-TF:
TODO:

- [ ] Remove option to chose "Other" as a VM Image (or remove completely the VM image menu)
- [ ] Add IPV4 in Contract table after deployment

Mik-TF:
@AhmedHanafy725 
@MohamedElmdary 

Apart from needing to promote the FList [(current tf-images issue)](https://github.com/threefoldtech/tf-images/issues/157), I think this is ready to go.

Please, if you want, have a test and let me know if I need to adjust anything.

Thanks!

Mik-TF:
@MohamedElmdary 

Do you think you could have a look?
On my end it did work perfectly when I deployed the playground locally.

I think new changes in the dev branch made the Full Clients Lint test not working.

Other than that, I would think it's ready.

Thank you.

packages/playground/src/weblets/tf_deployment_list.vue
```diff
@@ -184,6 +184,20 @@
...
+          <IconActionBtn
+            tooltip="Show Details"
+            icon="mdi-eye-outline"
+            @click="openDialog(tabs[activeTab].value, item)"
+          />
+          <IconActionBtn
+            tooltip="Open Nextcloud"
+            color="white"
+            icon="mdi-web"
+            :href="'https://' + item.value[0].publicIP?.ip.split('/')[0] + ':8080'"
```
zaelgohary, line 197: Can you change the URL to use a domain like it's been done in Caprover & Owncloud?

zaelgohary:
@Mik-TF Lint workflow is passed now. All I had to do is to re-run it.

packages/playground/src/weblets/tf_deployment_list.vue
Mik-TF: Hi @zaelgohary. First, thanks for your feedback and review :)

As I understand, the way Nextcloud AIO works is that it asks you to log to the IP address of the server first, then it gives a password and ask the user to give the domain name, then you can select how you want Nextcloud AIO installed (e.g. if you want Nextcloud Talk, etc.) then after the installations of those packages, the user can access the domain directly.

Mik-TF: There might be a way to set the domain directly. I will have a look.

Mik-TF: @zaelgohary I did the changes you asked.
Let me know if this is OK. Thanks!

packages/playground/src/weblets/tf_nextcloud.vue
```diff
@@ -0,0 +1,227 @@
...
+      <p :style="{ maxWidth: '880px' }">
+        You will need to add an A record (Host: "@", Value: &lt;VM_IP_Address&gt;) for this domain after deployment.
+        Otherwise, you won't be able to access the Nextcloud AIO instance using this domain.
+      </p>
+
+      <p class="font-weight-bold mt-4">
+        For more information, read the
+        <a
+          target="_blank"
+          href="https://www.manual.grid.tf/terraform/advanced/terraform_nextcloud_aio.html"
```
zaelgohary, line 55: I think the href should be linked to a page under the Playground directory in the info grid, not Terraform. Also, it should redirect to the domain section, not the whole page.

packages/playground/public/info/nextcloud.md
```diff
@@ -0,0 +1,5 @@
+---
+title: NextCloud
+---
+
+Nextcloud is a suite of client-server software for creating and using file hosting services. Nextcloud provides functionality similar to Dropbox, Office 365 or Google Drive. [Read the documentation](https://www.manual.grid.tf/terraform/advanced/terraform_nextcloud_aio.html) for more info.
```
zaelgohary, line 5: The URL should be changed to a playground one instead of terraform.

packages/playground/src/weblets/tf_nextcloud.vue
```diff
@@ -0,0 +1,227 @@
...
+      #="{ props }"
+    >
+      <input-tooltip tooltip="Domain name.">
+        <v-text-field label="Domain" v-model="domain" v-bind="props" />
+      </input-tooltip>
+    </input-validator>
+
+    <v-alert type="warning" variant="tonal" class="mb-6">
+      <p :style="{ maxWidth: '880px' }">
+        You will need to add an A record (Host: "@", Value: &lt;VM_IP_Address&gt;) for this domain after deployment.
```
zaelgohary, line 47: What's an A record? And can we provide a valid domain example either here or in the docs?

zaelgohary (changes requested):

packages/playground/public/info/nextcloud.md
Mik-TF: Noted. New documentation [here](https://github.com/threefoldtech/info_grid/pull/240).

packages/playground/src/weblets/tf_nextcloud.vue
Mik-TF: Added a valid domain example in docs and in weblet frontend. Also gave the official DNS A Record term.

packages/playground/src/weblets/tf_nextcloud.vue
Mik-TF: Noted. New documentation [here](https://github.com/threefoldtech/info_grid/pull/240).

Mik-TF:
@zaelgohary 
@AhmedHanafy725 
@MohamedElmdary 

I did all the changes requested.

Mik-TF:
## Additional note for branch development_ncgateway

I am also testing the same nextcloud weblet, but this time with the possibility of setting a gateway and the gateway domain. Much like it is possible with CasperLabs. I took the CasperLabs.vue template for this other branch.

The gateway version is available in this branch: https://github.com/threefoldtech/tfgrid-sdk-ts/tree/development_ncgateway
Perhaps it would be possible to set it correctly. But I haven't succeeded yet (started not long ago).

So far here is what works with this other branch (ncgateway):

* ipv4 and custom domain
  * it works on both ipv4:8080 and domain:8443
* ipv4 and gateway domain
  *  it works on ipv4:8080 but not gateway_domain:8443
* gateway and gateway domain
   * doesn't work in both case so far
   
***

This was requested by some TF users that is why I try with gateway + gateway domains.

zaelgohary:
@Mik-TF Deploy button is not disabled when there is no node chosen.

image

Mik-TF:
@zaelgohary 
Thanks for the feedback. I set the deploy button as false by default and set the form-validator so it checks all the deployment's parameters (as the standard with other playground weblets).

Mik-TF:
Thanks @zaelgohary for all your help.

Thanks also to @AhmedHanafy725 @MohamedElmdary

Issue #922, opened 2023-08-06
Profile Manager: `Generate Account` sometimes gets stuck at loading the `Terms and conditions` page and keeps loading forever


mohamedamer453:
### Description

The `Generate Account` button sometimes gets stuck at loading the `Terms and conditions` page and keeps loading forever with no errors in the console.

[Untitled_ Aug 6, 2023 3_14 PM.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/101194226/2f2ca368-02fb-47c9-b6db-f974d308fb2a)

### Steps to reproduce

> This issue is not always reproducible out of 4 tries it got stuck once.

- Navigate to https://dashboard.qa.grid.tf/
- In the `Connect your wallet` tab, click on `Generate Account`

> Another scenario that will always reproduce this issue is generating an account and accepting the terms and conditions then clearing the mnemonic field and clicking on `Generate Account` again.

[Untitled_ Aug 6, 2023 3_29 PM.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/101194226/5b4fc0c7-da14-4d83-8365-5a86bda6e8f8)

### Environment

- Network: `qanet`
- Dashboard Version: `2.1.0-rc5`

zaelgohary:
Still investigating.

mohamedamer453:
Verified on devnet.

I've tried generating multiple accounts and it worked every time.


> Another scenario that will always reproduce this issue is generating an account and accepting the terms and conditions then clearing the mnemonic field and clicking on `Generate Account` again.

this scenario was also fixed and now the terms and conditions page appear normally.

[TF Playground _ Full Virtual Machine (9).webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/101194226/64d0b9a8-5d9a-4d21-b241-dc1aadca77c1)

PR #970, opened 2023-08-13
Fix terms & conditions loading

zaelgohary:
### Description

change loading terms outside the iframe

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/922

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/packages/dashboard/src/components/TfChainConnector.vue
+++ b/packages/dashboard/src/components/TfChainConnector.vue
@@ -143,7 +143,7 @@
                       :loading="generatingAccount"
                       class="mt-2 ml-4"
                       text
-                      @click="openAcceptTerms = termsLoading = true"
+                      @click="openTerms"
                     >
                       Generate Account
                     </v-btn>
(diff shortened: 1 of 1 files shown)


packages/dashboard/src/components/TfChainConnector.vue
```diff
@@ -465,6 +464,13 @@ export default class TfChainConnector extends Vue {
   public openAcceptTerms = false;
   public generatingAccount = false;
   public termsLoading = false;
+
+  public openTerms() {
+    this.openAcceptTerms = this.termsLoading = true;
+    setTimeout(() => {
+      this.termsLoading = false;
+    }, 2000);
```
Mahmoud-Emad, line 472: Why we used this timeout function?

zaelgohary: The iframe visibility is determined on termsLoading boolean, which changes once it gets loaded. This happens only once and after that, it gets stuck in the loading phase. I set the boolean outside the iframe itself and the issue was fixed.

zaelgohary: The timeout function for setting a delay so the progress circle would be shown for 2 seconds. The iframe itself doesn't need loading as it's already included in the dom but I didn't want to change the flow.

Mahmoud-Emad (approved):

Issue #919, opened 2023-08-06
The profile manager fields should be cleared after closing it or logging in/out


mohamedamer453:
### Description

The profile manager fields should be cleared after closing it or logging in/out.

[Untitled_ Aug 6, 2023 3_08 PM.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/101194226/775d4f79-f3b0-4222-a75e-3be2a02bfb67)

A-Harby:
The fields are cleared, but only for the tab that was selected before closing; the other tab isn't affected.
[Untitled_ Aug 13 2023 11_51 AM.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/4afbdf38-5925-4f54-838c-40bf001fa461)

mohamedamer453:
Verified on devnet.

All fields are now cleared after closing the profile manager or logging in/out.

[TF DASHBOARD (8).webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/101194226/7ab8b065-4c72-4669-8690-50cb3052039d)

PR #971, opened 2023-08-13
Fix profile input resetting

zaelgohary:
### Changes

- Change the condition to v-show instead of v-if so the forms wouldn't be destroyed if tab is changed

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/919

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/packages/dashboard/src/components/TfChainConnector.vue
+++ b/packages/dashboard/src/components/TfChainConnector.vue
@@ -431,6 +431,10 @@ export default class TfChainConnector extends Vue {
       name: "accounts",
       path: `/`,
     });
+    this.loginPassword = "";
+    this.password = "";
+    this.password2 = "";
+    this.mnemonic = "";
   }
 
   public close() {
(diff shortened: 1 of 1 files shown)


Mahmoud-Emad (commented):
I think the change you did doesn't fix the issue

zaelgohary:
@Mahmoud-Emad I think it does. :D
[screen-capture.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/40770501/e83054ca-76eb-4225-a302-878dc2036dff)

Omarabdul3ziz:
Hi, what i noticed is `close` button does something different than clicking outside the profile component.
if you click close button it do clear the fields but clicking outside the component will not

Mahmoud-Emad:
@Omarabdul3ziz 
Exactly :)

Omarabdul3ziz:
yes this works.
a small final comment after you do logout, fields are still there only if you close the component it gets cleared. could we make logout clear the fields also or what do you think?

zaelgohary:
@Omarabdul3ziz I couldn't use clearfields method as the refs get destroyed so I reset the fields one by one.

Omarabdul3ziz (approved):

Issue #979, opened 2023-08-14
Balance doesn't updated after transaction and locked TFT disappear 


A-Harby:
### Description

After transfering TFT, Locked TFT amount disapear.
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/5c5b9122-1ca0-4c41-bd04-a5ae9ffa87a2)
[Untitled_ Aug 14, 2023 2_28 PM.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/19ca5d0c-e628-41b5-b8cf-cb913c374ca3)

And the twin that TFT transferred to, doesn't updated its balance without a refresh.
In the video the twin should have received 1 TFT, but I had to logout and relogin or refresh to updated my balance.
[Untitled_ Aug 14 2023 2_11 PM.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/8c2b8a1e-2702-4f9c-ba9f-cd142b6fcc60)

mohamedamer453:
Verified on devnet.

The balance now gets updated after a transfer and the locked tft doesn't disappear.

[TF DASHBOARD (11).webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/101194226/9db34044-749a-4463-a440-ea4253c67188)

PR #991, opened 2023-08-15
Fix disappearance of locked TFTs

AlaaElattar:
- Fix disappearance of locked TFTs after performing any transaction.
- Fix format of TFTs that changes after any transaction to stay the same without refreshing.
- The balance of reciever will be updated every 2 min without refreshing or logging out to see recieved TFTs.
- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/979

### DIFF
--- a/packages/dashboard/src/components/TfChainConnector.vue
+++ b/packages/dashboard/src/components/TfChainConnector.vue
@@ -373,7 +373,6 @@ export default class TfChainConnector extends Vue {
   stopBalanceInterval() {
     if (this.interval) {
       clearInterval(this.interval);
-      this.interval = null;
     }
   }
 
--- a/packages/dashboard/src/components/TfChainConnector.vue
+++ b/packages/dashboard/src/components/TfChainConnector.vue
@@ -316,6 +316,8 @@ export default class TfChainConnector extends Vue {
   public validatingMnemonic = false;
   public mnemonicError: null | string = null;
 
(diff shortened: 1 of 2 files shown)


packages/dashboard/src/components/TfChainConnector.vue
```diff
@@ -350,18 +352,47 @@ export default class TfChainConnector extends Vue {
...
+    }
+  }
+
+  /* Load Balance every 2 min */
+
+  startBalanceInterval() {
+    this.interval = setInterval(this.loadBalanceData, 2 * 60 * 1000);
+  }
+
+  stopBalanceInterval() {
```
zaelgohary, line 373: I don't think this function is needed.

packages/dashboard/src/components/TfChainConnector.vue
```diff
@@ -350,18 +352,47 @@ export default class TfChainConnector extends Vue {
...
+      this.balance = await loadBalance(grid);
     }
   }
 
+  mounted() {
+    this.startBalanceInterval();
+  }
+
+  beforeDestroy() {
+    this.stopBalanceInterval();
```
zaelgohary, line 393: instead of using stropBalanceInterval, you can clear the interval directly.
```suggestion
   clearInterval(this.interval);
```

zaelgohary (changes requested):

zaelgohary (approved):

Issue #839, opened 2023-07-24
Twins with farms but no nodes are notified to vote, although voting is only available to farms with nodes.


A-Harby:
### Description

While voting is only available to farms containing nodes, twins with farms that have no nodes receive a notification to vote.
 
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/880667e0-a290-4593-86a1-171b6da9824f)
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/c7c04883-0354-4481-aede-bd2b06344698)

ramezsaeed:
USER MUST HAVE ONLINE NODES IN HIS FARM SO HE CAN VOTE.

zaelgohary:
Added user nodes to the store so that it would be accessible by all portal components. However, the voting dialogue shows right after the logging and before the state gets updated. I think the issue would be fixed if we created a grid proxy API that returns nodes by user twin ID.

zaelgohary:
Added a condition to check if the user has nodes before showing the voting dialogue.

A-Harby:
Verified, Devnet 2.1.0-rc11 .
Now voting notification only appears to twins with nodes in it.
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/eccba081-c767-4c41-bbea-079cc2ffe279)

PR #994, opened 2023-08-15
Fix voting dialogue

zaelgohary:
### Description

- Condition the dialogue to show only if user nodes exist
- Add getUserNodes

### Changes

List of changes this PR includes

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/839

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/packages/dashboard/src/portal/components/VotePopup.vue
+++ b/packages/dashboard/src/portal/components/VotePopup.vue
@@ -1,4 +1,4 @@
-<template>
+<template v-if="nodes">
   <v-dialog v-model="openProposalDialog" persistent max-width="350">
     <template v-slot:default="dialog">
       <v-card>
@@ -20,6 +20,9 @@
 <script lang="ts">
 import { Component, Vue } from "vue-property-decorator";
 
+import config from "../config";
+import { getFarm } from "../lib/farms";
+
 @Component({
   name: "VotePopup",
 })
@@ -27,10 +30,17 @@ export default class VotePopup extends Vue {
   openProposalDialog = false;
   proposals = 0;
   voteMsg = "You have a pending proposal to vote!";
-
+  $api: any;
+  farms: any = [];
+  farmsIds: any = [];
+  nodes: any = [];
   private __destroyed = false;
 
   async mounted() {
+    this.farms = await getFarm(this.$api, this.$store.state.credentials.twin.id);
+    this.farmsIds = this.farms?.map((farm: any) => farm.id);
+    this.nodes = await this.getUserNodes(this.farmsIds);
+
     await this.$store.dispatch("portal/getProposal", this.$store.state.credentials.twin.id);
 
     this.proposals = this.$store.state.portal.proposals;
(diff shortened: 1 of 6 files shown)


packages/dashboard/src/portal/components/VotePopup.vue
```diff
@@ -20,22 +20,32 @@
...
 export default class VotePopup extends Vue {
   openProposalDialog = false;
   proposals = 0;
   voteMsg = "You have a pending proposal to vote!";
-
-  private __destroied = false;
+  $api: any;
+  farms: any = [];
+  farmsIds: any = [];
+  nodes: any = [];
```
Mahmoud-Emad, line 36: can we use typing instead of any?

zaelgohary: edited.

packages/dashboard/src/portal/components/VotePopup.vue
```diff
@@ -44,6 +55,13 @@ export default class VotePopup extends Vue {
     this.openProposalDialog = true;
   }
 
+  async getUserNodes(farmIDs: any[]): Promise<any> {
+    const url = `${config.gridproxyUrl}/nodes?ret_count=true&page=1&size=10&farm_ids=${farmIDs}`;
```
AhmedHanafy725, line 59: we need also to make sure these nodes are online
```suggestion
    const url = `${config.gridproxyUrl}/nodes?status=up&ret_count=true&page=1&size=10&farm_ids=${farmIDs}`;
```

packages/dashboard/src/portal/components/VotePopup.vue
```diff
@@ -44,6 +55,13 @@ export default class VotePopup extends Vue {
     this.openProposalDialog = true;
   }
 
+  async getUserNodes(farmIDs: any[]): Promise<any> {
```
AhmedHanafy725, line 58: we don't send the twin id to this method so I think the method name is invalid. maybe we can call it `getFarmsNodes`

AhmedHanafy725, line 58: we already know the type of the farmIDs
```suggestion
  async getUserNodes(farmIDs: number[]): Promise<any> {
```

packages/dashboard/src/portal/components/VotePopup.vue
```diff
@@ -1,4 +1,4 @@
-<template>
+<template v-if="nodes">
```
AhmedHanafy725, line 1: not sure if the `v-if` works with list length or not, can you make sure?

packages/dashboard/src/portal/components/VotePopup.vue
AhmedHanafy725, line 59: the farmIDs here in the query are a list while the gridproxy is waiting for the farmIDs just numbers separated by a comma.

packages/dashboard/src/portal/components/VotePopup.vue
AhmedHanafy725, line 58: we already know the return type, so we shouldn't use `any`

packages/dashboard/src/portal/components/VotePopup.vue
```diff
@@ -20,22 +20,33 @@
...
 @Component({
   name: "VotePopup",
 })
 export default class VotePopup extends Vue {
   openProposalDialog = false;
   proposals = 0;
   voteMsg = "You have a pending proposal to vote!";
-
-  private __destroied = false;
+  $api: any;
```
AhmedHanafy725, line 34: the api type here is `ApiPromise` you can import it

`import { ApiPromise } from "@polkadot/api"`

packages/dashboard/src/portal/components/VotePopup.vue
zaelgohary: Checked & it works.

AhmedHanafy725 (approved):

Issue #848, opened 2023-07-26
caprover: use a better generated leader name


xmonader:
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/64129/1f82be4f-3213-4dd2-a099-f08fc7809a7e)

khaledyoussef24:
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/e14e10e6-78af-4a56-8676-5f48620b4451)
now generated names will use cr prefix then 10 random numbers and letters
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/4fe15b5b-6874-4487-b8fa-637af77bf621)

statically : the name generation function changed

PR #1001, opened 2023-08-16
Better generated leader name for caprover

AlaaElattar:
- Improve function of generateRandomName
- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/848
- Examples of new generated names of length 5 : 

image

Screenshot from 2023-08-16 14-11-26


packages/playground/src/utils/strings.ts
```diff
@@ -9,11 +9,21 @@ export interface GenerateNameOptions {
   suffix?: string;
 }
 export function generateName(length = 12, options?: GenerateNameOptions): string {
```
zaelgohary, line 11: I think the changes done here are not needed as we only need to adjust the length where we call the function.

zaelgohary (changes requested):
The same generateName function is being called in all solutions, so I think it'd be better if we adjusted the length in all of them to match what's been done in Caprover.

packages/playground/src/utils/strings.ts
```diff
@@ -9,11 +9,21 @@ export interface GenerateNameOptions {
...
+  const remainingLength = length - prefix.length - suffix.length;
+
   return (
-    (options?.prefix ?? "") + randomChoice(chars) + generateString(chars + nums, length - 1) + (options?.suffix ?? "")
+    (options?.prefix ?? "") +
+    randomChoice(consonants) +
+    generateString(vowels + consonants, Math.ceil((remainingLength - 2) / 2)) +
+    generateString(consonants, Math.floor(remainingLength / 3)) +
+    generateString(nums, Math.floor(remainingLength / 6)) +
+    (options?.suffix ?? "")
```
zaelgohary, line 26: I think the changes done here are not needed as we only need to adjust the length where we call the function.

zaelgohary (approved):

Issue #1009, opened 2023-08-18
Can't activate existing account using new login flow (devnet)


scottyeager:
I tried connecting to the dashboard using an existing seed phrase on devnet. After hitting the *Activate Account* button, I see the error `Couldn't find a user for the provided mnemonic on dev network.`

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/17682591/c36c8c03-1300-4ad9-b006-7e781f5bdf82)

Mahmoud-Emad:
The problem arose due to the resolution of the issue outlined in https://github.com/threefoldtech/tfgrid-sdk-ts/issues/992. Although the fix for this problem should have been resolved in my local branch, I am currently exploring methods to apply the fix to address the issue mentioned in https://github.com/threefoldtech/tfgrid-sdk-ts/issues/992 as well.

Mahmoud-Emad:
The issue should be fxed in https://github.com/threefoldtech/tfgrid-sdk-ts/pull/1031/

khaledyoussef24:
verified on devnet fixed now
[Screencast from 08-24-2023 11:02:42 AM.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/57712e2a-6f97-4d96-b9b7-767ecccf9bf6)

Issue #992, opened 2023-08-15
Socket Error: read ETIMEDOUT disconnected from wss://tfchain.grid.tf/ws


costarassco:
### Description

I have a service allowing users to connect to the Grid using the GridClient.
The code is provided in the "Steps to reproduce" section

The connection with the grid client can be a quite long and sometimes in several hours,
I do get such an error: "API-WS: disconnected from wss://tfchain.grid.tf/ws: 1006:: Socket Error: read ETIMEDOUT"
All the details can be found in the "Logs" section
 
I have put everything what's possible in the try-catch block, but I still can't handle this error, which destroying my node connection

Can you please help me with this? I can't use the grid client with such erros at all :( 

I am using 2.0.0-rc6

EDIT:
I also use this functionality btw https://github.com/threefoldtech/grid_weblets/blob/development/src/utils/findNodes.ts

### Steps to reproduce

```
import { GridClient, NetworkEnv } from "@threefold/grid_client";

const options = {
  network: NetworkEnv.main,
  mnemonic,
  storeSecret: "coolSecretPhrase",
};

const gridClient = new GridClient(options);
await gridClient.connect();
```

### Logs/Alerts

```
API-WS: disconnected from wss://tfchain.grid.tf/ws: 1006:: Socket Error: read ETIMEDOUT

/my_cool_folder/node_modules/@threefold/tfchain_client/node_modules/@polkadot/api/cjs/base/index.js:45
    return this._rpcCore.connect();
                         ^
TypeError: Cannot read properties of undefined (reading 'connect')
    at EventEmitter.connect (/my_cool_folder/node_modules/@threefold/tfchain_client/node_modules/@polkadot/api/cjs/base/index.js:45:26)
    at EventEmitter.emit (/my_cool_folder/node_modules/@threefold/tfchain_client/node_modules/eventemitter3/index.js:180:35)
    at ApiPromise.emit (/my_cool_folder/node_modules/@threefold/tfchain_client/node_modules/@polkadot/api/cjs/base/Events.js:22:31)
    at ApiPromise._onProviderDisconnect2 (/my_cool_folder/node_modules/@threefold/tfchain_client/node_modules/@polkadot/api/cjs/base/Init.js:449:8)
    at EventEmitter.<anonymous> (/my_cool_folder/node_modules/@threefold/tfchain_client/node_modules/@polkadot/api/cjs/base/Init.js:103:148)
[...]

PR #1031, opened 2023-08-21
Development fix activation, disconnecting issues 

Mahmoud-Emad:
### Description

Fix the activation issue by creating a new provider when reconnecting to the grid, and also delete the old API connection.

Use the script to test the grid connection:
This script creates 2 instances of the grid client, then after 10 minutes, it performs a query on both instances.

### Steps:
- Connect to the grid
- Close the internet connection
- After 9 minutes, open the internet connection again
  
In the tenth minute the grid should automatically reconnect and request the twin ID.

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1009
- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/992

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/packages/grid_client/scripts/connection.ts
+++ b/packages/grid_client/scripts/connection.ts
@@ -12,20 +12,11 @@
 import { GridClient, NetworkEnv } from "../src";
 
 // Replace with actual mnemonics
-const clientMnemonic1 = "actual reveal dish guilt inner film scheme between lonely myself material replace";
-const clientMnemonic2 = "actual reveal dish guilt inner film scheme between lonely myself material replace";
+const clientMnemonic1 = "your_mnemonic_here";
+const clientMnemonic2 = "your_mnemonic_here";
 
 // Choose the desired network (main, test, qa, dev)
 const NETWORK = NetworkEnv.dev;
-let sessionTime = 0;
-// Function to log the current minute along with the client's name
-function logCurrentMinute(time: number) {
-  if (time < 9) {
-    console.log(`${time}: Please disconnect your internet connection.`);
-  } else {
-    console.log(`${time}: Please re-connect your internet connection.`);
-  }
-}
 
 // Client 1 logic
 async function newClient(mnemonic: string) {
@@ -38,13 +29,11 @@ async function newClient(mnemonic: string) {
     const gridClient = new GridClient(options);
     await gridClient.connect();
 
-    // Run the task every 10 minutes
-    setInterval(async () => {
+    setTimeout(async () => {
       console.log(`Requist the grid to get the twin id.`);
       const twinID = await gridClient.twins.get_my_twin_id();
       console.log(`Twin id: ${twinID}`);
-      sessionTime = 0;
-    }, 10 * 60 * 1000); // Run every 10 minutes (10 * 60 * 1000 milliseconds)
+    }, 10 * 60 * 1000);
   } catch (error) {
     console.error("Error connecting to Grid:", error);
   }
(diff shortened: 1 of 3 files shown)


packages/grid_client/scripts/connection.ts
```diff
@@ -0,0 +1,90 @@
...
+    3 - After 9 minutes, open the internet connection again
+  The grid should automatically reconnect and request the twin ID.
+*/
+
+import { GridClient, NetworkEnv } from "../src";
+
+// Replace with actual mnemonics
+const clientMnemonic1 = "your_mnemonic_here";
+const clientMnemonic2 = "your_mnemonic_here";
+const storeSecret = "store_secret";
```
AhmedHanafy725, line 17: no need for store secret here

packages/grid_client/scripts/connection.ts
```diff
@@ -0,0 +1,90 @@
...
+
+// Choose the desired network (main, test, qa, dev)
+const NETWORK = NetworkEnv.dev;
+
+// Function to log the current minute along with the client's name
+function logCurrentMinute(clientName) {
+  const now = new Date();
+  console.log(`${clientName} running in minute ${now.getMinutes()}`);
```
AhmedHanafy725, line 25: logging the minute only looks strange, I think you should show the date in hour:min:sec

packages/grid_client/scripts/connection.ts
```diff
@@ -0,0 +1,90 @@
...
+    storeSecret: storeSecret,
+  };
+  try {
+    const gridClient = new GridClient(options);
+    await gridClient.connect();
+
+    // Log the current minute every minute
+    setInterval(() => {
+      logCurrentMinute("Client 1");
+    }, 60000); // Log every minute (60000 milliseconds)
```
AhmedHanafy725, line 42: do we need to do that? it says nothing imo. we need to know start of the connection only I think

packages/grid_client/scripts/connection.ts
Mahmoud-Emad: Will remove the whole date and make count up instead

packages/grid_client/scripts/connection.ts
```diff
@@ -0,0 +1,64 @@
...
+// Replace with actual mnemonics
+const clientMnemonic1 = "your_mnemonic_here";
+const clientMnemonic2 = "your_mnemonic_here";
+
+// Choose the desired network (main, test, qa, dev)
+const NETWORK = NetworkEnv.dev;
+
+// Function to log the current minute along with the client's name
+function logCurrentMinute(clientName: string, time = 1) {
+  console.log(`${clientName} running count ${time}`);
```
AhmedHanafy725, line 23: what are you counting on here? the log will log like 
`Client: 1 running count 1`

packages/grid_client/scripts/connection.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  let time = 1;
+
+  try {
+    const gridClient = new GridClient(options);
+    await gridClient.connect();
+
+    // Log the current minute every minute
+    setInterval(() => {
+      time += 1; // script starts at the first minute, then after one minute will log the second..etc
```
AhmedHanafy725, line 41: you already started the variable with 1 and after increasing it here, the first log will be 2 mins while it should be 1.

packages/grid_client/scripts/connection.ts
```diff
@@ -0,0 +1,64 @@
...
+      console.log(`Client ${clientNumber} twin id: ${twinID}`);
+    }, 10 * 60 * 1000); // Run every 10 minutes (10 * 60 * 1000 milliseconds)
+  } catch (error) {
+    console.error("Error connecting to Grid:", error);
+  }
+}
+
+// Main function
+async function main() {
+  logCurrentMinute("Script start");
```
AhmedHanafy725, line 58: do we need to use this logging method to log this message?

packages/grid_client/scripts/connection.ts
Mahmoud-Emad: No will log 1 if I don't pass the time, otherways will log the actual time, I want to set this as optional because I don't wanna send it at the beginning of the script.

AhmedHanafy725: I mean here that the message doesn't tell me what are we counting

Mahmoud-Emad: updated to be like that `Client 1 is in its 3 execution cycle.`

AhmedHanafy725 (approved):

Issue #1014, opened 2023-08-20
Add Dashboard Twin page


zaelgohary:
- Implement Twin in Vue3

PR #1051, opened 2023-08-27
Add Edit button in twin details page.

AlaaElattar:
- The button won't appear unless the relay doesn't match target one.
- If pk only doesn't match, it will be updated silently without showing the button.
- Then, user can update relay and pk will be updated internally.
- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1014

- If there's no edit button
image

- If there's edit button
Screenshot from 2023-08-27 16-14-47
image

### DIFF
--- a/packages/playground/src/portal/twin_view.vue
+++ b/packages/playground/src/portal/twin_view.vue
@@ -83,13 +83,17 @@ function editTwin() {
 }
 
 async function UpdateRelay() {
-  const grid = await getGrid(profileManager.profile!);
-  const twins = new Twins(grid!.tfclient);
-  const newPk = await generatePublicKey(profileManager.profile!.mnemonic);
-  await twins.update({ pk: newPk, relay: relay.value });
-  profileManager.updateRelay(window.env.RELAY_DOMAIN);
-  profileManager.updatePk(newPk);
-  updateRelay.value = false;
+  try {
+    const grid = await getGrid(profileManager.profile!);
+    const twins = new Twins(grid!.tfclient);
+    const newPk = await generatePublicKey(profileManager.profile!.mnemonic);
+    await twins.update({ pk: newPk, relay: relay.value });
+    profileManager.updateRelay(window.env.RELAY_DOMAIN);
+    profileManager.updatePk(newPk);
+    updateRelay.value = false;
+  } catch (e) {
+    console.log("could not update relay, Error: ", e);
+  }
 }
 </script>
 
--- a/packages/playground/src/portal/twin_view.vue
+++ b/packages/playground/src/portal/twin_view.vue
@@ -49,7 +49,7 @@
 </template>
 
 <script lang="ts" setup>
-import { GridClient } from "@threefold/grid_client";
+import { generatePublicKey } from "@threefold/rmb_direct_client";
 import { onMounted, ref } from "vue";
 
 import { useProfileManager } from "../stores";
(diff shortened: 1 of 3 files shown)


packages/playground/src/portal/twin_view.vue
```diff
@@ -25,15 +40,65 @@
...
 
 const profileManager = useProfileManager();
+
+const editingTwin = ref(false);
+const relay = ref(profileManager.profile?.relay || "");
+const updateRelay = ref(false);
+
+onMounted(validateEdit);
+async function validateEdit() {
+  const client = new GridClient({ mnemonic: profileManager.profile!.mnemonic, network: window.env.NETWORK });
```
AhmedHanafy725, line 68: do we use the client here?

packages/playground/src/portal/twin_view.vue
```diff
@@ -25,15 +40,65 @@
...
+  console.log("editing twin");
+  editingTwin.value = true;
+}
+
+async function UpdateRelay() {
+  try {
+    const grid = await getGrid(profileManager.profile!);
+    const twins = new Twins(grid!.tfclient);
+    const newPk = await generatePublicKey(profileManager.profile!.mnemonic);
+    await twins.update({ pk: newPk, relay: relay.value });
```
AhmedHanafy725, line 94: we can use the twin module here on the grid client if we already have it.
`grid.twins.update` and it doesn't need pk, it's already updating it

packages/playground/src/portal/twin_view.vue
```diff
@@ -25,15 +40,65 @@
...
+    profileManager.updateRelay(window.env.RELAY_DOMAIN);
+    profileManager.updatePk(newPk);
+    updateRelay.value = false;
+  } catch (e) {
+    console.log("could not update relay or pk, Error: ", e);
```
AhmedHanafy725, line 99: in case the update fails, we should show the error to the user

packages/playground/src/portal/twin_view.vue
```diff
@@ -25,15 +40,62 @@
...
+  console.log("editing twin");
+  editingTwin.value = true;
+}
+
+async function UpdateRelay() {
+  try {
+    const grid = await getGrid(profileManager.profile!);
+    const newPk = await generatePublicKey(profileManager.profile!.mnemonic);
+    await grid?.twins.update({ relay: relay.value });
+    profileManager.updateRelay(window.env.RELAY_DOMAIN);
```
AhmedHanafy725, line 91: why the profile's relay and the twin's relay are not updated with the same value?

packages/playground/src/portal/twin_view.vue
```diff
@@ -25,15 +40,62 @@
             </v-col>
           </v-row>
         </v-list>
+        <v-card-actions v-if="updateRelay" class="justify-end">
+          <v-btn color="primary" class="custom-button" @click="editTwin">Edit</v-btn>
+        </v-card-actions>
```
zaelgohary, line 45:
```suggestion
        <v-card-actions v-if="updateRelay" class="justify-end mx-4 mb-4">
          <v-btn color="primary" class="custom-button bg-primary" @click="editTwin">Edit</v-btn>
        </v-card-actions>
```

packages/playground/src/portal/twin_view.vue
```diff
@@ -25,15 +40,62 @@
...
+
+const editingTwin = ref(false);
+const relay = ref(profileManager.profile?.relay || "");
+const updateRelay = ref(false);
+const errorMsg = ref("");
+
+onMounted(validateEdit);
+async function validateEdit() {
+  try {
+    const pk = await generatePublicKey(profileManager.profile!.mnemonic);
```
zaelgohary, line 68: Instead of generating a new PK, can we get the existing one from the client? As the value of the generated PK might not match the one on the grid.

packages/playground/src/portal/twin_view.vue
```diff
@@ -25,15 +40,62 @@
...
+
+function editTwin() {
+  console.log("editing twin");
+  editingTwin.value = true;
+}
+
+async function UpdateRelay() {
+  try {
+    const grid = await getGrid(profileManager.profile!);
+    const newPk = await generatePublicKey(profileManager.profile!.mnemonic);
```
zaelgohary, line 89: Also here, we can get the existing pk and store it in profile manager

zaelgohary (changes requested):

zaelgohary (approved):

Issue #1024, opened 2023-08-20
Implement Calculator in Vue3


zaelgohary:

maayarosama:
Working on refactoring the resource pricing calculator, and still working on the simulator

maayarosama:
Pr for resource pricing is ready and waiting for review. Currently resumed working on the simulator

khaledyoussef24:
simulator related for this issue or got it's separate issue ?

AhmedHanafy725:
> simulator related for this issue or got it's separate issue ?

it has a separate issue #1135

khaledyoussef24:
calculator verified and working now

![Image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/4c991bdb-985d-47a0-92c7-2b83d90a473a)

PR #1078, opened 2023-09-07
added resource pricing

maayarosama:
### Description

Added Resource pricing calculator
Screenshot from 2023-09-07 11-54-32
Screenshot from 2023-09-07 11-55-22

### Changes

Added a resource pricing view page, calculator.ts file and a connect.ts file
### Related Issues

#1024
### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation


ehab-hassan:
enhancement for UI and gold, silver, and Bronze color
image

amiraabouhadid (changes requested):
Don't see UI changes reflected yet. Please make PR into draft if not ready. 
image

maayarosama:
> Don't see UI changes reflected yet. Please make PR into draft if not ready. image

UI colors change according to the package, the default package is the same color as the one in your screenshot, however colors are reflected when the package change as follows
Screenshot from 2023-09-17 11-32-55

packages/playground/src/plugins/vuetify.ts
```diff
@@ -20,6 +20,7 @@ const vuetify = createVuetify({
         colors: {
           primary: "#7de3c8",
           secondary: "#14A1B8",
+          calcprimary: "#1AA18F",
```
MohamedElmdary, line 23: Please rename color to a more generic  name
if it won't be used anywhere else so delete it from vars and use it directly in component

MohamedElmdary:
Please consider handling the possible error because right now if the grid failed to initialize the Mount the entire component won't work untill the user refresh the page

packages/playground/src/plugins/vuetify.ts
maayarosama: It will be used somewhere else so i'll leave it in the vars but change it to have a more generic name

MohamedElmdary (approved):

Issue #1062, opened 2023-08-30
[WordPress] The password generated shouldn't contain space


AhmedHanafy725:
### Description

If the password has a space the deployment will be okay, but the solution itself will be broken.

### Steps to reproduce

- Go to WordPress solution.
- Update the password to contain a space.
- Deploy the instance.

### Logs/Alerts
The password was `8 D!eomkm0RX`.

zinit logs for wp process.

```
[-] wp: Error: Too many positional arguments: D!eomkm0RX
[-] wp: + chmod +x /usr/local/bin/wp_entrypoint.sh
[-] wp: + cd /var/www/html
[-] wp: + a2enmod headers
[+] wp: Module headers already enabled
[-] wp: + /usr/local/bin/wp_entrypoint.sh apache2-foreground
```

AhmedHanafy725:
The generated password shouldn't contain space, and the validation shouldn't allow that

khaledyoussef24:
verified 
![Screenshot from 2023-09-24 11-47-40](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/66d00c53-dc82-49b8-b524-d12dc7bb2612)

PR #1085, opened 2023-09-11
Fix password validation

zaelgohary:
### Description

Fix password validation in all solutions

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1062

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/packages/playground/src/weblets/profile_manager.vue
+++ b/packages/playground/src/weblets/profile_manager.vue
@@ -185,6 +185,7 @@
                   :rules="[
                     validators.required('Password is required.'),
                     validators.minLength('Password must be at least 6 characters.', 6),
+                    validators.isAlphanumeric('Password should consist of letters and numbers only.'),
                     validatePassword,
                   ]"
                   #="{ props: validationProps }"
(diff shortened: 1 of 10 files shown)


maayarosama:
Password field doesn't accept special characters which shouldn't be the case
password

packages/playground/src/weblets/profile_manager.vue
```diff
@@ -185,6 +185,7 @@
                   :rules="[
                     validators.required('Password is required.'),
                     validators.minLength('Password must be at least 6 characters.', 6),
+                    validators.isAlphanumeric('Password should consist of letters and numbers only.'),
```
AhmedHanafy725, line 188: not sure if we should change the rules of the password of the wallet connector as it's frequently used by the user and that may prevent the login for them

zaelgohary: I reverted the previous changes and verified on white spaces only. should I remove the changes from the profile manager?

zaelgohary:
@maayarosama I added a new validation pattern which checks if the string contains a space or not.

packages/playground/src/weblets/profile_manager.vue
AhmedHanafy725: yes, it is used only for encrypting the mnemonic in local storage and works fine with spaces

zaelgohary: Done.

AhmedHanafy725 (approved):

Issue #1070, opened 2023-09-03
Add Get TFT button


amiraabouhadid:
### Description

New to add a get TFT button in the new Vue3 portal integrated in playground

AlaaElattar:
- For now it's fixed by connecting again to client before transferring with Alice account.

khaledyoussef24:
verified 
[Screencast from 10-02-2023 01:32:53 PM.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/303a7fe8-153b-45a7-862b-f17a768bd0eb)

PR #1086, opened 2023-09-11
Add Get TFT button in new Dashboard

AlaaElattar:
- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1070

Screenshot from 2023-09-12 12-15-34

Screenshot from 2023-09-12 12-15-51


amiraabouhadid (changes requested):
My bad needed to run yarn lerna run build first.

amiraabouhadid (approved):

packages/grid_client/src/clients/tf-grid/balances.ts
```diff
@@ -16,6 +17,13 @@ class TFBalances extends Balances {
     const decimalAmountInTFT = decimalAmount.mul(10 ** 7).toNumber();
     return await super.transfer({ address: options.address, amount: decimalAmountInTFT });
   }
+
+  async getMoreFunds(address: string, callback: any) {
```
AhmedHanafy725, line 21: this method should work with 0 args. we already know the caller's address and let's find a way to not send the callback to it and handle it internally.

- I suggest to create an instance from the tfchain client with `Alice` as mnemonic (it will fail in validation so let's make it as exception).
- Then use this client to do the transfer.

packages/playground/src/components/funds_card.vue
```diff
@@ -0,0 +1,143 @@
...
+                    });
+                  }
+                }
+              },
+            )
+            .catch((err: { message: string }) => {
+              console.log(err.message);
+              loadingAddTFT.value = false;
+              createToast(
+                ` Get more TFT failed! <br>Maybe the funding wallet has run out of TFTs. Please contact support`,
```
AhmedHanafy725, line 110:
```suggestion
                `Get more TFT failed! <br>Maybe the funding wallet has run out of TFTs. Please contact support`,
```

packages/playground/src/components/funds_card.vue
```diff
@@ -0,0 +1,143 @@
...
+                          hideProgressBar: true,
+                          toastBackgroundColor: "#1aa18f",
+                          timeout: 5000,
+                          type: "success",
+                          showIcon: true,
+                        });
+                        const profile = await loadProfile(client);
+                        profileManager.set(profile);
+                      } else if (section === "system" && method === "ExtrinsicFailed") {
+                        createToast(` Get more TFT failed!`, {
```
AhmedHanafy725, line 91:
```suggestion
                        createToast(`Get more TFT failed!`, {
```

AhmedHanafy725 (changes requested):

MohamedElmdary (approved):

AhmedHanafy725 (approved):

Issue #1118, opened 2023-09-18
Update balance after transactions


amiraabouhadid:
### Description

A method to trigger balance displayed in playground to update after transactions is needed.

### Implementation

If you've a specific approach to discuss, describe it here clearly.

khaledyoussef24:
verified now balance is updated instantly after transactions 
[Screencast from 10-02-2023 01:32:53 PM.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/c1af3c24-04c1-44b5-86d2-8833e824f1bf)

PR #1123, opened 2023-09-18
Support reload balance

MohamedElmdary:
## Issues
- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1118

### Usage
```vue
<script lang="ts" setup>
import { useProfileManagerController } form "components/profile_manager_controller.vue";

const profileManagerController = useProfileManagerController()

// to reload (safe to call it if there is no profile)
profileManagerController.reloadBalance()

</script>
```

### DIFF
--- a/packages/playground/src/App.vue
+++ b/packages/playground/src/App.vue
@@ -1,26 +1,50 @@
 <template>
   <v-app>
-    <v-navigation-drawer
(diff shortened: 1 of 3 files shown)


amiraabouhadid (approved):
works, thanks

PR #1176, opened 2023-10-04
added wireguard address in deployment table

Mik-TF:
Since we added wireguard connection for full VM and micro VM, this PR proposes to add the wireguard address in the deployment table

I tested it and it works well but please don't hesitate to run some more tests before merging this PR.

Thanks!

### DIFF
--- a/packages/playground/src/components/vm_deployment_table.vue
+++ b/packages/playground/src/components/vm_deployment_table.vue
@@ -36,6 +36,10 @@
         {{ item.value[0].planetary || "-" }}
       </template>
 
+      <template #[`item.wireguard`]="{ item }">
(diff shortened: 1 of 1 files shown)


xmonader:
Hmm isn't it visible on the details page? because the table has more information than it needs imo

Mik-TF:
Yes it is visible on the details page. But we had many feedback from farmers/users that wireguard was easier and more reliable than planetary network (e.g. users with windows had trouble sshing with planetary network, while wireguard was working quickly).

That's why I propose this. If you think it isn't needed, we can discard this change and close the PR.

Mahmoud-Emad (approved):

xmonader:
@Mik-TF Thanks for your contribution, I merged :)

Issue #1095, opened 2023-09-13
GPU name filter validation in dedicate node should allow special char


A-Harby:
### Description

GPU'S VENDOR NAME and GPU'S DEVICE NAME Filter input validation in dedicate node should allow some special character as GPU card name can contain them like `[ ] / , .`

### Logs/Alerts

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/1b420eae-ca68-41a4-afcf-a67daab8c7a9)
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/37c7a8bc-5b4a-4398-94cb-c01e7cb4d5e1)

AlaaElattar:
- fixed and handled in dedicated node pr

A-Harby:
Verified,
Devnet, 28d6c07.

The filter validation now accepts special char [ ] / - , .
https://gridproxy.dev.grid.tf/nodes?rentable=true&status=up&ret_count=true&size=10&page=1&gpu_vendor_name=amd%20[gpu/test]&gpu_device_name=amd%20[gpu/test]
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/7f576a9c-5b71-4888-8bc8-7f987a3e4189)

[TC2230 - Filter dedicated nodes by GPU's device name](https://app.testlodge.com/a/26076/projects/40893/suites/234374?expand_section=422339#case_4150439)
[TC2229 - Filter dedicated nodes by GPU's vendor name](https://app.testlodge.com/a/26076/projects/40893/suites/234374?expand_section=422339#case_4150438)

Issue #1018, opened 2023-08-20
Implement Dedicated Nodes in Vue3


zaelgohary:

amiraabouhadid:
On hold till transfer page is complete

amiraabouhadid:
On hold until #1090 is resolved

AlaaElattar:
- The implementation of filters card is done, waiting to confirm the design with the rest of the team to make filters compatible in all pages.  
- Now working on implementing the table.

AlaaElattar:
- still working on it but most of the page is done.

AlaaElattar:
- All page is done except applying filters on table.

AlaaElattar:
- working on apply pr comments

AlaaElattar:
- waiting till explorer nodes pr is merged as there're methods shared between us.

AlaaElattar:
- I've rebased my pr on explorer nodes, applied all pr comments and waiting for review.

AlaaElattar:
- will reuse cards from node explorer pr, so i put it in draft.

AlaaElattar:
- Cards applied and now look the same in dashboard. 
- waiting for review.

AlaaElattar:
- Applied pr comments && fixed new issue https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1226

AlaaElattar:
- working on fixing reload on gpu card.

AlaaElattar:
- Gpu reload handled and waiting for review

khaledyoussef24:
dedicated now is accessible 
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/a6e91f0c-ebc5-4ca7-8747-d0ba149765bf)
rented tab is removed 
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/9db030ac-58cb-4490-a54e-32128f3d2167)
gpu filtering options are also available 
info for every node (show details)is functional:
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/f23f78dc-f797-471a-88b4-046cec1ddfa4)

PR #1180, opened 2023-10-05
Dedicated Nodes in vue3

AlaaElattar:
### Description
- Implement dedicated nodes page in vue 3 with using gridproxy client
- User can filter on nodes, see node details, reserve and unreserve node. 
- Note that reserve or unreserve node may take avg 20 seconds 
- Pagination handled like in mainnet that contains many nodes.
- Also this issue is considered here https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1095

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1018
### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

image
 - nodes with GPU
 
Screenshot from 2023-10-04 15-10-08
Screenshot from 2023-10-04 15-10-25

https://github.com/threefoldtech/tfgrid-sdk-ts/assets/54944491/81a3a7e1-e568-482a-8f0d-b0e23529035b

https://github.com/threefoldtech/tfgrid-sdk-ts/assets/54944491/27d24b51-701c-48a5-8c88-39d34390e8e0


packages/playground/src/utils/filter_nodes.ts
```diff
@@ -117,3 +127,68 @@ export const inputsInitializer: FilterInputs = {
     type: "text",
   },
 };
+
+export const DedicatedNodeInitializer: DedicatedNodeFilters = {
+  total_sru: {
+    label: "Total SRU (GB)",
```
AhmedHanafy725, line 133: please use CPU, RAM, SSD, and HDD for labels

AhmedHanafy725 (changes requested):

AhmedHanafy725:
the word is not centered in the tab

image

AhmedHanafy725:
when we go to `Mine` tab and then go to `Rentable`, the loading progress indicator of the table is not working


https://github.com/threefoldtech/tfgrid-sdk-ts/assets/41957921/b693e4d1-3d86-4fc8-96e1-4858dbc97659

AhmedHanafy725:
let's make the price toFixed 2 or 3 max

image

packages/playground/src/components/dedicated_nodes_table.vue
```diff
@@ -0,0 +1,282 @@
...
+  loading.value = true;
+  try {
+    const data = await gridProxyClient.nodes.list({
+      ...params,
+      size: pageSize.value,
+      page: page.value,
+      freeSru: filterInputs.value.total_sru.value ? +filterInputs.value.total_sru.value : 0,
+      freeHru: filterInputs.value.total_hru.value ? +filterInputs.value.total_hru.value : 0,
+      freeMru: filterInputs.value.total_mru.value ? +filterInputs.value.total_mru.value : 0,
+      totalCru: filterInputs.value.total_cru.value ? +filterInputs.value.total_cru.value : 0,
```
AhmedHanafy725, line 165: is it possible to set the filter to an empty string or `undefined` in case the filter is not required cause the query would take much more time when they are set to 0

AhmedHanafy725:
can't press the links on the price tool-tip


https://github.com/threefoldtech/tfgrid-sdk-ts/assets/41957921/abac8ad6-e61e-49fc-aa88-bdcead9de18f

AhmedHanafy725:
we should show a loading spinner while loading the GPU card


https://github.com/threefoldtech/tfgrid-sdk-ts/assets/41957921/121adfd2-54f7-4fd4-b398-02c63496346a

AlaaElattar:
> when we go to `Mine` tab and then go to `Rentable`, the loading progress indicator of the table is not working
- This occurs due to the timing of the data retrieval and the visibility of the loading indicator so, I've added small timeout befor set loading to false.

AhmedHanafy725:
are the filters working with bytes or gigabytes

image

packages/playground/src/components/reserve_action_btn.vue
```diff
@@ -0,0 +1,199 @@
...
+
+    function removeReserve() {
+      openUnreserveDialog.value = true;
+    }
+
+    async function unReserveNode() {
+      loadingUnreserveNode.value = true;
+      try {
+        const grid = await getGrid(profileManager.profile!);
+        createToast(`check for contracts on node ${props.nodeId}`, {
```
AhmedHanafy725, line 77: custom toast

packages/playground/src/utils/filter_nodes.ts
```diff
@@ -147,3 +157,80 @@ export const inputsInitializer: FilterInputs = {
...
+      [
+        isDecimal("This Field accepts only a valid number."),
+        min("This Field must be a number larger than 0.", 1),
+        validateResourceMaxNumber("This value is out of range."),
+      ],
+    ],
+  },
+  total_hru: {
+    label: "Total HDD (GB)",
+    placeholder: "Filter by total HDD greater than or equal to.",
```
AhmedHanafy725, line 176: placeholder is bigger than the text field. can we have a smaller one?

image

AhmedHanafy725:
please order the filter to be cpu, mem, ssd, and hdd 

image

AhmedHanafy725:
please add debounce while loading the data.

- delay the request 1 sec after the user stops editing/entering values
- disable the filters while loading the data

packages/playground/src/components/dedicated_nodes_table.vue
```diff
@@ -0,0 +1,259 @@
...
+                  <a
+                    target="_blank"
+                    href="https://manual.grid.tf/dashboard/portal/dashboard_portal_dedicated_nodes.html#billing--pricing"
+                  >
+                    node
+                  </a>
+                </li>
+                <li>
+                  You're receiving {{ item.raw.discount }}% discount as per the
+                  <a target="_blank" href="https://manual.grid.tf/cloud/cloudunits_pricing.html#discount-levels">
```
AhmedHanafy725, line 70:
```suggestion
                  <a target="_blank" href="https://manual.grid.tf/cloud/cloudunits_pricing.html#staking-discount">
```

AhmedHanafy725:
- can we add style to the error message?
- also some spacing
- and reload button would be great too

image

maybe like

image

AhmedHanafy725:
please re-order the fields here to cpu, mem, ssd, hdd

image

AhmedHanafy725:
I think this confirmation popup needs some styles

image

AhmedHanafy725:
let's re-order the location info (country, city, lat, long)

image

packages/playground/src/components/node_details.vue
```diff
@@ -0,0 +1,308 @@
...
+                      <span>Card id that's used in a deployment</span>
+                    </v-tooltip>
+                    <v-col class="pa-0 d-flex justify-end align-center pl-2">
+                      <v-select
+                        style="max-width: 250px"
+                        v-if="gpuItem.length > 1"
+                        append-outer-icon="mdi-content-copy"
+                        hide-details
+                        dense
+                        v-model="gpuItem[0].id"
```
AhmedHanafy725, line 179: we should show the rest gpu cards as well. please check the old dashboard implementation

AlaaElattar:
- Added some style to pop up question 

image

- also if there's any error while getting node details 

image

AhmedHanafy725:
reload button does the request but doesn't reload the UI


https://github.com/threefoldtech/tfgrid-sdk-ts/assets/41957921/7207e90f-70b3-47b1-bee4-481b4ce76299

AhmedHanafy725:
error text is not centered, also the reload button does not load the UI

image

packages/playground/src/components/dedicated_nodes_table.vue
```diff
@@ -0,0 +1,268 @@
...
+  loading.value = true;
+  isFormLoading.value = true;
+  try {
+    const data = await gridProxyClient.nodes.list({
+      ...params,
+      size: pageSize.value,
+      page: page.value,
+      totalSru: filterInputs.value.total_sru.value ? toBytes(+filterInputs.value.total_sru.value) : undefined,
+      totalMru: filterInputs.value.total_hru.value ? toBytes(+filterInputs.value.total_hru.value) : undefined,
+      totalHru: filterInputs.value.total_mru.value ? toBytes(+filterInputs.value.total_mru.value) : undefined,
```
AhmedHanafy725, line 170:
```suggestion
      totalMru: filterInputs.value.total_mru.value ? toBytes(+filterInputs.value.total_mru.value) : undefined,
      totalHru: filterInputs.value.total_hru.value ? toBytes(+filterInputs.value.total_hru.value) : undefined,
```

AhmedHanafy725:
please disable the sorting on the table cause we don't support it so far

Mahmoud-Emad:
Please note that I have updated the `form filters` in my branch with an attribute called `disable-filter` that takes a `bool` value and disables the inputs when the `bool value` is `True`, you'll need to update your table with this field.

AlaaElattar:
- The button will be disabled in both reserving or unreserving states, so user won't click while submitting transaction.




https://github.com/threefoldtech/tfgrid-sdk-ts/assets/54944491/eccc31d4-28fc-4334-b8cf-576af49b4f0c

AlaaElattar:
- With using cards from explorer nodes.
- Node with GPU

Screenshot from 2023-10-16 13-20-16

- Node without GPU

Screenshot from 2023-10-16 13-20-32

packages/playground/src/components/node_details.vue
```diff
@@ -0,0 +1,176 @@
...
+    <div class="pa-1">
+      <v-progress-circular indeterminate model-value="20" :width="3"></v-progress-circular>
+    </div>
+  </td>
+  <td :colspan="columnsLen" v-else-if="dNodeError" style="text-align: center">
+    <div class="pt-4">
+      <v-alert
+        variant="tonal"
+        class="d-flex justify-between"
+        color="#f44336"
```
Mahmoud-Emad, line 12: Why static color? we can use `error` instead in case of the danger color

packages/playground/src/components/node_details.vue
```diff
@@ -0,0 +1,176 @@
...
+      <v-col :cols="getColSize" :class="{ 'mt-n8': getColSize() === 6 }">
+        <div class="mt-3">
+          <card-details
+            :loading="false"
+            title="Farm details"
+            icon="mdi-silo"
+            :items="FarmResourceFields"
+          ></card-details>
+        </div>
+      </v-col>
```
Mahmoud-Emad, line 59: How about moving each of these three cards into its own isolated component? This approach would streamline the code and enhance the overall structure of our project. What are your thoughts on this idea?

packages/playground/src/components/node_details.vue
```diff
@@ -0,0 +1,176 @@
...
+const publicIps = ref(0);
+const NodeResourceFields = ref<NodeDetailsCard[]>();
+const CountryResourceFields = ref<NodeDetailsCard[]>();
+const FarmResourceFields = ref<NodeDetailsCard[]>();
+const node = ref<GridNode>(nodeInitializer);
+
+const props = defineProps({
+  item: {
+    required: true,
+    type: null,
```
Mahmoud-Emad, line 93: What is this item of type null?

packages/playground/src/components/node_details.vue
```diff
@@ -0,0 +1,176 @@
...
+  },
+  columnsLen: {
+    required: true,
+    type: Number,
```
Mahmoud-Emad, line 97: columnsLen? what this used for?

packages/playground/src/components/dedicated_nodes_table.vue
```diff
@@ -0,0 +1,281 @@
...
+        hru: toGigaBytes(item.total_resources.hru.toString()),
+        mru: toGigaBytes(item.total_resources.mru.toString()),
+        sru: toGigaBytes(item.total_resources.sru.toString()),
+        ipv4u: false,
+        certified: item.certificationType === CertificationType.Certified,
+      });
+      return {
+        ...item,
+        price: (price?.dedicatedPrice ?? 0 + (fee || 0)).toFixed(3),
+        discount: price?.sharedPackage.discount ?? 0,
```
Mahmoud-Emad, line 207: Property 'discount' does not exist on type 'string'

packages/playground/src/components/node_details.vue
```diff
@@ -0,0 +1,176 @@
...
+async function getNodeDetails() {
+  try {
+    dNodeLoading.value = true;
+    const res = await gridProxyClient.farms.list({ farmId: props.item.value.farmId });
+    farmName.value = res.data[0].name;
+    publicIps.value = res.data[0].publicIps.length;
+    if (Array.isArray(res) && !res.length) throw new Error("Can't resolve farm data");
+    NodeResourceFields.value = getNodeResourceCard();
+    CountryResourceFields.value = getCountryResourceCard();
+    FarmResourceFields.value = getFarmResourceCard();
```
Mahmoud-Emad, line 114: If that's the case, you can call the function directly. Additionally, in TypeScript, the naming convention suggests using `camelCaseName`. For instance, instead of `NodeResourceFields`, it should be `nodeResourceFields`.

packages/playground/src/components/node_details.vue
```diff
@@ -0,0 +1,176 @@
...
+import CardDetails from "@/explorer/components/node_details_cards/card_details.vue";
+import GPUDetailsCard from "@/explorer/components/node_details_cards/gpu_details_card.vue";
+
+export default {
+  name: "Node Details",
+  components: {
+    CardDetails,
+    GPUDetailsCard,
+  },
+};
```
Mahmoud-Emad, line 175: I suggest moving `node_details` to `portal/components/node_details.vue` for clarity. Placing it in the main component folder might cause confusion, especially for new developers. This way, it's easier for everyone to understand its purpose.

All Portal related component can be moved IMHO

packages/playground/src/components/reserve_action_btn.vue
```diff
@@ -0,0 +1,149 @@
+<template>
+  <container>
+    <v-dialog v-model="openUnreserveDialog" max-width="600" class="custom-dialog">
+      <v-card>
+        <v-card-title class="pt-4">
+          <span style="display: flex; align-items: center">
+            <v-icon color="warning" style="margin-right: 8px">mdi-alert</v-icon>
```
Mahmoud-Emad, line 7: use `mr-1:15` for `margin-right`

packages/playground/src/components/reserve_action_btn.vue
```diff
@@ -0,0 +1,149 @@
...
+            Are you sure you want to unreserve this dedicated node?
+          </span>
+        </v-card-title>
+        <v-divider style="color: #1aa18f" />
```
Mahmoud-Emad, line 11: What is this color is used for?

packages/playground/src/components/reserve_action_btn.vue
```diff
@@ -0,0 +1,149 @@
...
+          <v-btn variant="outlined" color="primary" @click="openUnreserveDialog = false">No</v-btn>
+        </v-card-actions>
+      </v-card>
+    </v-dialog>
+    <v-btn
+      small
+      outlined
+      :loading="loadingReserveNode"
+      :disabled="disableButton"
+      color="#064663"
```
Mahmoud-Emad, line 24: The same goes here.

packages/playground/src/components/reserve_action_btn.vue
```diff
@@ -0,0 +1,149 @@
...
+      style="background: #1aa18f"
```
Mahmoud-Emad, line 25: The same goes here.

packages/playground/src/components/reserve_action_btn.vue
```diff
@@ -0,0 +1,149 @@
...
+    >
+      Reserve
+    </v-btn>
+    <v-btn
+      small
+      outlined
+      color="red"
+      :loading="loadingUnreserveBtn"
+      :disabled="disableButton"
+      style="background: #1aa18f"
```
Mahmoud-Emad, line 37: The same goes here.

packages/playground/src/components/reserve_action_btn.vue
```diff
@@ -0,0 +1,149 @@
...
+import { ref } from "vue";
+
+import { useProfileManager } from "../stores";
+import { createCustomToast, ToastType } from "../utils/custom_toast";
+import { getGrid } from "../utils/grid";
+
+const profileManager = useProfileManager();
+
+export default {
+  name: "ReserveBtn",
```
Mahmoud-Emad, line 57: UnReserveDialog?

packages/playground/src/components/reserve_action_btn.vue
```diff
@@ -0,0 +1,149 @@
...
+      required: true,
+    },
+    nodeId: {
+      type: String,
+      required: true,
+    },
+    twinId: {
+      type: Number,
+      required: true,
+    },
```
Mahmoud-Emad, line 70: Why not pass the entire node object instead of extracting these values here? It would simplify the process and allow us to access these fields directly from the node object within this component.

also the twinId, can be loaded from the `profileManager`

```ts
profileManager.profile?.twinId"
```

packages/playground/src/components/reserve_action_btn.vue
```diff
@@ -0,0 +1,149 @@
...
+
+    function removeReserve() {
+      openUnreserveDialog.value = true;
+    }
+
+    async function unReserveNode() {
+      loadingUnreserveNode.value = true;
+      try {
+        const grid = await getGrid(profileManager.profile!);
+        createCustomToast(`check for contracts on node ${props.nodeId}`, ToastType.info);
```
Mahmoud-Emad, line 87: Shall we improve the message to be -> `Verify contracts for node <NodeID>`?

packages/playground/src/utils/filter_nodes.ts
```diff
@@ -147,3 +157,81 @@ export const inputsInitializer: FilterInputs = {
...
+      ],
+    ],
+  },
+  total_sru: {
+    label: "Total SSD (GB)",
+    placeholder: "Filter by total SSD.",
+    type: "text",
+    rules: [
+      [
+        isDecimal("This Field accepts only a valid number."),
```
Mahmoud-Emad, line 188: isDecimal? why it should be Decimal? You can use `isNumeric` to validate if the resource type is number

packages/playground/src/utils/helpers.ts
```diff
@@ -57,3 +57,15 @@ export function getDashboardURL(network: string) {
 export function getCardName(card: NodeGPUCardType): string {
   return card.vendor + " - " + card.device;
 }
+
+export function toGigaBytes(value?: string) {
+  const giga = 1024 ** 3;
+
+  if (!value) return 0;
+
+  const val = +value;
```
Mahmoud-Emad, line 66: I suggest accepting the value as a number directly instead of converting it from a string. What are your thoughts on this?

Mahmoud-Emad (changes requested):
Well done, Alaa!

I've added some comments that might enhance our code/project; please take a look at them.

AlaaElattar:
- This issue is solved here  https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1226

Screenshot from 2023-10-17 10-48-20

Screenshot from 2023-10-17 10-48-30

packages/playground/src/components/node_details.vue
AlaaElattar: The cards are already components, what's the point of moving components to another component ?

packages/playground/src/components/node_details.vue
AlaaElattar: The item contains data of node at specific row.

packages/playground/src/components/node_details.vue
AlaaElattar: To specify number of columns a cell should span.

packages/playground/src/components/dedicated_nodes_table.vue
AlaaElattar: - both of type number, where is type `string` ??

image

packages/playground/src/components/reserve_action_btn.vue
AlaaElattar: Screenshot from 2023-10-11 14-30-36

packages/playground/src/components/reserve_action_btn.vue
AlaaElattar: I don't understand ur comments. The component is for button to reserve or unreserve.

packages/playground/src/components/dedicated_nodes_table.vue
```diff
@@ -0,0 +1,281 @@
...
+      ...params,
+      size: pageSize.value,
+      page: page.value,
+      totalSru: filterInputs.value.total_sru.value ? toBytes(+filterInputs.value.total_sru.value) : undefined,
+      totalMru: filterInputs.value.total_mru.value ? toBytes(+filterInputs.value.total_mru.value) : undefined,
+      totalHru: filterInputs.value.total_hru.value ? toBytes(+filterInputs.value.total_hru.value) : undefined,
+      totalCru: filterInputs.value.total_cru.value ? +filterInputs.value.total_cru.value : undefined,
+      gpuVendorName: filterInputs.value.gpu_vendor_name.value ? filterInputs.value.gpu_vendor_name.value : "",
+      gpuDeviceName: filterInputs.value.gpu_device_name.value ? filterInputs.value.gpu_device_name.value : "",
+    });
```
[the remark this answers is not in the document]
AlaaElattar: It will be the same, there's already try & catch here for handling any error.

packages/playground/src/components/node_details.vue
Mahmoud-Emad: Absolutely, your points make sense:

1. **Isolating Components:** While the cards are components, isolating them further could enhance code readability and maintainability. Each card can have its dedicated file, making it easier to comprehend and manage. It's a common practice to break down complex components for better organization, especially as our project evolves.

2. **Current State as Components:** You're right; the cards are components already. However, the idea is to make them even more modular. Each card could have its directory, containing the card component file, its specific styles, and any related assets. This way, each card is self-contained, which can be beneficial as the complexity of our project increases.

3. **Optimizing Readability:** Reducing the amount of code per component can indeed enhance readability. By isolating the cards, we're essentially allowing each component to focus on a specific functionality, avoiding clutter and improving the overall comprehension of the codebase.

packages/playground/src/components/node_details.vue
Mahmoud-Emad: So `node` with node type will be better.

packages/playground/src/components/dedicated_nodes_table.vue
Mahmoud-Emad: Forget about it, it was my **VSCODE :D

packages/playground/src/components/reserve_action_btn.vue
Mahmoud-Emad: it could be our primary color.

AlaaElattar: yes but if i typed `primary` it doesn't reflect, that's why i added it like that.

packages/playground/src/components/node_details.vue
AlaaElattar: yess, I got your poitn but still it's not that huge part of code. I don't think it will make difference if i added another component, also it won't be used somewhere else.

Mahmoud-Emad (approved):

AhmedHanafy725:
the cards are not aligned when the gpu card is there

also, the gpu card is so stretched when the screen becomes wider. I think having the gpu card in the same row with the rest was looking better

image

AhmedHanafy725:
in case the gpu request fails, it will keep loading forever. while we should show the other info and say that failed to load the gpu info as it was earlier in this PR

try with node 692 on mainnet

AhmedHanafy725 (approved):

PR #1208, opened 2023-10-12
Nextcloud with disk and gateway

scottyeager:
This PR updates the Nextcloud solution to use a disk (for improved performance, especially with Docker) and also to support gateways.

### DIFF
--- a/packages/playground/src/weblets/tf_deployment_list.vue
+++ b/packages/playground/src/weblets/tf_deployment_list.vue
@@ -187,13 +187,13 @@
           />
           <IconActionBtn
             tooltip="Open Nextcloud"
-            color="white"
+            color="anchor"
             icon="mdi-web"
             :href="'https://' + item.value[0].env.NEXTCLOUD_DOMAIN"
           />
           <IconActionBtn
             tooltip="Nextcloud Setup"
-            color="white"
+            color="anchor"
             icon="mdi-view-dashboard"
             :href="'https://' + item.value[0].env.NEXTCLOUD_AIO_LINK"
           />
--- a/packages/playground/src/weblets/tf_deployment_list.vue
+++ b/packages/playground/src/weblets/tf_deployment_list.vue
@@ -187,9 +187,15 @@
           />
           <IconActionBtn
             tooltip="Open Nextcloud"
+            color="white"
             icon="mdi-web"
-            color="anchor"
-            :href="'https://' + item.value[0].env.NEXTCLOUD_DOMAIN + ':8443'"
+            :href="'https://' + item.value[0].env.NEXTCLOUD_DOMAIN"
+          />
+          <IconActionBtn
+            tooltip="Nextcloud Setup"
+            color="white"
+            icon="mdi-view-dashboard"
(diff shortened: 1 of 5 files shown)


packages/playground/src/weblets/tf_nextcloud.vue
```diff
@@ -28,199 +29,222 @@
...
+  layout.value.setStatus(
+    "success",
+    "Successfully deployed a Nextcloud instance. Under Actions, click on the button Nextcloud Setup to set up Nextcloud. After installation, you can access the Nextcloud instance by clicking on the Open Nextcloud button or navigating to your Nextcloud domain.",
+  );
+  layout.value.openDialog(deployment, deploymentListEnvironments.nextcloud);
+}
+async function deploy(gatewayName: GatewayNode, customDomain: boolean) {
   layout.value.setStatus("deploy");
 
   const projectName = ProjectName.Nextcloud.toLowerCase();
```
MohamedElmdary, line 140:
```suggestion
  const projectName = ProjectName.Nextcloud.toLowerCase() + "/" + name.value;
```
Please Add /{name} as suffix to projectName as we changing the way we storing all machines to include gateways

Mahmoud-Emad:
Hello @scottyeager :)

Good job :), I've tried to deploy an instance of `NextCloud` multiple times and I got this error.

image

AhmedHanafy725:
> Good job :), I've tried to deploy an instance of `NextCloud` multiple times and I got this error.
> 
> image

I can see that the name starts with `NaN` which means something undefined is used in the name and the name contract is rejected as it contains uppercase letters

scottyeager:
I've reverted the suggestion that creates the error. @Mahmoud-Emad you can try testing again now :)

@MohamedElmdary, it looks like we need to wait for your PR with the gateway updates to be merged before changing the project name here.

packages/playground/src/weblets/tf_deployment_list.vue
```diff
@@ -187,9 +187,15 @@
           />
           <IconActionBtn
             tooltip="Open Nextcloud"
+            color="white"
```
Mahmoud-Emad, line 190: Please remove the static color here, it should be dynamic

see in dark mode 

- image

light mode 
- image

packages/playground/src/weblets/tf_deployment_list.vue
```diff
@@ -187,9 +187,15 @@
...
             icon="mdi-web"
-            color="anchor"
-            :href="'https://' + item.value[0].env.NEXTCLOUD_DOMAIN + ':8443'"
+            :href="'https://' + item.value[0].env.NEXTCLOUD_DOMAIN"
+          />
+          <IconActionBtn
+            tooltip="Nextcloud Setup"
+            color="white"
```
Mahmoud-Emad, line 196: The same goes here

Mahmoud-Emad (changes requested):

packages/playground/src/weblets/tf_deployment_list.vue
Mik-TF: Done. We set "anchor" as in all the other weblets, so it's uniform on the playground.

Issue #1209, opened 2023-10-13
The user is unable to access WireGuard on Umbrel Weblet:


shng4you:
Description

The user is trying to access/enable the WireGuard network. Unfortunately, the option is not available.

![WireGuard](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/95933607/b09034ee-efe9-4bac-9210-edd831399009)

And the user is requesting to access the WireGuard on Umbrel weblet.

Mik-TF:
Thanks for creating the issue Shang.

The [PR 1211](https://github.com/threefoldtech/tfgrid-sdk-ts/pull/1211) adds this feature. Once mergee we will be able to close this issue.

PR #1211, opened 2023-10-13
added wireguard access to umbrel

Mik-TF:
From [this issue](https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1209) this PR adds the wireguard access for Umbrel on the Playground.

Did some testings and it works fine. 

***

EDIT: New commits include those changes:

* Added WireGuard and IPv4 network options
  * Nested ternary for network options (IPv4, Planetary, WireGuard) 
* IPv4 wasn't set completely on the Umbrel weblet
  * Now users can decide to use the 3 network options and the Admin panel button adjust accordingly with the proper link
* Added Umbrel in the WireGuard table so users can see the address when they use WireGuard
* Planetary is set by default to true
* One network choice is required as a minimum

### DIFF
--- a/packages/playground/src/components/vm_deployment_table.vue
+++ b/packages/playground/src/components/vm_deployment_table.vue
@@ -116,7 +116,7 @@ const filteredHeaders = computed(() => {
     { title: "Public IPv4", key: "ipv4" },
     { title: "Public IPv6", key: "ipv6" },
     { title: "Planetary Network IP", key: "planetary" },
-    { title: "Wireguard", key: "wireguard" },
+    { title: "WireGuard", key: "wireguard" },
     { title: "Flist", key: "flist" },
     { title: "Cost", key: "billing" },
     { title: "Actions", key: "actions" },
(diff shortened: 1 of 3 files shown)


Mahmoud-Emad:
Hey Mik, thanks for your help!

Issues are organized based on assigned labels and priority determined by the teams/leads. Once an issue has a `project` or `milestone` designation, it can be addressed accordingly.

I'm going to review this PR when the mentioned issue gets an update from @xmonader or @AhmedHanafy725

Mahmoud-Emad:
Great work, Mik! 👏

In case we decide to integrate Wireguard as a network option, please follow these steps:

1. **Update the Network Component**: Modify the `Network` component by adding the `v-model:wireguard="wireguard"` option. This will bind it to the `const wireguard = ref()` you've added earlier.

   ```html
   <Network v-model:wireguard="wireguard" />
   ```

2. **Set Default Value to `false`**: For consistency and clarity, it's recommended to set the default value of `wireguard` to `false` instead of `true` in the component's script.

   ```ts
   const wireguard = ref(false);
   ```

Mik-TF:
Thanks Emad.

I added what you propose as well as some other codes to make sure it all works with WireGuard and also IPv4.

I will edit the first comment so reviewers can see what the PR does now.

Mahmoud-Emad (approved):

Issue #1055, opened 2023-08-29
Add actions to vms to add/remove gateway workload to/from a vm


AhmedHanafy725:
Parent issue: https://github.com/threefoldtech/home/issues/1452

- User should be able to attach name gateway or FQDN gateway workloads to a VM.
- Add a link for the gateway on the vm workload so we can list the gateways attached to the vm
- From the listed gateways, the user should be able to remove any one of them.

MohamedElmdary:
Still investigating error while listing deployments
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/31689104/5e199154-8656-44e6-9438-b97213f19b19)

MohamedElmdary:
Issue found and fixed

MohamedElmdary:
PR should be ready to be merged 
- Add prefix for fvm/vm deployed gateways
- add mirgration for old solutions gateways

MohamedElmdary:
Contracts listing doesn't get updated correctly due to https://github.com/threefoldtech/tfchain_graphql/issues/135

A-Harby:
Devnet c9f2406.

A new tab was added to add a gateway to deployment.
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/7c4b9a82-371f-4322-86e8-ae1f2f3b939e)
And it was successfully deployed and added to the VM.
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/1d10bbd4-4208-45ab-a9f2-d9d7dca03f64)
Listing to the gateways on FullVM.
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/00791c01-aa40-4fea-8c1b-47dd66782669)
with TLS and on MicroVM.
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/73b73214-a2c2-4214-b0e3-3bd7f7e928ee)
Delete option, and it was successful.
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/c2a2cbf9-766d-4e16-88c8-b5c19840ef17)

xmonader:
While this looks great, the terms we are using aren't 100% accurate :(

A-Harby:
The only issue I found is that after deploying a VM and adding a gateway to it,
[...]

PR #1215, opened 2023-10-15
Implement manage gateways in playground

MohamedElmdary:
## Issues
- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1055


packages/playground/src/components/manage_gateway_dialog.vue
```diff
@@ -0,0 +1,280 @@
...
+              >Reload</v-btn
+            >
+          </div>
+          <list-table
+            :headers="[
+              { title: 'Name', key: 'name' },
+              { title: 'Contract ID', key: 'contractId' },
+              { title: 'Domain', key: 'domain' },
+              { title: 'TLS Passthrough', key: 'tls_passthrough' },
+              { title: 'Backends', key: 'backends' },
```
AhmedHanafy725, line 38: we only support one backend on the gateway deployment
```suggestion
              { title: 'Backend', key: 'backends' },
```

packages/playground/src/components/manage_gateway_dialog.vue
```diff
@@ -0,0 +1,280 @@
...
+
+    async function deployGateway() {
+      layout.value.setStatus("deploy");
+
+      try {
+        const [x, y] = ip.split(".");
+        const grid = await getGrid(profileManager.profile!, props.vm.projectName);
+        await grid.networks
+          .addNode({ name: networkName, ipRange: `${x}.${y}.0.0/16`, nodeId: gatewayNode.value.id })
+          .catch(() => null);
```
AhmedHanafy725, line 216: please check if the network already has this node before starting to add it.

AhmedHanafy725, line 216: In case of failure while adding the node to the network, show the error message.

packages/playground/src/weblets/tf_deployment_list.vue
```diff
@@ -25,6 +25,14 @@
             icon="mdi-eye-outline"
             @click="openDialog(tabs[activeTab].value, item)"
           />
+
+          <IconActionBtn icon="mdi-cog" tooltip="Deploy FQDN Gateway" @click="dialog = item.value.deploymentName" />
```
AhmedHanafy725, line 29:
```suggestion
          <IconActionBtn icon="mdi-cog" tooltip="Manage Gateways" @click="dialog = item.value.deploymentName" />
```

packages/playground/src/weblets/tf_deployment_list.vue
```diff
@@ -33,6 +41,14 @@
...
```
AhmedHanafy725, line 45:
```suggestion
          <IconActionBtn icon="mdi-cog" tooltip="Manage Gateways" @click="dialog = item.value.deploymentName" />
```

AhmedHanafy725 (changes requested):

AhmedHanafy725:
after deploying a micro vm, couldn't list it

image

AhmedHanafy725:
the network name and the IP shouldn't be editable 

image

AhmedHanafy725:
add an option for the user while deploying the gateway to allow pass-through or not. (add tool-tip to explain what it's used for)
(default: false)

packages/grid_client/src/modules/base.ts
```diff
@@ -135,8 +174,8 @@ class BaseModule {
     ext2 = await Promise.all(ext2.flat(1));
 
     if (this.backendStorage.type === BackendStorageType.tfkvstore) {
-      ext1 = ext1.filter(x => !!x);
-      ext2 = ext2.filter(x => !!x);
+      ext1 = ext1.flat(1).filter(x => !!x);
+      ext2 = ext2.flat(1).filter(x => !!x);
 
       await this.tfClient.applyAllExtrinsics(ext1);
       await this.tfClient.applyAllExtrinsics(ext2);
```
AhmedHanafy725, line 181: let's apply these extrinsics together but the new ones first on the list then the old ones

MohamedElmdary: @AhmedHanafy725 What text should I add on the tooltip for explaining pass-through?

MohamedElmdary:
> the network name and the IP shouldn't be editable
> 
> image

It's already readonly user can only copy them!

packages/grid_client/src/modules/base.ts
MohamedElmdary: > let's apply these extrinsics together but the new ones first on the list then the old ones

Merged both together

MohamedElmdary:
> after deploying a micro vm, couldn't list it
> 
> image

Found the issue and fixed and tested on all solutions

packages/grid_client/src/modules/base.ts
MohamedElmdary: Added tls pass through option

Mahmoud-Emad:
Great job, one small comment, what do you think if we replaced the check box of the custom domain option to be the same as the custom domain as added in the other solutions, `Switch` I mean

Mahmoud-Emad (approved):

AhmedHanafy725:
when deploying a peertube instance, the name contract contains `NaN`

image

mohamedamer453:
I was trying to test the changes but the playground failed while building the project.

image_2023-10-19_14-17-37

MohamedElmdary:
> I was trying to test the changes but the playground failed while building the project.
> 
> image_2023-10-19_14-17-37

Should be fixed was an issue related to types but didn't shown in development mode for some reason

mohamedamer453:
While testing i found that all deployments that i had already deployed from playground.dev.grid.tf were loaded successfully when i ran the playground from this branch except for kubernetes, when i tried to list the k8s deployment it failed to do so, however the key was sucessfully updated as seen in the polkadot ui.

image_2023-10-19_15-36-17

and when i tried to deploy a k8s deployment directly from this branch it failed to load and it also got stuck at the `Waiting for deployment` screen.

image_2023-10-19_15-50-15

MohamedElmdary:
> While testing i found that all deployments that i had already deployed from playground.dev.grid.tf were loaded successfully when i ran the playground from this branch except for kubernetes, when i tried to list the k8s deployment it failed to do so, however the key was sucessfully updated as seen in the polkadot ui.
> 
> image_2023-10-19_15-36-17
> 
> and when i tried to deploy a k8s deployment directly from this branch it failed to load and it also got stuck at the `Waiting for deployment` screen.
> 
> image_2023-10-19_15-50-15

We didn't apply any updates to k8s but by default k8s applying migration for it's contracts so that why it cases the issue
so I applied the exact solution from vms to k8s and it should be fixed now

mohamedamer453:
> We didn't apply any updates to k8s but by default k8s applying migration for it's contracts so that why it cases the issue so I applied the exact solution from vms to k8s and it should be fixed now

Indeed, when i checked the instances that were deployed before they were listed without any issues and then i deployed a new instance and again it got listed without any issues.

image

AhmedHanafy725 (approved):

Issue #1192, opened 2023-10-09
[Transfer] Error validation in tranfer by twinID


AlaaElattar:
- When i added my address in `By Twin ID` tab it didn't show any error.

![Screenshot from 2023-10-09 16-11-57](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/54944491/92407dbe-6e3c-4d54-b880-9a215f67acd9)

AlaaElattar:
- Bug fixed, waiting for review.

khaledyoussef24:
fixed and verified 
![Screenshot from 2023-10-17 12-05-10](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/7a8435b8-50b3-40f2-b85b-3434be74d89b)
![Screenshot from 2023-10-17 12-04-49](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/87d12df6-0c31-4902-96f4-ddda554bc7fc)

PR #1221, opened 2023-10-16
[Transfer] fix validation bug.

AlaaElattar:
### Description

- In transfer with Twin ID, the address passed without any error.

### Changes

- Validate that twin id number

### Related Issues
- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1192

image

### DIFF
--- a/packages/playground/src/portal/transfer_view.vue
+++ b/packages/playground/src/portal/transfer_view.vue
@@ -20,6 +20,7 @@
                   validators.required('Recepient Twin ID is required'),
                   validators.isNotEmpty('Invalid Twin ID'),
                   validators.isNumeric('Invalid Twin ID'),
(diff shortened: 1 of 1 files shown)


packages/playground/src/portal/transfer_view.vue
```diff
@@ -19,6 +19,7 @@
                 :rules="[
                   validators.required('Recepient Twin ID is required'),
                   validators.isNotEmpty('Invalid Twin ID'),
+                  validators.isNumeric('Invalid Twin ID'),
```
AhmedHanafy725, line 22: can we check also it's larger than 0

AlaaElattar: image

AhmedHanafy725 (approved):

Issue #1235, opened 2023-10-18
Expose `all` and `filter` methods on the nodes module to list/filter all nodes


Mahmoud-Emad:
They can call the `capacity.getAllNodes` and `capacity.filterNodes` methods

Mahmoud-Emad:
Should be added in https://github.com/threefoldtech/tfgrid-sdk-ts/pull/1238

Mahmoud-Emad:
### Issue update

- Created an `all` method to load all nodes from the capacity module
- Created a `filter` method to filter nodes from the capacity module
- Added one more field named `numberOfPublicIPs` to get the public ip count of the contract

Mahmoud-Emad:
The PR has been merged and is ready to be in the verification column.

AhmedHanafy725:
the required methods are having the `expose` decorator

PR #1238, opened 2023-10-18
Expose all and filter methods on the nodes module to list/filter all nodes

Mahmoud-Emad:
### Description

Created 2 methods 'all' and 'filter' on the nodes module to list and filter the nodes, and added one more field named numberOfPublicIPs returns the number of public IPs registered on the deployment.

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1235

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/packages/grid_client/src/modules/models.ts
+++ b/packages/grid_client/src/modules/models.ts
@@ -31,7 +31,7 @@ enum ContractStates {
   GracePeriod = "GracePeriod",
 }
 
-export enum NodeStatusFilter {
+export enum NodeStatus {
   up = "up",
   down = "down",
   standBy = "standby",
--- a/packages/grid_client/src/modules/models.ts
+++ b/packages/grid_client/src/modules/models.ts
@@ -590,7 +590,7 @@ class FilterOptions {
   @Expose() @IsOptional() @IsBoolean() hasGPU?: boolean;
   @Expose() @IsOptional() @IsBoolean() rentable?: boolean;
   @Expose() @IsOptional() @IsInt() @Min(1) rentedBy?: number;
-  @Expose() @IsOptional() @IsString() status?: NodeStatusFilter;
+  @Expose() @IsOptional() @Transform(({ value }) => NodeStatusFilter[value]) @IsEnum() status?: NodeStatusFilter;
 }
 
 enum CertificationType {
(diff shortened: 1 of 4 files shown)


packages/grid_client/src/modules/models.ts
```diff
@@ -584,6 +590,7 @@ class FilterOptions {
   @Expose() @IsOptional() @IsBoolean() hasGPU?: boolean;
   @Expose() @IsOptional() @IsBoolean() rentable?: boolean;
   @Expose() @IsOptional() @IsInt() @Min(1) rentedBy?: number;
+  @Expose() @IsOptional() @IsString() status?: NodeStatusFilter;
```
AhmedHanafy725, line 593: please use `isEnum` instead of string and also check the transformation should be done. You will find an example here in models file

packages/grid_client/src/modules/nodes.ts
```diff
@@ -79,6 +81,20 @@ class Nodes {
   async setNodePower(options: NodePowerModel) {
     return (await this.client.nodes.setPower(options)).apply();
   }
+
+  @expose
+  @validateInput
+  @checkBalance
```
AhmedHanafy725, line 87: we don't need to `validateInput` decorator for a method that doesn't have input.

also, for `checkBalance` user doesn't have to have tokens to do a query

packages/grid_client/src/modules/nodes.ts
```diff
@@ -79,6 +81,20 @@ class Nodes {
...
+  async all() {
+    return await this.capacity.getAllNodes();
+  }
+
+  @expose
+  @validateInput
+  @checkBalance
```
AhmedHanafy725, line 94: no need for `checkBalance` decorator here

AhmedHanafy725 (changes requested):

packages/grid_client/src/modules/models.ts
AhmedHanafy725: please check this [example](https://github.com/threefoldtech/tfgrid-sdk-ts/blob/development/packages/grid_client/src/modules/models.ts#L172)

packages/grid_client/src/modules/models.ts
```diff
@@ -31,6 +31,12 @@ enum ContractStates {
   GracePeriod = "GracePeriod",
 }
 
+export enum NodeStatusFilter {
```
AhmedHanafy725, line 34:
```suggestion
export enum NodeStatus {
```

AhmedHanafy725 (approved):

Issue #1021, opened 2023-08-20
Implement Explorer Farms in Vue3


zaelgohary:

ehab-hassan:
it'll be like that with the new filter and color 
https://www.figma.com/proto/b50GilnYwLrhwOUIhfzHNA/dashboard-Nav?type=design&node-id=528-1234&t=WHZwxIHMt9huGKI2-0&scaling=contain&page-id=0%3A1&starting-point-node-id=124%3A196

samaradel:
added selected farm and twin details

samaradel:
working on filters

samaradel:
worked on some UI in farms details dialog

samaradel:
tables and filters were implemented in wrong scenario and, fixing them...

samaradel:
The tables fixed and most of the filters

samaradel:
fix filters and details dialog

samaradel:
all is done and it needs pr review

amiraabouhadid:
addressing 16 comments on PR

amiraabouhadid:
Resolved 10 comments on PR, 6 more comments left to resolve.

amiraabouhadid:
rewritten farm_dialog to address comments on PR.

amiraabouhadid:
Making card details more dynamic.

amiraabouhadid:
rewritten farm_details_card

amiraabouhadid:
rewritten twin_details_card. Only filter farms remains to be fixed now.

amiraabouhadid:
working on farm filters

amiraabouhadid:
fixed farm filter by name. Working on reducing constants used.

amiraabouhadid:
working on making filters applicable to both farms and nodes to reduce repeated code

amiraabouhadid:
Work Completed:
submitted PR for review 

Work in Progress (WIP):
addressing comments on PR

amiraabouhadid:
Work Completed:
Fixed filter

Work in Progress (WIP):
Pagination

amiraabouhadid:
Work Completed:
 Pagination

Work in Progress (WIP):
Remaining comments on PR

amiraabouhadid:
Work Completed:
Removed redundant code, made filters optional

Work in Progress (WIP):
Investigating nodes filter

amiraabouhadid:
Work in Progress (WIP):
Investigating pagination malfunction

[...]

PR #1240, opened 2023-10-18
explorer farms

samaradel:
implement explorer farms 

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1021


packages/playground/src/explorer/farms.vue
```diff
@@ -0,0 +1,133 @@
...
+
+const loading = ref<boolean>(false);
+const farms = ref<IFarm[]>([]);
+const isDialogOpened = ref<boolean>(false);
+const selectedFarm = ref<IFarm>();
+const filterInputs = ref<FarmFilterInputs>(inputsInitializer);
+const filterOptions = ref<FarmFilterOptions>(farmOptionsInitializer);
+const isValidForm = ref<boolean>(false);
+const farmsCount = ref<number>(0);
+const mixedFilters = ref<FarmMixedFilter>({
```
Mahmoud-Emad, line 51: You don't need this mixed filter, you only have inputs.

The `mixedFilter` was created in the `node explorer` because we have two types of filters
1 - input filters which represent the `form-filters`
2 - options filter which represents the options like `node status`, `GPU support`, and `gateway support`

but in your case, you only need to create the inputs filter, see the `dedicated-nodes` component for more details.

packages/playground/src/explorer/farms.vue
```diff
@@ -0,0 +1,133 @@
...
+  freePublicIp: number;
+}
+const route = useRoute();
+
+const loading = ref<boolean>(false);
+const farms = ref<IFarm[]>([]);
+const isDialogOpened = ref<boolean>(false);
+const selectedFarm = ref<IFarm>();
+const filterInputs = ref<FarmFilterInputs>(inputsInitializer);
+const filterOptions = ref<FarmFilterOptions>(farmOptionsInitializer);
```
Mahmoud-Emad, line 48: Do we really need to create it?

packages/playground/src/explorer/farms.vue
```diff
@@ -0,0 +1,133 @@
...
+  name: string;
+  twinId: number;
+  pricingPolicyId: number;
+  certificationType: CertificationType;
+  publicIps: PublicIp[];
+  stellarAddress?: string;
+  totalPublicIp: number;
+  usedPublicIp: number;
+  freePublicIp: number;
+}
```
Mahmoud-Emad, line 40: Let's move this interface to `explorer/utils/types`

packages/playground/src/explorer/farms.vue
```diff
@@ -0,0 +1,133 @@
...
+const route = useRoute();
```
Mahmoud-Emad, line 41: The route here was created for what? I have a small suggestion, please watch the attached record.

[Screencast from 10-22-2023 06:05:55 PM.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/57001890/aea49c2c-4156-4b50-83c8-b43ec2738ffc)

In the record, you push the `?twinId=<twin_id>` into the route after the second click after mount, also you push it in the first click on the row, I've tried several times and indeed it gets pushed in the first click only, which is unexpected behavior.

However, we really should push the `farmId` instead, and then after mounting the page we should check the rout if it includes the `farmId` we have to load the farm and open the dialog for the user, else we must close and clear any opened data.

Here's an example

```vue
<script lang="ts">
  const checkSelectedNode = async () => {
      if (route.query.farmId) {
        selectedFarmId.value = +route.query.farmId;
        isDialogOpened.value = true;
      }
    };
  
  onMounted(async () => {
      await checkSelectedNode();
    });
    
    watch(selectedFarmId, async () => await getFarmById())
</script>
```

This is an example, you must clear the route `onClose`.

```ts
  const closeDialog = () => {
    if (route.query.farmId) {
      router.replace(route.path);
    }
    isDialogOpened.value = false;
    selectedFarmId.value = undefined;
  };
```

So we will push the `farmId` instead and will automatically open the dialog `onMounted` if there is a `farmId` in the route.

packages/playground/src/explorer/farms.vue
```diff
@@ -0,0 +1,133 @@
...
+      const used = ips.filter(x => x.contract_id === 0).length;
+      return {
+        ...farm,
+        totalPublicIp: total,
+        usedPublicIp: used,
+        freePublicIp: total - used,
+      };
+    });
+  } catch (error) {
+    console.log(error);
```
Mahmoud-Emad, line 75: can we display the error with a `reload` button instead?

packages/playground/src/explorer/farms.vue
```diff
@@ -0,0 +1,133 @@
...
+};
+
+watch(
+  mixedFilters,
+  async () => {
+    const queries = getFarmQueries(mixedFilters.value);
+    await _getFarms(queries);
+  },
+  { deep: true },
+);
```
Mahmoud-Emad, line 106: I think it needs more efforts

1. we need to change the `mixedFilter` since we'll delete it and use the `inputFilters` instead.
2. we have to wait on request for 1 or 2 seconds when the user stops writing, please see the attached record.
3. we need to disable the form fields while requesting the gridProxy, please if you implemented the filter forms from scratch you don't need to do it, you can use the implemented one.
4. display a message to the user to inform them that the request will happen after 1 second. see the `nodes` view.
5. we must make the request only if there are no errors. 
6. the pagination seems to be doesn't work as expected | `Items per page:`, it even doesn't reflect the selection.

[Screencast from 10-22-2023 06:29:06 PM.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/57001890/8b501003-9609-4d2a-b487-6a096d4d31e4)

instead of it you can use the `debounce` method to make your request, see this example

```vue
<script lang="ts">
  const request = debounce(_requestNodes, 1000);
  watch(
      inputFilters,
      async () => {
        await request(attrs);
      },
      { deep: true },
    );
</script>
```
The idea is we delay the request for one second after the user stops writing to avoid making many requests to the grid proxy.

packages/playground/src/explorer/utils/helpers.ts
```diff
@@ -137,3 +142,37 @@ export const getQueries = (mixedFilters: MixedFilter): Partial<NodesQuery> => {
...
+    size: mixedFilters.options.size,
+    retCount: mixedFilters.options.retCount,
+  };
+  return options;
+};
+
+export async function getFarms(queries: Partial<FarmsQuery>): Promise<Pagination<Farm[]>> {
+  try {
+    const farms = await gridProxyClient.farms.list(queries);
+    return JSON.parse(JSON.stringify(farms));
```
Mahmoud-Emad, line 163: Why we do `JSON.parse(JSON.stringify(farms));`? can't we just return the farms?

packages/playground/src/explorer/utils/helpers.ts
```diff
@@ -137,3 +142,37 @@ export const getQueries = (mixedFilters: MixedFilter): Partial<NodesQuery> => {
...
+  } catch (error) {
+    console.error("An error occurred while requesting farms:", error);
+    throw error;
+  }
+}
+
+export async function getTwins(queries: Partial<TwinsQuery> = {}): Promise<Pagination<Twin[]>> {
+  try {
+    const twins = await gridProxyClient.twins.list(queries);
+    return JSON.parse(JSON.stringify(twins));
```
Mahmoud-Emad, line 173: The same goes here.

packages/playground/src/explorer/utils/types.ts
```diff
@@ -18,6 +18,11 @@ export type MixedFilter = {
   options: FilterOptions;
 };
 
+export type FarmMixedFilter = {
+  inputs: FarmFilterInputs;
+  options: FarmFilterOptions;
```
Mahmoud-Emad, line 23: as we agreed above, we need to delete the `options` here.

packages/playground/src/explorer/utils/types.ts
```diff
@@ -121,3 +126,23 @@ export const nodeInitializer: GridNode = {
...
+  twinId: NodeInputFilterType;
+  certificationType: NodeInputFilterType;
+  pricingPolicyId: NodeInputFilterType;
+};
+
+export type FarmFilterOptions = {
+  page: number;
+  size: number;
+  retCount: boolean;
+};
```
Mahmoud-Emad, line 142: we don't need to create a type for the page and size, they can be `refs`, I've created them in the `explorer/nodes` because I have another field so I chose to include them in one interface.

packages/playground/src/utils/filter_farms.ts
```diff
@@ -0,0 +1,78 @@
+import type { CertificationType, Farm, FarmsQuery, Pagination } from "@threefold/gridproxy_client";
```
Mahmoud-Emad, line 1: Instead of recreating this component, we can definitely utilize the `node-filter` component. If there are any obstacles preventing its use, we can modify it to make it more versatile and reusable.

packages/playground/src/utils/filter_farms.ts
```diff
@@ -0,0 +1,78 @@
...
+    stellarAddress: options.stellarAddress,
+    pricingPolicyId: options.pricingPolicyId,
+    farmId: options.farmId,
+    twinId: options.twinId,
+    name: options.name,
+    certificationType: options.certificationType as CertificationType,
+  };
+  const farms = await gridProxyClient.farms.list(filters);
+  return farms;
+}
```
Mahmoud-Emad, line 27: Can't we simply pass the options to the list method?
```ts
const farms = await gridProxyClient.farms.list(options);
```

packages/playground/src/utils/filter_farms.ts
```diff
@@ -0,0 +1,78 @@
...
+        min("The farm id should be larger then zero.", 1),
+        startsWith("The farm id start with zero.", "0"),
+      ],
+    ],
+    type: "text",
+  },
+  name: {
+    label: "Farm Name",
+    placeholder: "Filter by farm name.",
+    type: "text",
```
Mahmoud-Emad, line 46: The farm name need a validator that validate that the farm name doesn't include any special chars

packages/playground/src/utils/filter_farms.ts
```diff
@@ -0,0 +1,78 @@
...
+        startsWith("The twin id start with zero.", "0"),
+      ],
+    ],
+    type: "text",
+  },
+  certificationType: {
+    label: "Certification Type",
+    placeholder: "Filter by certification type.",
+    type: "text",
+  },
```
Mahmoud-Emad, line 64: please add a validation if needed.

packages/playground/src/utils/get_farms.ts
```diff
@@ -1,4 +1,5 @@
 import type { FarmFilterOptions, FarmInfo, GridClient } from "@threefold/grid_client";
+import type { Farm as IFarm } from "@threefold/gridproxy_client";
 
 import type { Farm } from "@/types";
```
Mahmoud-Emad, line 4: What is the difference between the `IFarm` and `Farm`?

packages/playground/src/utils/get_farms.ts
```diff
@@ -19,23 +20,35 @@ export async function getFarms(
     farms = farms.filter(farm => !blockedFarms.has(farm.farmId));
   }
 
-  return farms.map(farm => ({ name: farm.name, farmID: farm.farmId, country: filters.country }));
+  return farms.map(farm => ({
+    name: farm.name,
+    farmID: farm.farmId,
+    country: filters.country,
+  }));
```
Mahmoud-Emad, line 27: Can't we simply return the farms list?

packages/playground/src/utils/get_farms.ts
```diff
@@ -44,3 +57,35 @@ export async function getBlockedFarmSet(exclusiveFor: string): Promise<Set<numbe
   );
   return new Set(farms.map(farm => farm.farmId));
 }
+
+export async function getAllFarms() {
+  const farms = await gridProxyClient.farms.listAll();
+  console.log(farms);
```
Mahmoud-Emad, line 63: Remove this console log

packages/playground/src/utils/get_farms.ts
```diff
@@ -44,3 +57,35 @@ export async function getBlockedFarmSet(exclusiveFor: string): Promise<Set<numbe
...
+  return farms;
+}
+
+// export function fallbackDataExtractor<T = GetDataQueryType, K extends keyof T = keyof T>(key: K): (state: IState) => T[K]; // prettier-ignore
+// export function fallbackDataExtractor<T = GetDataQueryType, K extends keyof T = keyof T>(key: K, state: IState): T[K]; // prettier-ignore
```
Mahmoud-Emad, line 68: remove these lines

packages/playground/src/utils/get_farms.ts
```diff
@@ -44,3 +57,35 @@ export async function getBlockedFarmSet(exclusiveFor: string): Promise<Set<numbe
...
+export function fallbackDataExtractor(key: any, state?: any) {
```
Mahmoud-Emad, line 69: Please document this function and any function created outside `vue component`

Mahmoud-Emad (changes requested):
Excellent work! 😊 I've added some comments that could enhance our performance. Please take a look! 😊

AlaaElattar:
- There is weird behavior, all data in table not returned after i removed the filter.

https://github.com/threefoldtech/tfgrid-sdk-ts/assets/54944491/58d490f9-b5d5-4ee3-9af4-4ed85d4a4c37

AlaaElattar:
- I think the  bar with x icon inside it too big
image

AlaaElattar:
- Here, the table keeps loading forever despite i added wrong filter.
image

zaelgohary (changes requested):
Request returns first 50 count only and paginations doesn't work well if more than 50 is chosen.

image

farms from Dashboard.
image

zaelgohary:
If the stellar address is empty, can we show a "-" instead of blank?

image


Also, if the farm details pop-up was closed in any way other than using the x icon, it won't be opened again, neither in this farm or any other farm.

[screen-capture.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/40770501/a6e9c4a7-70b6-4ac5-abff-88d0687ddaa0)

packages/playground/src/explorer/components/farms_table.vue
```diff
@@ -0,0 +1,97 @@
...
+    return {
+      headers,
+      itemPerPage,
+      openSheet,
+    };
+  },
+};
+</script>
+
+<style>
```
zaelgohary, line 86: Can we make sure the style is scoped so it won't affect other table styling?

packages/playground/src/explorer/components/node_details_cards/farm_details_card.vue
```diff
@@ -17,7 +17,11 @@ export default {
   props: {
     node: {
```
zaelgohary, line 18: Why are we passing the node to farm details? What if the farm has more than 1 node like freefarm?

packages/playground/src/explorer/components/node_details_cards/farm_details_card.vue
```diff
@@ -39,12 +43,22 @@ export default {
     };
 
     const getFarmDetails = (): NodeDetailsCard[] => {
+      let farmId, farmName, farmStellarAddress;
+      if (props.node) {
+        farmId = props.node.farmId.toString();
+        farmName = props.node.farm.name;
+        farmStellarAddress = props.node.farm.stellarAddress;
```
zaelgohary, line 50: I don't think this is needed.
```suggestion
```

packages/playground/src/explorer/components/node_details_cards/twin_details_card.vue
```diff
@@ -17,38 +18,58 @@ export default {
...
         {
           name: "Account ID",
-          value: props.node.twin.accountId,
+          value: accountId,
           icon: "mdi-content-copy",
           callback: copy,
           hint: "Copy the account id to the clipboard.",
         },
-        { name: "Relay", value: props.node.twin.relay },
+        { name: "Relay", value: relay },
```
zaelgohary, line 72: Can we set a default value as some values might return empty?

image

packages/playground/src/explorer/components/node_details_cards/twin_details_card.vue
```diff
@@ -17,38 +18,58 @@ export default {
...
+    });
 
     const copy = (address: string) => {
       navigator.clipboard.writeText(address);
       createCustomToast("Copied!", ToastType.success);
     };
 
     const getNodeTwinDetailsCard = (): NodeDetailsCard[] => {
+      let twinId, accountId, relay;
+      if (twin.value) {
```
zaelgohary, line 56: What is twin doesn't have value?

packages/playground/src/explorer/utils/helpers.ts
```diff
@@ -137,3 +151,23 @@ export const getQueries = (mixedFilters: MixedFilter): Partial<NodesQuery> => {
...
   return options;
 };
+
+export async function getFarms(queries: Partial<FarmsQuery>): Promise<Pagination<Farm[]>> {
+  try {
+    const farms = await gridProxyClient.farms.list(queries);
+    return farms;
+  } catch (error) {
+    console.error("An error occurred while requesting farms:", error);
+    throw error;
```
zaelgohary, line 161: Can we show a toast with the error here?

packages/playground/src/explorer/utils/helpers.ts
```diff
@@ -137,3 +151,23 @@ export const getQueries = (mixedFilters: MixedFilter): Partial<NodesQuery> => {
...
+  }
+}
+
+export async function getTwins(queries: Partial<TwinsQuery> = {}): Promise<Pagination<Twin[]>> {
```
zaelgohary, line 165: This function is created to get the twin id of a specific farm. Can we change its name to getTwinId and edit the return value to the first object in the list?

packages/playground/src/explorer/utils/types.ts
```diff
@@ -1,51 +1,38 @@
...
-export type FilterInputs = {
-  nodeId: NodeInputFilterType;
-  farmIds: NodeInputFilterType;
-  farmName: NodeInputFilterType;
-  country: NodeInputFilterType;
-  freeSru: NodeInputFilterType;
-  freeHru: NodeInputFilterType;
-  freeMru: NodeInputFilterType;
+export type FarmFilterOptions = {
+  farmId: number | undefined;
```
zaelgohary, line 19: We can set the properties to be optional instead of setting the type to undefined.

packages/playground/src/explorer/utils/types.ts
```diff
@@ -1,51 +1,38 @@
...
+  farmName: string | undefined;
 };
 
-export const optionsInitializer: FilterOptions = {
+export const nodeOptionsInitializer: NodeFilterOptions = {
   gateway: undefined,
   gpu: undefined,
   page: 1,
   size: 10,
```
zaelgohary, line 27: page & size shouldn't be hard coded.

packages/playground/src/explorer/utils/types.ts
```diff
@@ -1,51 +1,38 @@
...
   status: capitalize(NodeStatus.Up) as NodeStatus,
 };
 
+export const farmOptionsInitializer: FarmFilterOptions = {
+  farmId: 1,
```
zaelgohary, line 32: can't we set the filters to be optional?

packages/playground/src/explorer/views/farms.vue
```diff
@@ -0,0 +1,105 @@
...
+const filterOptions = ref<FarmFilterOptions>(farmOptionsInitializer);
+const mixedFarmFilters = ref<MixedFarmFilter>({ inputs: filterFarmInputs.value, options: filterOptions.value });
+const isFormLoading = ref<boolean>(true);
+const isValidForm = ref<boolean>(false);
+
+const _getFarms = async (queries: Partial<FarmsQuery>) => {
+  loading.value = true;
+  isFormLoading.value = true;
+
+  const { count, data } = await getFarms(queries);
```
zaelgohary, line 43: Can we do the query in a try & catch? In case catch entered, we can show a toast to the user with the error. check createCustomToast.

packages/playground/src/explorer/views/farms.vue
```diff
@@ -0,0 +1,105 @@
...
+        <filters
+          v-model="filterFarmInputs"
+          @update:model-value="inputFiltersReset"
+          :form-disabled="isFormLoading"
+          v-model:valid="isValidForm"
+        />
+        <FarmsTable :items="farms" :loading="loading" :selectedFarm="selectedFarm" @open-dialog="openDialog" />
+      </v-col>
+    </v-row>
+    <farmDialog v-if="selectedFarm" :openDialog="isDialogOpened" :farm="selectedFarm" @close-dialog="closeDialog" />
```
zaelgohary, line 14: close dialogue should be triggered with any event that makes the dialogue close
```suggestion
    <farmDialog v-if="selectedFarm" :openDialog="isDialogOpened" :farm="selectedFarm" @close-dialog="closeDialog"  @update:model-value="closeDialog" />
```

packages/playground/src/explorer/views/nodes.vue
```diff
@@ -172,22 +172,22 @@ export default {
...
-      mixedFilters.value.options.gpu = undefined;
-      mixedFilters.value.options.gateway = undefined;
-      mixedFilters.value.options.page = 1;
-      mixedFilters.value.options.size = 10;
+    const inputFiltersReset = (nFltrNptsVal: FilterNodeInputs) => {
+      MixedNodeFilters.value.inputs = nFltrNptsVal;
+      MixedNodeFilters.value.options.status = NodeStatus.Up;
+      MixedNodeFilters.value.options.gpu = undefined;
+      MixedNodeFilters.value.options.gateway = undefined;
+      MixedNodeFilters.value.options.page = 1;
```
zaelgohary, line 189: page & size should be set in the table itself and could be passed here instead of being hardcoded.

packages/playground/src/types/index.ts
```diff
@@ -1,8 +1,18 @@
...
 
+import type { AsyncRule, SyncRule } from "@/components/input_validator.vue";
 import type { INode } from "@/utils/filter_nodes";
 
 import type * as validators from "../utils/validators";
+// Input attrs
+export type InputFilterType = {
+  label: string;
+  placeholder: string;
+  value?: string | undefined;
```
zaelgohary, line 11: No need to set it to undefined.

zaelgohary (changes requested):

packages/playground/src/explorer/components/node_details_cards/farm_details_card.vue
amiraabouhadid: This component isn't just for farms, this card is used for the nodes page as well so when it is used for the nodes page, the node prop is passed. When it is used for the farms page the farm prop is passed. Hope this clears it up.

packages/playground/src/explorer/components/node_details_cards/farm_details_card.vue
amiraabouhadid: This is needed for handling the card contents based on which page it is being used in (i.e nodes page or farms page)

packages/playground/src/explorer/components/node_details_cards/twin_details_card.vue
amiraabouhadid: handled by adding default values if twin is undefined, thanks

packages/playground/src/explorer/views/nodes.vue
amiraabouhadid: this is outside the scope of this PR as it is in explorer nodes and not farms , please create an issue for it

packages/playground/src/explorer/utils/types.ts
amiraabouhadid: please include these lines in the issue as well

mohamedamer453 (commented):
The table is showing only 50 farms while there's 4k+ farms on devnet.

image

amiraabouhadid:
> The table is showing only 50 farms while there's 4k+ farms on devnet.
> 
> image

The 50 limit is implemented from the gridproxy

packages/playground/src/utils/filter_farms.ts
```diff
@@ -0,0 +1,49 @@
...
+export const inputsInitializer: FilterFarmInputs = {
+  farmId: {
+    label: "Farm IDs",
+    placeholder: "e.g. 1,2,3.",
+    rules: [
+      [isNumeric("Should be a number"), min("The ID should be larger then zero.", 1), isInt("should be an integer")],
+    ],
+    type: "text",
+  },
+  name: {
```
mohamedamer453, line 19: The farm name filter is not working correctly as it only returns the farm if the exact name is entered. and if a part of the name was entered it return all farms even if the entered string doesn't match any of the returned farms.

[Screencast from 21-11-23 15:31:56.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/101194226/92ce0293-3e5c-4970-bb6a-9f54d011cff0)

packages/playground/src/explorer/views/farms.vue
```diff
@@ -0,0 +1,217 @@
...
+  selectedFarm.value = item;
+  isDialogOpened.value = true;
+};
+
+const closeDialog = () => {
+  isDialogOpened.value = false;
+};
+const headers: VDataTableHeader = [
+  { title: "ID", key: "farmId" },
+  { title: "Name", key: "name" },
```
mohamedamer453, line 167: Sorting is working now but there's a weird behavior, if i switched from the first page and sorted the id or name it will then return to the first page and sort the values that are available there instead of sorting the values in the selected page.

[Screencast from 21-11-23 15:23:40.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/101194226/9f1a1e26-8e2f-464b-b2a1-b88d4810e90f)

amiraabouhadid: it is actually not weird behaviour to go back to page 1 when sorting is selected please refer to vuetify server side table [docs](https://vuetifyjs.com/en/components/data-tables/server-side-tables/#server-side-paginate-and-sort)

mohamedamer453: Indeed after checking the [docs](https://vuetifyjs.com/en/components/data-tables/server-side-tables/#data-table-server-side-tables) i found that this is the normal behavior for server side tables in vuetify

packages/playground/src/utils/filter_farms.ts
```diff
@@ -0,0 +1,49 @@
+import type { InputFilterType } from "../types";
+import { isAlphanumeric, isInt, isNumeric, min, validateResourceMaxNumber } from "./validators";
+
+export type FilterFarmInputs = {
+  farmId: InputFilterType;
+  name: InputFilterType;
+  pricingPolicyId: InputFilterType;
+  totalIps: InputFilterType;
+};
+export const inputsInitializer: FilterFarmInputs = {
```
mohamedamer453, line 10: The behavior of error messages in the filters is incorrect, for some cases when an incorrect value is entered it shows two messages saying `Failed to get farm` and it shows multiple errors in the console and in other cases as seen in the video below for the farms name filter it doesn't show anything.

[Screencast from 22-11-23 13:59:06.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/101194226/a5e934bb-8d64-405d-8b2d-e179b3343592)

mohamedamer453 (commented):
Currently The farm name search will only return the farm if its exact name is entered and if a sub-string of its name is entered it will return `No data available` is this the expected behavior for this filter? to search only by the exact name of the farm?

[Screencast from 22-11-23 13:53:23.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/101194226/9469eb2a-9c00-4f5b-a89e-960d791d9bc7)

amiraabouhadid:
> Currently The farm name search will only return the farm if its exact name is entered and if a sub-string of its name is entered it will return `No data available` is this the expected behavior for this filter? to search only by the exact name of the farm?
> 
>  Screencast.from.22-11-23.13.53.23.webm

This is the same behaviour as in the nodes page filter

packages/playground/src/utils/filter_farms.ts
amiraabouhadid: fixed, good catch

mohamedamer453: The behavior of the error messages was fixed but the filtering itself is now broken.

[Screencast from 22-11-23 14:35:37.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/101194226/29494bf5-a93b-4fe0-9e2f-44557249dfbe)

amiraabouhadid: fixed, please check

AlaaElattar:
- Why the toast appears ? the filtration is invalid already 
[Screencast from 11-22-2023 03:20:56 PM.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/54944491/49c3a5b9-b9f1-49e6-943e-d6dcebf3631a)

AlaaElattar (changes requested):
- there shouldn't be arrow with the name 
- what do u think ?
image

packages/playground/src/utils/filter_farms.ts
mohamedamer453: The original issue where the error message appeared twice + multiple errors in the console is back.

[Screencast from 22-11-23 15:30:38.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/101194226/8080c094-219f-4649-b83a-cfbbd50e7f40)

amiraabouhadid:
> * there shouldn't be arrow with the name
> * what do u think ?
>   image

sometimes people want to sort by farm name, it is an additional helpful feature

Mahmoud-Emad (approved):
Good job

AlaaElattar (approved):
All working fine

AlaaElattar:
- Here u're handling already while fetching data.
- But shouldn't there be a time out like in my case i changed network to offline. ( for example show msg and reload button as in dedicated nodes and explorer farms)
[Screencast from 11-21-2023 10:45:12 AM.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/54944491/43db5737-6b22-4edc-9d14-33ab9d7f9356)

packages/playground/src/utils/filter_farms.ts
```diff
@@ -0,0 +1,53 @@
...
+    placeholder: "e.g myfarm",
+    rules: [[isAlphanumeric("Farm name should be made of either numbers or letters")]],
+    type: "text",
+  },
+  pricingPolicyId: {
+    label: "Pricing Policy",
+    placeholder: "e.g 1",
+    rules: [
+      [
+        isNumeric("This field accepts numbers only.", { no_symbols: true }),
```
mohamedamer453, line 34: Inconsistency between error messages for the same validation error.

image

packages/playground/src/utils/filter_farms.ts
```diff
@@ -0,0 +1,53 @@
...
+  totalIps: InputFilterType;
+};
+export const inputsInitializer: FilterFarmInputs = {
+  farmId: {
+    label: "Farm IDs",
+    placeholder: "e.g. 1,2,3.",
+    rules: [
+      [
+        isNumeric("Should be a number", { no_symbols: true }),
+        min("The ID should be larger then zero.", 1),
```
mohamedamer453, line 17:
```suggestion
        min("The ID should be larger than zero.", 1),
```

packages/playground/src/utils/filter_farms.ts
```diff
@@ -0,0 +1,53 @@
...
+    rules: [[isAlphanumeric("Farm name should be made of either numbers or letters")]],
+    type: "text",
+  },
+  pricingPolicyId: {
+    label: "Pricing Policy",
+    placeholder: "e.g 1",
+    rules: [
+      [
+        isNumeric("This field accepts numbers only.", { no_symbols: true }),
+        min("The pricing policy should be larger then zero.", 1),
```
mohamedamer453, line 35:
```suggestion
        min("The pricing policy should be larger than zero.", 1),
```

packages/playground/src/utils/filter_farms.ts
```diff
@@ -0,0 +1,53 @@
...
+    ],
+    type: "text",
+  },
+  totalIps: {
+    label: "Total  PubIPs",
+    placeholder: "e.g. 1,2,3.",
+    rules: [
+      [
+        isNumeric("Should be a number"),
+        min("Total Public IP should be larger then zero.", 1),
```
mohamedamer453, line 47:
```suggestion
        min("Total Public IP should be larger than zero.", 1),
```

zaelgohary:
I think there's sth wrong with the filter as it searches within this specific page instead of searching in all farms. Maybe we can create a new request instead of filtering in the existing farms?

image


- [ ]  Farm name might include "-" or "_"  so I suggest to update the filter accordingly.

image

image


- [ ] Total IPs filter doesn't work

image

- [ ] Also, there is only 1 pricing policy, so I don't think we need to filter by it.

mohamedamer453 (approved):

AlaaElattar (approved):

amiraabouhadid:
> * [ ]  I think there's sth wrong with the filter as it searches within this specific page instead of searching in all farms. Maybe we can create a new request instead of filtering in the existing farms?
> 
> image
> 
> * [x]   Farm name might include "-" or "_"  so I suggest to update the filter accordingly.
> 
> image
> 
> image
> 
> * [x]  Total IPs filter doesn't work
> 
> image
> 
> * [x]  Also, there is only 1 pricing policy, so I don't think we need to filter by it.

filter searches all farms and a request is made for every valid search , it is working fine for me 
[ here ](https://www.loom.com/share/f79eeaea6209491a87a0d3c23e0bd461)

zaelgohary:
- [ ] It shows no farms but I have a farm with the same name.

image

image

- [ ] Also, when I set the filters then go to another page then return back to farms, filters won't reset.

[node-filter.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/40770501/b694ed61-fe95-439b-9683-7c7aa7a9b28b)

amiraabouhadid:
@zaelgohary the filter is working fine for zee-farm
Screenshot from 2023-11-23 16-55-24
thanks for the comment, I have added a beforeUnMount action to clear filter inputs.

zaelgohary (approved):

Issue #1237, opened 2023-10-18
loading doesn't stop if the node is not responding


AhmedHanafy725:
https://github.com/threefoldtech/tfgrid-sdk-ts/assets/41957921/b572de6a-8edf-4d41-b539-3af5f4d9ea9e

Mahmoud-Emad:
The issue happened because of the the grid prox, but of course, it needs to be handled from our side

Mahmoud-Emad:
PS: Waiting on new style will be provided by @ehab-hassan

ehab-hassan:
https://www.figma.com/proto/b50GilnYwLrhwOUIhfzHNA/dashboard-Nav?type=design&node-id=1267-2525&t=MoDFEiajud7rXrwN-0&scaling=contain&page-id=0%3A1&starting-point-node-id=124%3A196

Mahmoud-Emad:
- Update the build method in the proxy client to set timeout 5000 on the request and return an error in case the node is not responding, 
- Update the node details with a re-request button to let the user re-request the node in case of not responding.
- Updated the dialog header size
- Updated the style of error loading the node

A-Harby:
Verified,
Devnet 8c32384.

When the node is not responding or an error occurs, the try again message is shown.
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/0a697cbc-8d66-4899-840e-31c11c39838b)

[TC2252 - Offline/Standby Nodes Information](https://app.testlodge.com/a/26076/projects/40893/suites/234374?expand_section=422340#case_4150461)

ramezsaeed:
Verified:
- Devnet
- 000eea2

PR #1242, opened 2023-10-19
loading doesn't stop if the node is not responding

Mahmoud-Emad:
### Description

- Update the build method in the proxy client to set timeout 5000 on the request and return an error in case the node is not responding, 
- Update the node details with a re-request button to let the user re-request the node in case of not responding.
- Updated the dialog header size
- Updated the style of error loading the node 

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1237

### Screenshots

- image

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation


packages/gridproxy_client/src/builders/abstract_builder.ts
```diff
@@ -40,23 +40,37 @@ export abstract class AbstractBuilder<T> {
...
-      out.push(`${this.__mapper[key]}=${this.__queries[key]}`);
-    }
+      const query = out.length > 0 ? `?${out.join("&")}` : "";
+      const response = await fetch(`${this.uri}${path}${query}`, {
+        signal: controller.signal,
+      });
 
-    const query = out.length > 0 ? `?${out.join("&")}` : "";
-    return fetch(`${this.uri}${path}${query}`);
+      clearTimeout(timeoutId);
```
0oM4R, line 67: is it better to make this line in `finally` ?

packages/playground/src/explorer/components/node_details.vue
```diff
@@ -99,35 +114,57 @@ export default {
...
+    async function requestNode() {
+      if (props.nodeId > 0) {
+        try {
           loading.value = true;
-          const _node: GridNode = await getNode(nodeId, props.options);
+          const _node: GridNode = await getNode(props.nodeId, props.options);
           node.value = _node;
           router.push({ path: route.path, query: { nodeId: node.value.nodeId } });
+        } catch (err) {
+          console.error(err);
```
0oM4R, line 139: do we need to log the error ?

packages/gridproxy_client/src/builders/abstract_builder.ts
```diff
@@ -40,23 +40,35 @@ export abstract class AbstractBuilder<T> {
     }
   }
 
-  public async build(path: string): Promise<Response> {
+  public async build(path: string, timeout = 5000): Promise<Response> {
```
AhmedHanafy725, line 43: I think 5 seconds is small as if the filters are applied the request might take more time, especially on mainnet

packages/gridproxy_client/src/builders/abstract_builder.ts
```diff
@@ -40,23 +40,35 @@ export abstract class AbstractBuilder<T> {
...
 
-      this.__validator[key](this.__queries[key]!);
-      out.push(`${this.__mapper[key]}=${this.__queries[key]}`);
+      const query = out.length > 0 ? `?${out.join("&")}` : "";
+      const response = await fetch(`${this.uri}${path}${query}`, {
+        signal: controller.signal,
+      });
+      return response;
+    } catch (error: any) {
+      throw new Error(error);
```
AhmedHanafy725, line 68: why do we catch the error if we are going to throw it at the end?

packages/playground/src/explorer/components/node_details.vue
```diff
@@ -99,35 +114,56 @@ export default {
...
+    function closeDialog(newValue: boolean) {
+      isError.value = false;
+      errorMessage.value = "";
+      emit("close-dialog", newValue);
+    }
+
+    async function requestNode() {
+      if (props.nodeId > 0) {
+        try {
           loading.value = true;
```
AhmedHanafy725, line 134: I think we should set `isError.value` to `false` here in case the user clicks on `try again` to reset it

packages/gridproxy_client/src/builders/abstract_builder.ts
Mahmoud-Emad: Would you like to make it configurable?

packages/playground/src/explorer/components/node_details.vue
```diff
@@ -99,35 +114,56 @@ export default {
...
+    const errorMessage = ref("");
     const route = useRoute();
     const node = ref<GridNode>(nodeInitializer);
 
-    watch(
-      () => props.nodeId,
-      async (nodeId: number) => {
-        if (nodeId > 0) {
+    function closeDialog(newValue: boolean) {
+      isError.value = false;
```
0oM4R, line 126: this make the card appear for a while. i think it is better to reset the value before the request

packages/playground/src/explorer/components/node_details.vue
```diff
@@ -41,7 +54,7 @@
           <v-col cols="12" md="6" sm="8">
             <interfaces-details-card :node="node" />
           </v-col>
-          <v-col v-if="node.cards.length" cols="12" md="6" sm="8">
+          <v-col v-if="node.cards && node.cards.length" cols="12" md="6" sm="8">
```
0oM4R, line 57: What do you think about `node?.cared?.length` ?

packages/playground/src/explorer/components/node_details.vue
```diff
@@ -99,35 +114,56 @@ export default {
...
-        if (nodeId > 0) {
+    function closeDialog(newValue: boolean) {
+      isError.value = false;
+      errorMessage.value = "";
+      emit("close-dialog", newValue);
+    }
+
+    async function requestNode() {
+      if (props.nodeId > 0) {
+        try {
```
0oM4R, line 133: we can reset the error and error msg after this line

packages/playground/src/explorer/components/node_details.vue
```diff
@@ -141,4 +177,7 @@ export default {
 .v-toolbar__content {
   justify-content: end !important;
 }
+.v-toolbar__content {
+  height: 48px !important;
```
0oM4R, line 181: are we sure that this style doesn't have any unwanted effect in some other place ?

0oM4R (commented):
it's very nice, Good job 
just small questions and suggestions that may help

packages/playground/src/explorer/components/node_details.vue
Mahmoud-Emad: It'll effect only on the component

0oM4R (approved):

packages/playground/src/explorer/utils/helpers.ts
```diff
@@ -129,7 +129,7 @@ export const getQueries = (mixedFilters: MixedFilter): Partial<NodesQuery> => {
     freeHru: mixedFilters.inputs.freeHru.value ? toBytes(+mixedFilters.inputs.freeHru.value!) : undefined,
     freeMru: mixedFilters.inputs.freeMru.value ? toBytes(+mixedFilters.inputs.freeMru.value!) : undefined,
     freeSru: mixedFilters.inputs.freeSru.value ? toBytes(+mixedFilters.inputs.freeSru.value!) : undefined,
-    hasGpu: mixedFilters.options.gpu,
+    hasGpu: mixedFilters.options.gpu ? mixedFilters.options.gpu : undefined,
```
0oM4R, line 132: is that necessary?

AhmedHanafy725 (approved):

packages/playground/src/explorer/utils/helpers.ts
Mahmoud-Emad: Yes it is

PR #1263, opened 2023-10-23
Fix grid_client tests

MohamedElmdary:
Add disabledMigration prop in clientConfig to avoid migration if needed


packages/grid_client/tests/client_loader.ts
```diff
@@ -30,6 +30,7 @@ async function getClient(): Promise<GridClient> {
     network: config.network,
     mnemonic: config.mnemonic,
     storeSecret: config.storeSecret,
+    disableMigration: true,
```
AhmedHanafy725, line 33: The migration shouldn't be disabled while running the tests.

AhmedHanafy725:
another fix was done on #1316

PR #1266, opened 2023-10-23
Development 313 enhance ui2

ehab-hassan:
Removed all containers and made it take the full width to make it consistent with tables in Nodes and Dedicated Nodes 
I also added a border and fixed some colors and font sizes. 
image
image
image
image
image
image
image
image
image


packages/playground/src/explorer/views/nodes.vue
```diff
@@ -240,3 +240,8 @@ export default {
   },
 };
 </script>
+<style>
```
zaelgohary, line 243: Can we make this scoped style?

packages/playground/src/portal/dedicated_nodes_view.vue
```diff
@@ -23,4 +23,14 @@ export default {
 };
 </script>
 
-<style></style>
+<style>
```
zaelgohary, line 26: Also, can we make this scoped style?

zaelgohary (changes requested):
Great job! just some minor comments that need to be taken into consideration.

maayarosama (approved):

zaelgohary (approved):

Issue #1213, opened 2023-10-15
(Mainnet verification) Listing vms takes much time to load on mainnet


AhmedHanafy725:
Require investigation into why it takes so long to finish loading.

## background

it took around 1 minute to list 2 deployments in the listing page

AlaaElattar:
- still investigating but, sometimes it takes about 0.4 second and sometimes 8 seconds.

AlaaElattar:
- I did investigated with 4 deployments and that's what i got


![Screenshot from 2023-10-24 11-13-52](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/54944491/f89eaaaa-f943-42ea-bda8-432a157fce82)


![Screenshot from 2023-10-24 11-14-12](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/54944491/1ad013f7-08ed-48f5-a35d-171d696d7180)

- I believe that it took time in playground as there're 2 deployments failed to load.
- Also there's something that may cause problem if there're many deployments , that the pagination not handled from server side.

AlaaElattar:
- I added race condition and timeout for 5 seconds instead of 60 and it worked fine
- But there's smth more simple is to add timeout from function in `grid_client` working on it.

AlaaElattar:
- I added new pr against development with all changes that user can check number, name, and nodeId of failed deployments.
- waiting for review

khaledyoussef24:
env used: mainnet 
version : 2.2.0
[Screencast from 01-16-2024 01:31:10 PM.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/21bcbfc3-c6d0-4d63-bb01-a57aee49b03c)
wait time does not exceed 20 seconds and is acceptable now

PR #1268, opened 2023-10-24
Development 313 failed deployments fix

maayarosama:
### Description

making sure that chunk3 isn't empty before adding it to failedDeployments
### Changes

List of changes this PR includes

### Related Issues

https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1267
### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation
- [ ] Code format and docstrings
- [ ] Screenshots/Video attached (needed for UI changes)

PR #1274, opened 2023-10-24
Reduce vm Listing time in main net

AlaaElattar:
### Description
- Listing vms takes much time if any deployments failed to load.
- There's max 60 seconds trial before saying we failed to get the deployments

### Changes

- I added race condition and timeout for 5 seconds instead of 60.

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1213

https://github.com/threefoldtech/tfgrid-sdk-ts/assets/54944491/b3c3c867-5698-4252-a175-f0fd85b9e491


AlaaElattar:
- I will also apply the same for k8s.

zaelgohary:
I tried to load deployments and it showed me this err: 

image

I think this issue is already fixed in another PR.

AlaaElattar:
> I tried to load deployments and it showed me this err:
> 
> image
> 
> I think this issue is already fixed in another [PR](https://github.com/threefoldtech/tfgrid-sdk-ts/pull/1268)

yes that's because chunk 3, already fixed in the other pr.

AhmedHanafy725:
the fix is already done on the development branch

PR #1278, opened 2023-10-25
Fix time in Loading deployments main net. 

AlaaElattar:
- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1213
- make listing vms && k8s faster (it takes max 10 seconds before timeout)
- show number, name and node id of failed deployments

image

- If failed

image
image

- In case the node is not reachable

image
image

- same for k8s 

image

image


zaelgohary:
I was able to show the error without throttling my network. 

image

The warning is misleading as we couldn't show the deployment because we sat a timeout before retrieving the response. Not sure if this is the best solution, but I don't think we should show an error to the user instead of waiting a bit more.

AlaaElattar:
- This issue fixed here https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1291
- Now if we couldn't get the deployment data we will just display its name.

image

image

xmonader (approved):

xmonader (approved):

PR #1279, opened 2023-10-25
Add delete gateways from vm/fvm 

MohamedElmdary:
### Description
While deleting deployment we need to delete it's related gateways.

### Changes
- Add deleting gateways from fvm/vm while deleting them
- Fix deleting gateway from other solution (peertube, ..etc) as the projectName was changed
- Renamed gateways in fvm/vm to domain

### Related Issues
https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1055

### Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation

### DIFF
--- a/packages/playground/src/utils/delete_deployment.ts
+++ b/packages/playground/src/utils/delete_deployment.ts
@@ -3,7 +3,7 @@ import type { GridClient } from "@threefold/grid_client";
 import { ProjectName } from "@/types";
 
 import { loadVM } from "./deploy_vm";
-import { getSubdomain } from "./gateway";
+import { getSubdomain, loadDeploymentGateways } from "./gateway";
 import { updateGrid } from "./grid";
 
 export interface DeleteDeploymentOptions {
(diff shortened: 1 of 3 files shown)


packages/playground/src/utils/delete_deployment.ts
```diff
@@ -21,11 +21,20 @@ export async function deleteDeployment(grid: GridClient, options: DeleteDeployme
     }
   }
 
-  /* Delete gateway */
+  /* Start Delete gateway */
+
+  /* For fvm/vm */
+  if (isVm(options.projectName)) {
+    await deleteVmGateways(grid);
```
zaelgohary, line 28: Changes should be applied for all gateway naming in the changed files.

packages/playground/src/utils/delete_deployment.ts
```diff
@@ -3,7 +3,7 @@ import type { GridClient } from "@threefold/grid_client";
 import { ProjectName } from "@/types";
 
 import { loadVM } from "./deploy_vm";
-import { getSubdomain } from "./gateway";
+import { getSubdomain, loadDeploymentGateways } from "./gateway";
```
zaelgohary, line 6: Shouldn't this function change to loadDeploymentDomains?

zaelgohary (changes requested):

packages/playground/src/utils/delete_deployment.ts
MohamedElmdary: No, because we're using name domain for only fvm/vm but reset solution using gateway

packages/playground/src/utils/delete_deployment.ts
MohamedElmdary: We already handling other solution
https://github.com/threefoldtech/tfgrid-sdk-ts/pull/1279/files/0217c0145c3decf2d3743ea2f3eaa35a2f1e5e5a#diff-8a184bc478d27b259428cae59743ec4826372f23d6241ad87264806ce51d865dR31

zaelgohary (approved):

AhmedHanafy725 (approved):

Issue #1277, opened 2023-10-25
contacts view: include node id and node status


xmonader:
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/64129/857d1a88-0964-4ce4-a9f6-9af5d35fa676)

0oM4R:
status: 
![Screenshot from 2023-10-25 16-30-19](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/62248851/12040bcc-5d10-43fd-abf7-56d5a9e32454)

now i can get node status but can't handle standby nodes as it's supported on girdproxy client on 3.13; 
will handle name contracts tmw

0oM4R:
__Blocked__: can't handle standby nodes as it's supported on girdproxy client on 3.13;

0oM4R:
__Status__: PR is ready for review,

A-Harby:
Verified, Devnet 881667d.
Node ID and node status are now displayed.
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/8562fd24-df37-43bc-925d-65850e0a1653)

PR #1288, opened 2023-10-25
Contract view: add node details

0oM4R:
### Description

Show node id and its status in contracts table 

check node id and node status columns  
Screenshot from 2023-10-29 13-24-30

- couldn't test standby status but this is how it will look.
Screenshot from 2023-10-29 13-36-01

### Changes
- change gridproxy client version 
- add logic to get nodes status 
- add columns for node id and node status 

### Related Issues

[#1277](https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1277)

### Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation

### DIFF
--- a/packages/playground/src/weblets/tf_contracts_list.vue
+++ b/packages/playground/src/weblets/tf_contracts_list.vue
@@ -173,7 +173,7 @@ const headers: VDataTableHeader = [
   { title: "Solution Name", key: "solutionName" },
   { title: "Created At", key: "createdAt" },
   { title: "Expiration", key: "expiration" },
-  { title: "Node ID", key: "nodeId", sortable: false },
+  { title: "Node ID", key: "nodeId" },
   { title: "Node Status", key: "nodeStatus", sortable: false },
   { title: "Details", key: "actions", sortable: false },
 ];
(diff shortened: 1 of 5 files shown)


packages/playground/src/weblets/tf_contracts_list.vue
```diff
@@ -260,11 +279,35 @@ async function onDelete() {
...
+    if (typeof nodeId !== "number") return {};
+    const status = (await gridProxyClient.nodes.byId(nodeId)).status;
+    return { [nodeId]: status };
+  });
+
+  const resultsArray = await Promise.all(resultPromises);
+
+  return resultsArray.reduce((acc, obj) => Object.assign(acc, obj), {});
+}
+function getNodeStateColor(state: NodeStatus): string {
```
zaelgohary, line 294: Set a default case or the return type might be undefined.

zaelgohary (changes requested):

packages/playground/src/weblets/tf_contracts_list.vue
0oM4R: the nodeStatus have only three cases and each case is handled, i believe there is no need to set default case as long as we didn't have any unhanded NodeStatus value, what do you think ?

zaelgohary (approved):

amiraabouhadid (approved):
looks good 
image

Issue #1220, opened 2023-10-16
Add notification for contracts on offline nodes 


xmonader:
it can affect the user balance specially in case of having public ip reservation

Mahmoud-Emad:
Should be implemented in https://github.com/threefoldtech/tfgrid-sdk-ts/pull/1239

Mahmoud-Emad:
### Issue update

- Added a notification helper that will notify the user every 15 minutes if there are deployments on an offline node.
- Updated the toast warning background, icon color, and text color in dark/light mode.

Mahmoud-Emad:
Worked on some comments

- Created the `GqlContracts` interface in https://github.com/threefoldtech/tfgrid-sdk-ts/pull/1255 and used it
- Enhanced the performance by removing unneeded loops

Mahmoud-Emad:
### Update 

- Working on some enhancement that may enhance the performance

Mahmoud-Emad:
### Issue Update

- Returned back to in-progress because some changes requested, `Support links in the notification`

Mahmoud-Emad:
### Issue Update 

#### Support links in the notification pop-up
- ![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/57001890/aaed474d-6d6e-4c87-a451-454d2dbd9ad0)

should be supported in https://github.com/threefoldtech/tfgrid-sdk-ts/pull/1297 and the PR is ready for review.

Mahmoud-Emad:
### Issue update

- Worked on some enhancement -> Close the toast when navigating to the contracts list page
- Updated the `BillReports` constant name

Mahmoud-Emad:
@maayarosama 

I have kept this issue in the 'In Progress' column because the newly implemented feature involving the `<Link on the number of contracts>` needs to be merged into `3.13` to avoid conflicts. The last merge was performed against version `3.12`. Should we continue to work on this and keep it in the 'In Progress' column until the necessary changes are completed?

A-Harby:
verified, 
I deployed a VM, Kubernetes master and worker, and Funkwhale (solution) on the same node 21, and when the node went offline, I got notification for the contract on that node.

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/15906eaa-1306-40bf-9a5e-7a00d12c4754)
[...]

PR #1290, opened 2023-10-26
Support offline node contract notification on `mainnet`

Mahmoud-Emad:
This is open to be attached to the `mainnet` as a patch release.

### The original issue description

> #### Description

> - Added a notification helper that will notify the user every 15 minutes if there are deployments on an offline node.
> - Updated the toast warning background, icon color, and text color in dark/light mode.   

> #### Related Issues

> - https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1220

> #### Screenshots

> - image

### DIFF
--- a/packages/grid_client/src/clients/tf-grid/contracts.ts
+++ b/packages/grid_client/src/clients/tf-grid/contracts.ts
@@ -1,6 +1,7 @@
 import {
   ContractLockOptions,
   Contracts,
+  ExtrinsicResult,
   GetDedicatedNodePriceOptions,
   SetDedicatedNodeExtraFeesOptions,
 } from "@threefold/tfchain_client";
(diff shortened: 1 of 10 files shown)


packages/grid_client/src/clients/tf-grid/contracts.ts
```diff
@@ -107,28 +165,30 @@ class TFContracts extends Contracts {
...
-              createdAt = contracts[0]["createdAt"];
-              duration = (billReports[0]["timestamp"] - createdAt) / 3600;
+              createdAt = +contracts[0].createdAt;
+              duration = (billReports[0].timestamp - createdAt) / 3600;
               break;
             }
           }
         }
-        if (!duration) {
+        if (!duration!) {
```
0oM4R, line 191: Why do we need the second exclamation mark?
 if duration is null, undefined, or any other falsy value, it will be set to 1.

packages/grid_client/src/clients/tf-grid/contracts.ts
```diff
@@ -76,7 +133,8 @@ class TFContracts extends Contracts {
         rentContractsCount: rentContractsCount,
       });
 
-      return response["data"];
+      const data = response["data"] as GqlContracts;
+      return data;
```
0oM4R, line 137: We are not using the data variable anywhere; we can return its value directly.
```suggestion
    return response["data"] as GqlContracts;
```

0oM4R (changes requested):

Mahmoud-Emad:
@0oM4R 

This patch code is from the officially merged code that you have already reviewed at https://github.com/threefoldtech/tfgrid-sdk-ts/pull/1239. I am unable to modify the base code here to avoid conflicts when merging the complete code later. Could you please address your comments here and consider opening an issue instead of proposing changes that you believe will enhance our performance? Thank you.

0oM4R:
> Could you please address your comments here and consider opening an issue instead of proposing changes that you believe will enhance our performance?

the mentioned PR doesn't have any changes related to grid_client. not sure if you are mentioning the right pr but you can go on

Mahmoud-Emad:
@0oM4R 

Please take a look again :)

0oM4R (approved):

Issue #1281, opened 2023-10-25
Dockerfile and chart for new stats


maayarosama:
### Description
Add an image,  chart and workflows for new stats package

hossnys:
- where is this image to be created ? this repo or tf-images ? 
- where is that new stats package ? on which branch ?
- is there a docs for how it's being built ?
- workflows to be added here also ?

xmonader:
- tf-images is for solutions only, it should be in this repo
- it needs to be configurable with a tag
- should be in the packages/stats
- for the workflows, do as you see fit for image builds, autodeploy, .. etc

hossnys:
all comments on the PR resloved and tested chart and Dockerfile workflow on a fork and it works fine .

A-Harby:
The last release failed: https://github.com/threefoldtech/tfgrid-sdk-ts/actions/runs/7371704245.

AhmedHanafy725:
the last releases worked fine. please recheck it.

A-Harby:
Verified on the new workflows: https://github.com/threefoldtech/tfgrid-sdk-ts/actions/workflows/new_stats_build.yaml
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/27e681cd-4347-4f88-8ffa-55013d02464a)

PR #1304, opened 2023-10-30
add dockerfile and chart for new stats

hossnys:
### Description

adding new-stats package Dockerfile , chart , and workflows for building docker docker image .

### Changes

- add Dockerfile 
- add chart.yaml
- add values.yaml 
- add build workflow
- add docker image workflow 

### Related Issues

https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1281

### DIFF
--- a/.github/workflows/new_stats_build.yaml
+++ b/.github/workflows/new_stats_build.yaml
@@ -1,0 +1,45 @@
+# This workflow will do a clean install of node dependencies, cache/restore them, build the source code using yarn build:app
+
+name: New Stats Build
+
+on:
+  push:
+    branches:
+      - development
+    paths:
+      - "packages/new_stats/**"
+  pull_request:
+    branches:
+      - development
+    paths:
+      - "packages/new_stats/**"
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    env:
+      NODE_OPTIONS: "--max-old-space-size=8192"
+
+    strategy:
+      matrix:
+        node-version: [18.x]
+        # See supported Node.js release schedule at https://nodejs.org/en/about/releases/
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Setup Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v2
+        with:
+          node-version: ${{ matrix.node-version }}
+          cache: "yarn"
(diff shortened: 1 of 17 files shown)


zaelgohary (commented):
I think this package should be named stats and the old one's name should be changed to old-stats till it's archived.

hossnys:
> I think this package should be named stats and the old one's name should be changed to old-stats till it's archived.

we can keep the current naming till we get rid off the old one , then we rename it in next version as to avoid getting any of old images deployed .

0oM4R (commented):
### update not an issue; was running docker build in the new_stats directory
image
I tried to build the image locally and got that error

.github/workflows/new_stats_build.yaml
```diff
@@ -0,0 +1,45 @@
...
+          cache-dependency-path: "**/yarn.lock"
+
+      - name: Install dependencies
+        run: |
+          yarn install
+
+      - name: Build
+        run: |
+          lerna run build --no-private
+          yarn workspace @threefold/newstats build
```
0oM4R, line 45: i believe it should be `newstats` only without `@threefold`

packages/new_stats/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM node:18 AS build
+
+WORKDIR /app
+
+COPY . /app/
+
+# Install yarn and build project
+RUN yarn install
+RUN yarn lerna run build --no-private && yarn workspace @threefold/newstats build
```
0oM4R, line 9: same here docker build fails on this step 'error Unknown workspace "@threefold/newstats".'

0oM4R (changes requested):

packages/new_stats/Dockerfile
AhmedHanafy725: any package should be prefixed with `@threefold` in the sdk, So we should fix the package name

.github/workflows/new_stats_build.yaml
hossnys: as per @AhmedHanafy725 comment ,this prefix should be added from your side

0oM4R (approved):

amiraabouhadid (approved):

Issue #1291, opened 2023-10-26
Bug in loading deploymenst empty dialog


AlaaElattar:
- In k8s it said that it had 1 deployments failed, but the dialog was empty.

![Screenshot from 2023-10-26 10-17-14](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/54944491/169a0992-0c20-48aa-84e9-f3c1f843d342)

AlaaElattar:
- fixed and waiting for review .

A-Harby:
I tried deploying FullVM, MicroVm, and Funkwahle, and all the deployments were successful and lasted in the deployment list, but at the same time, they were also listed in the failed deployment list.

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/81c0bdc7-767d-4c11-87b8-49f8ec9973fd)
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/54e191db-b4c1-401f-9000-2c8b1776adac)
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/3c612c95-3237-46d1-8c25-c3dc33067b0e)

maayarosama:
Include contract Id of the failed deployment

A-Harby:
No, what I mean is that every time I deploy a solution, it's deployed successfully, and at the same time, it is shown in the failed deployment list.

https://drive.google.com/file/d/1Cn3g9mK5S-RLcrtGrSSD1HkomgeyJ3By/view?usp=sharing

AlaaElattar:
still working on it

AlaaElattar:
- Fixed and waiting for review.

A-Harby:
Verified, Devnet 2b364e9.

I tried many deployments, and none were shown on failed deployments as before.
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/17c4e30d-9eae-42a1-a3e5-2a23a728212c)

PR #1317, opened 2023-10-31
Fix bug of listing vms && add contract Id.

AlaaElattar:
- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1291
- Fix bug of deploying solution and find it in failed deployments while it also appears in list.
- Add contracts of failed deployments.

image

Screenshot from 2023-10-31 11-00-33

### DIFF
--- a/packages/playground/src/components/k8s_deployment_table.vue
+++ b/packages/playground/src/components/k8s_deployment_table.vue
@@ -15,17 +15,20 @@
         <v-card-title style="color: #ffcc00; font-weight: bold">Failed Deployments</v-card-title>
         <v-divider color="#FFCC00" />
         <v-card-text>
-          <ul>
-            <li v-for="deployment in failedDeployments" :key="deployment.name">
-              {{
-                deployment.nodes.length > 0
-                  ? `${deployment.name} on node${deployment.nodes.length > 1 ? "s" : ""}: ${deployment.nodes.join(
-                      ", ",
-                    )}`
-                  : deployment.name
-              }}
-            </li>
-          </ul>
+          <li v-for="deployment in failedDeployments" :key="deployment.name">
+            {{
+              deployment.nodes.length > 0
+                ? `${deployment.name} on node${deployment.nodes.length > 1 ? "s" : ""}: ${deployment.nodes.join(", ")}`
+                : deployment.name
+            }}
+            with contract id:
+            <template v-if="deployment.contracts && deployment.contracts.length > 0">
+              <span v-for="contract in deployment.contracts" :key="contract.contract_id">
+                {{ contract.contract_id }}&nbsp;
+              </span>
+            </template>
+            <template v-else> No contracts found. </template>
+          </li>
         </v-card-text>
         <v-card-actions class="justify-end">
           <v-btn @click="showDialog = false" class="grey lighten-2 black--text" color="#FFCC00">Close</v-btn>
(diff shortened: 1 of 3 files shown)


packages/playground/src/components/k8s_deployment_table.vue
```diff
@@ -15,17 +15,20 @@
...
-            </li>
-          </ul>
+          <li v-for="deployment in failedDeployments" :key="deployment.name">
+            {{
+              deployment.nodes.length > 0
+                ? `${deployment.name} on node${deployment.nodes.length > 1 ? "s" : ""}: ${deployment.nodes.join(", ")}`
+                : deployment.name
+            }}
+            with contract id:
+            <template v-if="deployment.contracts && deployment.contracts.length > 0">
```
0oM4R, line 25: if there is no contracts, the text will be `with contract id: No contracts found.` ?

packages/playground/src/components/k8s_deployment_table.vue
```diff
@@ -15,17 +15,20 @@
...
+              <span v-for="contract in deployment.contracts" :key="contract.contract_id">
+                {{ contract.contract_id }}&nbsp;
+              </span>
+            </template>
+            <template v-else> No contracts found. </template>
```
zaelgohary, line 30: Is this a valid case?

packages/playground/src/components/k8s_deployment_table.vue
AlaaElattar: - u're right, it doesn't make sense

packages/playground/src/components/k8s_deployment_table.vue
AlaaElattar: I believe it's not .

packages/playground/src/components/k8s_deployment_table.vue
```diff
@@ -15,17 +15,19 @@
...
+          <li v-for="deployment in failedDeployments" :key="deployment.name">
+            {{
+              deployment.nodes.length > 0
+                ? `${deployment.name} on node${deployment.nodes.length > 1 ? "s" : ""}: ${deployment.nodes.join(", ")}`
+                : deployment.name
+            }}
+            with contract id:
+            <template v-if="deployment.contracts && deployment.contracts.length > 0">
+              <span v-for="contract in deployment.contracts" :key="contract.contract_id">
+                {{ contract.contract_id }}&nbsp;
```
0oM4R, line 27: typo ?

packages/playground/src/components/k8s_deployment_table.vue
0oM4R: i guess we need to set the `with contract id` inside the if

zaelgohary:
There is one case where we won't be able to get the contract id which is if encryption didn't happen. In this case, we need to show a more meaningful message like "unknown contract id as it was not decrypted".

AlaaElattar:
> There is one case where we won't be able to get the contract id which is if encryption didn't happen. In this case, we need to show a more meaningful message like "unknown contract id as it was not decrypted".

will it be good to show msg next to deployment name or just not to show `with contract` as we do with nodeId.

zaelgohary:
> > There is one case where we won't be able to get the contract id which is if encryption didn't happen. In this case, we need to show a more meaningful message like "unknown contract id as it was not decrypted".
> 
> will it be good to show msg next to deployment name or just not to show `with contract` as we do with nodeId.

Sure we can do that.

AlaaElattar:
> > > There is one case where we won't be able to get the contract id which is if encryption didn't happen. In this case, we need to show a more meaningful message like "unknown contract id as it was not decrypted".
> > 
> > 
> > will it be good to show msg next to deployment name or just not to show `with contract` as we do with nodeId.
> 
> Sure we can do that.

done

packages/playground/src/components/k8s_deployment_table.vue
AlaaElattar: I removed it .

zaelgohary (approved):

MohamedElmdary (approved):

AhmedHanafy725 (approved):

Issue #1300, opened 2023-10-29
Deploy button enabled before checking pool storage


zaelgohary:
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/40770501/dbb08f06-d8a5-4437-af83-4de784925545)

After some time, the button was disabled as there was no enough storage.

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/40770501/469e3c9c-a3d4-40e2-879c-8cf3506ff88c)

0oM4R:
still investigating can't reproduce the issue i a clear way, but i think that the issue occurs on a slow connection when the grid take long time to load before validate the node's storage pool; will check tmw with alaa she had a [pr](https://github.com/threefoldtech/tfgrid-sdk-ts/pull/1305/files) related to this issue

0oM4R:
this can be noticeable on slow connection: a delay between validating the nodeid value and validating storage pool, on that delay the nodid field is valid, which make the deploy button enabled. 
on initial case where there is no node selected, the selectedNode is undefined until the validating storage is finished, then it update the value, but on the mentioned delay if you click on the deploy button will deploy on nodeId undefined as mentioned on #1244 
Not sure what the proper solution is. I will try to fix this today isA

0oM4R:
deploy button is working fine know, but we need to handle network errors, any error  while validating storage pool will show the error massage : `Couldn't fit the required disks in Node ${validatingNode.nodeId} storage pools, please select another node`
will check tmw if this will be in separate issue or in the same pr

A-Harby:
Verified, Devnet.
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/a368c04c-f6d6-4889-8a38-ae72225a70fe)
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/6cae32d1-4d1f-4d91-beea-024ea9c94a23)

Issue #1295, opened 2023-10-26
timeout contract details in 10s


xmonader:
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/64129/386840d5-ef56-4554-be25-7110ef38f1ce)

that seems to be a contract we can't get the details of, it keeps spinning for 60s then 
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/64129/f30d19d9-372d-4609-979f-9181b5d53a51)

It's better to be cancelled in the same UI without moving to another screen

0oM4R:
__Status__: 
![Image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/62248851/2382ac8a-2fef-4358-9bd6-320dae610977)
and will check with Ehab today isA

0oM4R:
will make another pr to set request time out to 10s

0oM4R:
changes applied on pr #1336

A-Harby:
Verified, Devnet 86b9355.

The error button appeared to retry with a failed notification, including the contract ID.
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/c84c8622-6ee7-4a7f-9457-9c078e03c830)

PR #1336, opened 2023-11-01
fix deploy button enabled while node is node validated  

0oM4R:
### Description
add async validation rule to check capacity pool

### Changes
- add async validation rule to chcek node's storage pool
- set default rmb request time out to be 10s 

### Related Issues
 - #1295 
 - #1300 

### Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation

### DIFF
--- a/packages/playground/src/components/select_node.vue
+++ b/packages/playground/src/components/select_node.vue
@@ -15,6 +15,7 @@
     <input-validator
       ref="validator"
       :rules="[validators.required('Node id is required.')]"
+      :async-rules="[() => validateNodeStoragePool(selectedNode)]"
       :value="selectedNode"
       #="{ props }"
     >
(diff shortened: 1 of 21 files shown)


MohamedElmdary:
It's not a good idea to update all solutions instead we can move validateNodeStoragePool to `Node(v-autocomplete)` in rules which will make sure it's valid before setting valid to true

0oM4R:
> It's not a good idea to update all solutions instead we can move validateNodeStoragePool to `Node(v-autocomplete)` in rules which will make sure it's valid before setting valid to true

@MohamedElmdary  applied, thanks!

packages/playground/src/components/select_node.vue
```diff
@@ -276,27 +264,30 @@ async function loadNodes(farmId: number) {
...
   farmManager?.setLoading(true);
   validator.value?.setStatus(ValidatorStatus.Pending);
   pingingNode.value = true;
   try {
-    await grid.capacity.checkNodeCapacityPool({
-      nodeId,
-      ssdDisks: disks,
-      rootfsDisks: [rootFileSystemSize],
+    const grid = await getGrid(profileManager.profile!);
+    await grid!.capacity.checkNodeCapacityPool({
```
amiraabouhadid, line 274: do an if(grid) instead of the forbidden null assertion. Also please remove unused imports on lines 82 and 83.

amiraabouhadid (changes requested):

packages/playground/src/components/select_node.vue
0oM4R: done, we can enhance the error message later on https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1341

MohamedElmdary (approved):

amiraabouhadid (approved):

Issue #1338, opened 2023-11-01
Notification colors lost in incognito mode


Mahmoud-Emad:
### Issue Description

**Problem:**
When using the playground in incognito mode, the theme does not load correctly. This issue arises because incognito mode does not have access to local storage, which is used to retrieve the current theme settings.

### Screenshots

**1. Normal Mode (Regular Browsing)**

- ![Normal Mode](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/57001890/7f965806-679f-455f-ba57-9a350eebd624)

**2. Incognito Mode (Private Browsing)**

- ![Incognito Mode](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/57001890/030151c9-1883-411c-ab3e-ee07d246172d)

Mahmoud-Emad:
### Work Completed:
The issue should be fixed in the mentioned PR.

### Work in Progress (WIP):
No work in progress related to this issue till now.

### Investigation and Solution:
was checking if there is a `theme` key stored in the `local-storage` and yes I found that if the user does not click on the theme mode we don't set it, maybe we need to set a default value for it, so this is the solution

A-Harby:
verified.
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/ddcbcd93-0dd5-4654-a860-235dd1cdebca)
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/7fc11f1f-d69b-44e1-be35-2743205c4fef)

PR #1339, opened 2023-11-01
Set defaut theme if there is no theme selected by the user.

Mahmoud-Emad:
### Description

The problem occurred due to the absence of a default value for the `app theme`. Without a user-selected theme, the system cannot determine the appropriate class, such as `mosha__icon__dark__warning`, to load based on the theme.

### Changes

set an initialization of the app theme variable that is stored in the local storage with the default `dark` value.

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1338

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/packages/playground/src/App.vue
+++ b/packages/playground/src/App.vue
@@ -130,6 +130,7 @@ const profileManager = useProfileManager();
 
 const openProfile = ref(true);
 const hasActiveProfile = computed(() => !!profileManager.profile);
+checkAppTheme();
 
 watch(
   () => $route.meta,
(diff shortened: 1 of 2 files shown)


MohamedElmdary:
Updating theme check should be enough in custom_toast.ts file

MohamedElmdary:
Use css variables from vuetify which inject in html tag

maayarosama (approved):

MohamedElmdary:
> Use css variables from vuetify which inject in html tag

It wasn't as easy as it should be to customize mosh toast ... we might need to create our own toast later to get control for ui requirements

MohamedElmdary (approved):

Issue #1091, opened 2023-09-12
(Mainnet Verification) Improve ux when selecting nodes


sabrinasadik:
We should start off by showing all nodes and then narrowing it down with filters if needed.

e.g. If I'm looking for a node it doesn't matter to me if it's certified or not so I should be able to see all of them. If then I want a certified, dedicated or GPU node, I can choose those so it filters the results.

RoberL72:
I thought the idea was for certified nodes to show up first.
And; how do we inform the benefits of silver and gold farms?

maayarosama:
WIP:
- trying to show all certified/uncertified nodes at first without applying filters

maayarosama:
Still investigating

maayarosama:
Couldn't investigate on development as graphql was down

maayarosama:
Work Completed:
- Added a chip for certified nodes
- All nodes load regardless of the certified filter at first
Work in Progress (WIP):

Investigation and Solution:
- Investigating why does it retrieve the certified nodes even when the filter is false

maayarosama:
Work Completed:

- Added a watch for the selectedNode so it the calculator recalculates.
- Added a warning within the calculator div telling the user that a certified node is being used.
- pr is pending review.

Work in Progress (WIP):
Investigation and Solution:

maayarosama:
Work Completed:

- Added a condition to recalculate cost when the certified filter is off but the user chooses a certified node
- pr is ready for review

Work in Progress (WIP):
Investigation and Solution:

A-Harby:
Verified, 
Devnet, 3fc8787.

All the nodes are now shown for the user first, and three toggles have been added: GPU, dedicated, and certified to specify node type; also, node region, country, farm name, and ID can be specified.
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/23fdf720-b440-4ded-8a7d-118686d7abc9)

ramezsaeed:
cannot find a certified node to test the calculator is working as expected
- devnet
- ebf75a3

![Screenshot from 2023-12-24 12-28-12](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/16083592/feb27c47-2a34-4101-a566-da0d07c37f39)
[...]

PR #1366, opened 2023-11-05
Adding a certified chip and loading all nodes regardless of filters a…

maayarosama:
### Description

Initializing certified to undefined so all nodes could load at the start regardless of the certified filter and adding a chip for certified nodes
Screenshot from 2023-11-05 14-08-24
### Changes
- Initializing certified variable to undefined in all solutions
- Adding a certified variable to INode interface
- Adding a chip and a chip color to certified nodes

### Related Issues
#1091
### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation


zaelgohary (commented):
Now all certified nodes are being returned even if it's set to false.

image

mohamedamer453 (commented):
Selecting a certified node without enabling the certified filter will not update the calculated resource pricing which is supposed to be 25% higher for certified nodes.

- Selecting a certified node without enabling the certified filter

   image

- Selecting the same certified node but with the certified filter enabled.

   image

maayarosama:
> Now all certified nodes are being returned even if it's set to false.
> 
> image

As per a discussion with @AhmedHanafy725 this's in fact the desired behavior.

maayarosama:
> Selecting a certified node without enabling the certified filter will not update the calculated resource pricing which is supposed to be 25% higher for certified nodes.
> 
> * Selecting a certified node without enabling the certified filter
>   image
> * Selecting the same certified node but with the certified filter enabled.
>   image

Added a watch foe the selectedNode so that the calculator can recalculate when it changes.

zaelgohary (commented):
Please pull latest changes from base branch as I can't review.
image

zaelgohary (changes requested):
As @mohamedamer453  mentioned, Same node has 2 different prices if the certified chip is enabled.
image

image

zaelgohary:
Can you merge base branch? There was a build error and it was fixed.

mohamedamer453 (approved):

packages/playground/src/weblets/full_vm.vue
```diff
@@ -230,6 +231,15 @@ watch(
...
 );
+watch(
+  selectedNode,
+  selectedNode => {
+    if (selectedNode) {
+      console.log("selectedNode in fullvm: ", selectedNode);
+    }
+  },
+  { immediate: true },
+);
```
zaelgohary, line 242:
```suggestion
```

Can we remove this watcher? It's not needed.

zaelgohary (changes requested):

zaelgohary (approved):

Issue #1312, opened 2023-10-30
(Mainnet verifications) Select farms randomly while deploying any solution


AhmedHanafy725:
this can be done by 2 requests

- Get the available farms and their count and how many pages.
- Select a random page and get a random farm from this page.

0oM4R:
@AhmedHanafy725 can you mention the endpoint for the first request please?

AhmedHanafy725:
the same as the current one but set the `ret_count` to `true`

0oM4R:
Investigation and Solution:
- currently we are using grid client to get the farms and the `send` method returns only the `data` we are locking for a header 
- using gridproxy client is also an option but we will have to change the filteroptions type 
- another solution is to have competition between gridporxy to get the count, then getting the farms after that by the existing option.

0oM4R:
Work Completed:
I implemented the needed function in gridclient and its call on utils/getfarms
implemented the randomization logic that store the selected pages and pick a new random number that isn't in the selected list 


Work in Progress (WIP):
testing the current flow and have some issue with reset the selected array when the filters got updated, also there is some requests that I didn't catch why they were called.


Investigation and Solution:
- will work on the reset section
- the two requests take so long time to pick the farm, not sure if this is a general issue or something related to my network

0oM4R:
Work Completed:
all needed logic is done.

Work in Progress (WIP):
 - faced an issue while testing that the website keep making requests and didn't stop,
 

[Screencast from 06 نوف, 2023 EET 05:59:33 م.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/62248851/4833d980-0fbf-45ee-8ee2-5e8b1732d9e4)


- the `{ exclusiveFor: props.exclusiveFor }` and filtering the blocked farms on get farms count not sure how to handle this point

0oM4R:
## status 
Work Completed:
- requests issue is not related to the current solution and i opened an issue for it #1373 
- all needed logic are completed and pr #1370 is ready for review.

A-Harby:
Verified on Devnet,
Every time a solution is selected, the farms are randomized, and a node is shown from that farm.
I tried on devnet, and it only had 2 farms, and both showed after a few refreshings.
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/789de5f4-1a2f-434f-97b4-a89798decbfc)
[...]

PR #1370, opened 2023-11-06
available farms randomization 

0oM4R:
##### Please make sure to test this pr against mainnet because it has a large number of farms
Select random page based on the total available pages and the already visited pages

### Changes
- gridclient
  - create `sendWithFullResponse` function that returns the whole response not just the `data`
  - create `getFarmsCount` that uses the `sendWithFullResponse` function, the passed filters, `ret_count = true` and set page to 1; to get the total count of farms 
- playground
  - `getFarmsPages` uses the gridclient to get the total farms pages 
  - changed the flow of getting farms to be: 
      1.   -  on mount or `shouldBeUpdated == true`: reset the variables and get the farmCount then set page to a randomvalue then loadfarms, and store the current page to `selectedpages `array
      2. create a new random page number in the range of pageCount that does not exist in selectedpages array
      3. if the response of get farms is empty and we can load more [ this case is happens when the  farms on the current page are blocked farms and we didn't visit all available pages ] we set a new random value and `loadFarms` again

### Related Issues

- #1312 

### Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation

### DIFF
--- a/packages/grid_client/src/helpers/requests.ts
+++ b/packages/grid_client/src/helpers/requests.ts
@@ -13,4 +13,17 @@ async function send(method: Method, url: string, body: string, headers: Record<s
   }
   return response.data;
 }
-export { send };
+async function sendWithFullResponse(method: Method, url: string, body: string, headers: Record<string, string>) {
+  const options = {
+    method: method,
+    url: url,
+    data: body,
+    headers: headers,
+  };
+  const response = await axios(options);
+  if (response.status >= 400) {
+    throw Error(`HTTP request failed with status code: ${response.status} due to: ${response.data}`);
+  }
+  return response;
+}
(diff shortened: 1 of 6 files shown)


packages/grid_client/src/helpers/requests.ts
MohamedElmdary, line 1: Please use sendWithFullResponse in send to avoid duplicated code

Mahmoud-Emad (approved):

packages/grid_client/src/helpers/requests.ts
0oM4R: thanks, done

amiraabouhadid (approved):

MohamedElmdary (approved):

Issue #1306, opened 2023-10-30
Show notification in case of slow connectivity 


maayarosama:

mohamedamer453:
Update:

Work Completed:

- Moved the grid client object inside the try/catch block and now appropriate error messages are displayed when a network error happens. 
- Didn't add a separate notification since already a clear error message appears for the user and the nodes field doesn't infinitely load anymore as seen in #1374 


Status:

Pr is ready for review.

A-Harby:
Verified,
- Devnet
- 6d0ec3d
- [TC2307 - Error message for filter failure](https://app.testlodge.com/a/26076/projects/40893/suites/234374?expand_section=422336#case_4155955)

I tried with slow and by turning the network offline and the message error was shown to me and the validating stopped loading.
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/9a814597-4864-4e48-aa1b-1ebc0d675318)

ramezsaeed:
verified:
- Devnet
- 000eea2

Issue #995, opened 2023-08-15
Introduce error types in clients


AhmedHanafy725:
Some of the features are depending on checking if the error message contains part of the message which might be re-phrased and break the feature. So it's better to have well-defined error types to check against.

0oM4R:
#### Status
Investigation and solution:
We may implement a new common package to include the common functionality and types 
we can start with Errors types and class

[...]

PR #1374, opened 2023-11-07
fix nodes field infinite loading

mohamedamer453:
### Description

Fix the infinite loading in the Nodes Id field when a connection error happens.

### Changes

- Moved the grid client object inside the try/catch block to catch network errors instead of loading infinitely 

- Now when a network error happens, the appropriate error message will be displayed instead of loading infinitely.

image
 
image

### Related Issues

#1306 

### Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation

### DIFF
--- a/packages/playground/src/components/select_node.vue
+++ b/packages/playground/src/components/select_node.vue
@@ -218,9 +218,9 @@ async function loadNodes(farmId: number) {
   errorMessage.value = "";
   const filters = props.filters;
   farmManager?.setLoading(true);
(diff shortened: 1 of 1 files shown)


amiraabouhadid (approved):

0oM4R (approved):
I'm not comfortable with the error message, should start with upper case, but i think we will handle this on [#995](https://github.com/threefoldtech/tfgrid-sdk-ts/issues/995)

Issue #8, opened 2023-04-12
Support hex based seeds


scottyeager:
For the purposes of the `farmerbot`, it would be beneficial to support the use of hex based seeds in `grid3_client_ts`.  Many farms have been created using the TF Connect App, whereby the farmer only has a hex secret to work with for their twin and farm.

Since `rmb-rs` [already](https://github.com/threefoldtech/rmb-rs/commit/00fdc4b8b6c10882a4c81f08244de1974008e964) supports this and it's possible to achieve all necessary TF Chain signing activities with a hex secret, I think it should be possible to support it for this client.

AhmedHanafy725:
for the deployment part, it could be supported as well for signing the deployment but the issue is saving the deployment on the kvstore on tfchain. the mnemonic is used as a symmetric key to encrypt the values. so if a user has mnemonic and secret seed for the same account and used them to deploy. both accounts will see the contracts without any problem but the deployed list will be different between them as the keys encrypted by one of them won't be decrypted by the other one.

AhmedHanafy725:
related 
- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/127

AhmedHanafy725:
maybe a suggestion:

- if a mnemonic is provided, convert it to a secret seed.
- if it's a secret seed keep it as is
- always use the seed for the encryption

but this will require migrating all the keys to be encrypted by the seed instead of the mnemonic

MohamedElmdary:
Blocked on some appeared issues described in PR https://github.com/threefoldtech/tfgrid-sdk-ts/pull/1380#issuecomment-1801782457

A-Harby:
I tried using the hex secrets from the Threefold Connect app, but the activate button wasn't enabled.
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/26656ce3-e4cc-44f6-86fc-cdeaa85b0f7f)

Also, can you link the documents for the hex seeds in the issue?

khaledyoussef24:
am still facing the same issue
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/b30359fe-eaf6-45f3-9b4d-4ab978d3eed3)

AhmedHanafy725:
@khaledyoussef24 can you please add the steps to reproduce? how did you get the hex seed?

khaledyoussef24:
@AhmedHanafy725 
[...]

PR #1380, opened 2023-11-08
Support hex based seed

MohamedElmdary:
### Description
Support hex based seed in grid_client and migrate keys decryption to use new hex seed.

### Changes
- Migrate keys encryption from mnemonic(Aka. words) into hex seed.
- Update `rmb_direct_client`, `zos module` and `grid_client` to  load mnemonic as hex seed in keyring.
- Handle error in `playground` was thrown whenever we fail to load metadata due to unable to decrypt.

### Related Issues
- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/8

### Checklist
- [ ] Tests included
- [ ] Build pass
- [ ] Documentation


MohamedElmdary:
After Adding the require hex based seed support 2 issue appeared. I think both of them might be caused by the same reason

- When trying to get a deployment
image

- Trying to check if a node has enough storage pool
image

Mahmoud-Emad (approved):

Mahmoud-Emad (changes requested):
Cant we change the validation in the dashboard as we did in the playground?

image

packages/grid_client/src/helpers/utils.ts
```diff
@@ -36,4 +39,15 @@ function log(message) {
   console.log(JSON.stringify(message, null, 2));
 }
 
-export { generateString, getRandomNumber, randomChoice, randomSecret, randomSecretAsHex, randomNonce, log };
+function toHexSeed(mnemonic: string): string {
```
AhmedHanafy725, line 42: is the input only mnemonic? it can be hex seed, please rename it

packages/rmb_direct_client/lib/sign.ts
```diff
@@ -16,10 +15,7 @@ export function sign(payload: string | Uint8Array, signer: KeyringPair) {
   return sigPrefixed;
 }
 
-export async function createShared(pubKey: Uint8Array, mnemonic: string) {
-  const seed = await bip39.mnemonicToSeed(mnemonic);
-  const privateKey = new Uint8Array(seed).slice(0, 32);
-  const pointX = secp.getSharedSecret(privateKey, pubKey);
-  const key = await secp.utils.sha256(pointX.slice(1, 33));
-  return key;
+export function createShared(pubKey: Uint8Array, hexSeed: string) {
```
AhmedHanafy725, line 18: any user can use this client without using the grid client, we shouldn't make work on hex seed only. please handle the both cases

packages/grid_client/src/helpers/migration.ts
```diff
@@ -0,0 +1,37 @@
+import type { GridClient } from "../client";
+
+type GridClientCtor = { new (...args: any[]): GridClient; migrated: Set<string> };
+
+export async function migrateKeysDecryption(this: GridClient, GridClient: GridClientCtor): Promise<void> {
```
AhmedHanafy725, line 5: are we migrating decryption? please use a better name

AhmedHanafy725 (changes requested):

packages/grid_client/src/helpers/utils.ts
```diff
@@ -36,4 +39,15 @@ function log(message) {
...
+  if (validateMnemonic(mnemonic)) {
+    const seed = mnemonicToMiniSecret(mnemonic);
+    const keypair = box.keyPair.fromSecretKey(seed.slice(0, 32));
```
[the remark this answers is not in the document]
MohamedElmdary: So how to get secretKey? @AhmedHanafy725

packages/rmb_direct_client/lib/sign.ts
```diff
@@ -16,10 +16,14 @@ export function sign(payload: string | Uint8Array, signer: KeyringPair) {
...
 }
 
-export async function createShared(pubKey: Uint8Array, mnemonic: string) {
-  const seed = await bip39.mnemonicToSeed(mnemonic);
-  const privateKey = new Uint8Array(seed).slice(0, 32);
-  const pointX = secp.getSharedSecret(privateKey, pubKey);
-  const key = await secp.utils.sha256(pointX.slice(1, 33));
-  return key;
+export function createShared(pubKey: Uint8Array, hexSeedOrMnemonic: string) {
+  let privateKey: Uint8Array | string = hexSeedOrMnemonic.slice(2);
```
AhmedHanafy725, line 20: can we check its length and if it has 0x before slicing?

Mahmoud-Emad (approved):

Issue #1365, opened 2023-11-05
Error on canceling multiple contracts on the Playground


AbdelrahmanElawady:
### Description

Checking multiple contracts and canceling them return an error that the contracts don't exist.
![image_2023-11-05_12-27-46](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/60783742/e11cfe7d-690f-4dd1-b568-3705ee1efa54)
![image_2023-11-05_12-27-58](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/60783742/8a6c931b-45e7-496e-b6e0-62e15000bccf)
![image_2023-11-05_12-28-41](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/60783742/16d05f3e-6166-438f-9887-8f3c3a97b386)

### Steps to reproduce

- Check a contract
- Press on refresh button
- Check it again
- Delete the contract

AlaaElattar:
- so u deleted some contracts then refreshed it and found them again ? Is that the case?
- If yes, there's already message says `Deleting contracts may take a while to complete.`
- If not, please i need more details.

AlaaElattar:
- I misunderstood the issue, the case is true.
- It consider deleting the same contract twice.

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/54944491/2e88e7d4-8743-4106-96c9-a03e32debed9)

AlaaElattar:
### Work Completed:
It's handled that each time user will refresh, all previous selections will be removed

AlaaElattar:
### Work Completed:
I've applied comment pr, which was resetting `selectedContracts` to empty before start loading.

khaledyoussef24:
verified 
[Screencast from 12-26-2023 08:53:16 AM.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/7a47d33f-7ec1-4c87-bfac-8efef8faecd0)

testlodge:https://app.testlodge.com/a/26076/projects/40893/suites/234374?expand_section=422334#case_4150373

ramezsaeed:
Verified:
- Devnet
- 842ebef

PR #1381, opened 2023-11-08
Fix duplicate contracts when deleting

AlaaElattar:
### Description
- When user check on contract then refresh and check to delete it, It will appear to be deleted twice.

### Changes
- check `selectedContracts` gets empty when refresh

### Related Issues

https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1365

- Before 

https://github.com/threefoldtech/tfgrid-sdk-ts/assets/54944491/141d34ac-181c-4dd8-bc2e-9e98c700bef5

- After 

https://github.com/threefoldtech/tfgrid-sdk-ts/assets/54944491/d3b9c2c5-25b1-4791-9338-f6ffbc2e634f

### DIFF
--- a/packages/playground/src/weblets/tf_contracts_list.vue
+++ b/packages/playground/src/weblets/tf_contracts_list.vue
@@ -180,6 +180,7 @@ const headers: VDataTableHeader = [
 ];
 
 async function onMount() {
+  selectedContracts.value = [];
   loading.value = true;
   failedContractId.value = undefined;
   contracts.value = [];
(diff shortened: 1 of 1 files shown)


packages/playground/src/weblets/tf_contracts_list.vue
```diff
@@ -187,6 +187,7 @@ async function onMount() {
   contracts.value = await getUserContracts(grid.value!);
   nodeStatus.value = await getNodeStatus(nodeIDs.value);
   loading.value = false;
+  selectedContracts.value = [];
```
0oM4R, line 190: I think it is better to reset it before load the contracts, i mean line 185;
or at least before setting loading to false

0oM4R (approved):

AhmedHanafy725 (approved):

Issue #1341, opened 2023-11-02
check node storage pool invalid error message  


0oM4R:
We need to handle network errors; any error while validating the storage pool will show the error message: `Couldn't fit the required disks in Node ${validatingNode.nodeId} storage pools, please select another node`

_Originally posted by @0oM4R in https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1300#issuecomment-1789402379_

mohamedamer453:
Update:

Investigation and Solution:

- Still investigating the issue, was able to output the error messages that come from the grid client but when i do that the storage pools message gets overridden by the one from the client. 
- Still looking for a solution to output both the network error messages from the client and the desired message for the storage pools.

0oM4R:
I believe we should wait for this issue [#995](https://github.com/threefoldtech/tfgrid-sdk-ts/issues/995)

mohamedamer453:
Update:

Work Done:

- Added a condition to check for the returned error if it's a storage pools error it'll return a message mentioning that the storage pools are not enough and if it's a network error it'll return an error message mentioning that something went wrong while checking the storage pools.
- Could be improved once #995 is available since the current method checks for a part of the error message.

Status: Pr is ready for review.

maayarosama:
Verified on Devnet
![2024-01-02_11-54](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/66883096/587ca19a-568c-4420-92fc-85d15ef6854e)

ramezsaeed:
Verified:
- Devnet
- 228c723

PR #1383, opened 2023-11-08
separate nework errors and storage pools errors

mohamedamer453:
### Description

Separate network errors and storage pools errors while checking the storage pools of a node.

### Changes

- When the returned error is a storage pools error then it'll be:

   image

- When the returned error is a network error then it'll be:

   image

### Related Issues

#1341 

### Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation

### DIFF
--- a/packages/playground/src/components/select_node.vue
+++ b/packages/playground/src/components/select_node.vue
@@ -289,11 +289,11 @@ async function validateNodeStoragePool(validatingNode: INode | undefined) {
 
     if (e?.toString().includes("Cannot fit the required SSD disk with size")) {
       return {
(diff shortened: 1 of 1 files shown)


AlaaElattar (approved):

xmonader:
Should we mention storage pools at all to the endusers?

0oM4R:
> I don't think we should we mention storage pools at all to the endusers?
@xmonader 
I think yes, or something related to it; we can't say to the user this node is available and then say sorry that node is not available.
Also, we will add error types to the client later after that, we will be able to handle each error message separately.

0oM4R (approved):

Issue #1359, opened 2023-11-05
Show total spending of user's contracts


xmonader:
Should show the amount spend / hr and its equivalent / month somewhere in the contracts view

AlaaElattar:
- The amount spend/hr already exists in `Billing Rate` .

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/54944491/fc099bd7-b6b8-4bba-8729-864fa51a23ff)
 - so would u like to add another column for equivalents/month or what ?

xmonader:
My bad it wasn't clear enough. I'm trying to figure out how much all of the deployments are costing me e.g hourly or monthly (not sure UI/UX wise) I'm suggesting it could be shown after the table?  `X TFT/hr and X TFT/month` and maybe a tooltip button that shows the equivalent in dollars?

AlaaElattar:
### Work in Progress (WIP):
Still working on the issue, trying different ideas to choose best UI of showing total spending of user.

AlaaElattar:
### Work Completed:
I've added `v-alert` at end of table showing total cost of contracts per hours and its equivalent per month. In case the table is empty, the alert won't appear.

AlaaElattar:
### Work Completed:
- I've modified the `NormalizedContract` to calculate consumption as number not string, so i won't have to use regular expression.
- Just checking in table if consumption == 0, so I will show `No Data Available`.

ramezsaeed:
regarding to this comment,
is the https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1359#issuecomment-1801903260

is this implemented:
tooltip button that shows the equivalent in dollars?

AlaaElattar:
> regarding to this comment, is the [#1359 (comment)](https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1359#issuecomment-1801903260)
> 
> is this implemented: tooltip button that shows the equivalent in dollars?
- It's implemented like this.

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/54944491/ee5df622-373b-413b-9a57-093b584ea836)

ramezsaeed:
Verified:
- Devnet
- ebf75a3
- Testlodge: https://app.testlodge.com/a/26076/projects/40893/suites/234374?expand_section=422339#case_4161190

[...]

PR #1391, opened 2023-11-09
Total Cost of contracts per hour and month

AlaaElattar:
### Description
- Showing total cost per hour and month

### Changes

- Add alert at the end of table showing total cost per hour or month.
- If there's no contracts, the alert won't appear

### Related Issues

https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1359

Screenshot from 2023-11-09 11-34-39

### DIFF
--- a/packages/playground/src/weblets/tf_contracts_list.vue
+++ b/packages/playground/src/weblets/tf_contracts_list.vue
@@ -22,6 +22,11 @@
       no-data-text="No contracts found on this account."
       v-bind:onClick:row="loading || deleting ? undefined : onClickRow"
     >
+      <template #[`item.consumption`]="{ item }">
+        <p v-if="item.raw.consumption !== 0">{{ item.raw.consumption.toFixed(3) }} TFT/hour</p>
+        <p v-else>No Data Available</p>
+      </template>
+
       <template #[`item.state`]="{ item }">
         <v-tooltip
           v-if="item && item.value.state === ContractStates.GracePeriod"
@@ -320,11 +325,7 @@ function getNodeStateColor(state: NodeStatus): string {
 function getTotalCost(contracts: NormalizedContract[]) {
   totalCost.value = 0;
   for (const contract of contracts) {
-    const matching = contract.consumption.match(/(\d+(\.\d+)?) TFT\/hour/);
-    if (matching) {
-      const value = new Decimal(matching[1]);
-      totalCost.value = +new Decimal(totalCost.value).add(value);
-    }
+    totalCost.value = +new Decimal(totalCost.value).add(contract.consumption);
   }
   return +totalCost.value.toFixed(3);
 }
--- a/packages/playground/src/weblets/tf_contracts_list.vue
+++ b/packages/playground/src/weblets/tf_contracts_list.vue
@@ -77,7 +77,7 @@
       </template>
     </ListTable>
     <div class="pt-4">
-      <v-alert variant="tonal" color="secondary" class="pt-4" v-if="contracts.length !== 0"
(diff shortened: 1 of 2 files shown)


packages/playground/src/weblets/tf_contracts_list.vue
```diff
@@ -76,6 +76,11 @@
         </v-tooltip>
       </template>
     </ListTable>
+    <div class="pt-4">
+      <v-alert variant="tonal" color="secondary" class="pt-4" v-if="contracts.length !== 0"
+        >Total Cost: {{ totalCost }} TFT/hour ≈ {{ (totalCost * 24 * 30).toFixed(3) }} TFT/month</v-alert
```
amiraabouhadid, line 81: you can just do `v-if="contracts.length"`

amiraabouhadid:
is it supposed to give 0 here? there are many contracts but the cost displayed is 0. 
image

amiraabouhadid (changes requested):

AlaaElattar:
> is it supposed to give 0 here? there are many contracts but the cost displayed is 0. image

I can check also if the `totalCost` is zero so i won't show the dialog. what do u think of this ?

packages/playground/src/weblets/tf_contracts_list.vue
```diff
@@ -306,6 +314,18 @@ function getNodeStateColor(state: NodeStatus): string {
...
   }
 }
+
+function getTotalCost(contracts: NormalizedContract[]) {
+  totalCost.value = 0;
+  for (const contract of contracts) {
+    const matching = contract.consumption.match(/(\d+(\.\d+)?) TFT\/hour/);
+    if (matching) {
+      const value = parseFloat(matching[1]);
+      totalCost.value += value;
```
AhmedHanafy725, line 324: please use `Decimal` in any money calculations

zaelgohary (approved):

packages/playground/src/weblets/tf_contracts_list.vue
```diff
@@ -306,6 +315,18 @@ function getNodeStateColor(state: NodeStatus): string {
...
   }
 }
+
+function getTotalCost(contracts: NormalizedContract[]) {
+  totalCost.value = 0;
+  for (const contract of contracts) {
+    const matching = contract.consumption.match(/(\d+(\.\d+)?) TFT\/hour/);
+    if (matching) {
+      const value = new Decimal(matching[1]);
+      totalCost.value += +value;
```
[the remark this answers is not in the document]
AlaaElattar: so u want `totalCost` to be of type `Decimal` ?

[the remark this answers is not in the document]
AhmedHanafy725: no, you can do the calculation with Decimal and then convert it to a number at the end to show it to the user

AhmedHanafy725 (approved):

packages/playground/src/weblets/tf_contracts_list.vue
```diff
@@ -306,6 +315,18 @@ function getNodeStateColor(state: NodeStatus): string {
       return "warning";
   }
 }
+
+function getTotalCost(contracts: NormalizedContract[]) {
```
xmonader, line 319: I believe we should introduce the raw field instead along with the consumption string, so field `tftPerHour` on the normalized contract object, and drop the regex part

xmonader (changes requested):
Added a change request, let me know if you guys still want to proceed with the regular expression and why.

xmonader (approved):
Looks much cleaner now 💯 Thank you

amiraabouhadid (approved):

Issue #548, opened 2023-06-12
Contract list stuck loading in a failed connection


A-Harby:
### Description

When there's a connection error or slow internet connection, Contract list is stuck at loading like this and this error is shown in the console. 

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/fb756063-7740-452e-ade0-eaabf224dbcb)

xmonader:
Should we flag this as a known issue in the release notes?

AlaaElattar:
- I've tried multiple situations and they were handled 

![Screenshot from 2023-08-16 17-03-19](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/54944491/16b2eca6-ee32-4cbc-b1fe-2755cf41736a)

![Screenshot from 2023-08-16 17-06-19](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/54944491/c1251afc-5aea-4afa-b8c0-5a8ab6dbfd5f)

A-Harby:
Still happening, steps to reproduce:
1. Login and go to Contracts.
2. Change network to Slow 3G. 
3. Press Refresh button.
4. Repeat step 3 until it happens.

[Untitled_ Aug 17 2023 9_47 AM.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/e8b68541-27dd-485a-8183-577dfb0b8f3e)

It's not a common bug but it will happen if user has a bad network or grid proxy/graphql is having a connection error.

khaledyoussef24:
verified 
it never gets stuck may take a long time to load but works fine
[Screencast from 12-27-2023 06:54:26 PM.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/15030e74-be37-4bbe-8f71-ded4beb1a33b)

ramezsaeed:
Verified:
- Devnet
- 842ebef

This is a random issue,
I tried multiple times did't happen with me.
so please re-open if happen again.

PR #1392, opened 2023-11-09
handle loading contracts error

0oM4R:
### Description

- Added getGrid into try block
- Added an alert to show the error message
### Changes

if any error happen while listing the user contracts; An alert with the error message will appears, if the user click on refresh it will try to reconnect for like 1.5 min and will show the error.
this long time of reconnecting is probably related to the gridclient, will open an issue to improve this flow from the gridclient side 

[Screencast from 09 نوف, 2023 EET 12:51:16 م.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/62248851/79a76982-3a98-4853-8035-2635a252653e)

### Related Issues

#548 
### Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation

### DIFF
--- a/packages/playground/src/weblets/tf_contracts_list.vue
+++ b/packages/playground/src/weblets/tf_contracts_list.vue
@@ -13,7 +13,11 @@
         refresh
       </v-btn>
     </template>
+    <v-alert type="error" variant="tonal" class="mt-2 mb-4" v-if="loadingError">
+      {{ loadingError }}
+    </v-alert>
     <ListTable
+      v-else
       :headers="headers"
       :items="contracts"
       :loading="loading"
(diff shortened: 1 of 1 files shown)


AlaaElattar:
- Please check conflicts.

AlaaElattar:
- In this case what do u think if we showed only the error `Error listing contracts by twin id 86: AxiosError: Network Error` 
image

0oM4R:
> * In this case what do u think if we showed only the error `Error listing contracts by twin id 86: AxiosError: Network Error`
>   image

the only way to handle this case is to check on the error message, which is not recommended. later we will create error types on clients and we can handle each case, currently if we think we don't need the prefix will totally remove it from all cases

mohamedamer453 (approved):

packages/playground/src/weblets/tf_contracts_list.vue
```diff
@@ -178,16 +182,22 @@ const headers: VDataTableHeader = [
...
   { title: "Details", key: "actions", sortable: false },
 ];
-
+const loadingError = ref("");
 async function onMount() {
+  loadingError.value = "";
   selectedContracts.value = [];
   loading.value = true;
   failedContractId.value = undefined;
   contracts.value = [];
```
zaelgohary, line 191: Can we reset these values in finally instead?

zaelgohary (changes requested):

packages/playground/src/weblets/tf_contracts_list.vue
0oM4R: we need to reset some values before starting, like hide the error message and remove field contracts ids, 
as we are not repeating this step inside catch, I think its fine to have them inside try block, what do you think ?

0oM4R: some of them must be in try like error message; if we reset it in finally , we will never have an error message

AlaaElattar (approved):

zaelgohary (approved):

Issue #1352, opened 2023-11-02
Investigate the performance and time taken by the mass_deployments workflow


mohamedamer453:
### Description

Investigate the performance and how long it takes for the mass_deployments workflow to run.

The latest workflow [run](https://github.com/threefoldtech/tfgrid-sdk-ts/actions/runs/6727307503) took almost 3 hours for 100 deployments. mainly because some of the failed deployments took ~10 minutes to timeout as seen in https://github.com/threefoldtech/tfgrid-sdk-ts/actions/runs/6727307503/job/18284925859#step:6:3325

AlaaElattar:
### Work in Progress (WIP):
- I got errors at first then opened this issue https://github.com/threefoldtech/tfgrid-sdk-go/issues/473
- I'm still investigating but with deleting randomization in both farms and nodes until the issue is solved

AlaaElattar:
### Work in Progress (WIP):
- I'm trying to start by running only 10 deployments and track their time. 
- But I've got the same error for them all, `Error getting node x, didn't respond after 10 seconds` on dev.
- I've tried to run on qa but got same error for 10 deployments.
- That's all with removing `randomize` from both nodes and farms.

AlaaElattar:
### Work in Progress (WIP):
- I've increased timeout to 20 seconds.
- I got `Error: Error getting node 2: Error: Failed to send request to twinId 9 with command: zos.statistics.get, payload:  due to 102 code: 102` for all 10 deployments but without timeout part.

AlaaElattar:
### Work in Progress (WIP):
- Blocked till this issue is solved https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1379

xmonader:
Issue blocked on dysfunctional devnet

AlaaElattar:
### Investigation :
- If deployment created successfully will take avg 15 to 24 seconds.
- If failed because the error of sending request (code 102) will take 10 seconds (timeout of request).
- If failed because `Node 128 is not available for user with twinId: 214, maybe it's rented by another user or node is dedicated. use capacity planning with availableFor option.` will take avg 3 to 4 seconds. (That's an issue needs to be handled here or from gridproxy)

AlaaElattar:
### Work Completed::
- Script updated to deploy batches of vms together.
- For example, 1000 vms will be deployed each 100 batch together. If one on the batch failed, the whole batch will be failed.
- This will decrease the time alot.
[...]

PR #1421, opened 2023-11-15
Developemnt 313 mass deployment

AlaaElattar:
### Description

Make mass deployments up to 1000 workload, each batch (100 deployment) will get deployed together to reduce time and not deploying each vm one by one.

### Related Issues

https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1352


packages/grid_client/scripts/mass_deployments.ts
```diff
@@ -3,107 +3,201 @@ import {
...
   let failedCount = 0;
   let successCount = 0;
+  const batchSize = 50;
+  const totalVMs = 250;
+  const batches = totalVMs / batchSize;
+
+  // resources
+  const cru = 1;
+  const mru = 256;
+  const diskSize = 1;
```
AhmedHanafy725, line 51: no need for creating a disk for these vms

packages/grid_client/scripts/mass_deployments.ts
```diff
@@ -3,107 +3,201 @@ import {
...
-      nodeMRU: mru / 1024,
-      nodeSRU: diskSize + rootFs,
-      publicIp: publicIp,
-      availableFor: await grid3.twins.get_my_twin_id(),
-      randomize: true,
-    } as FarmFilterOptions);
-
-    if (farms.length < 1) {
-      throw new Error("No farms found");
+    const nodesPromises = Array.from({ length: batchSize }, async () => {
```
AhmedHanafy725, line 74: here we make many requests to the grid proxy to get the nodes. However, we can set all the farm ids we got in the request and just get a random page by using the `randomize` query

packages/grid_client/scripts/mass_deployments.ts
```diff
@@ -3,107 +3,201 @@ import {
...
+
+    // Batch Deployment
+    const batchVMs: MachinesModel[] = [];
+    for (let i = 0; i < batchSize; i++) {
+      const vmName = "vm" + generateString(8);
+
+      const disk1 = new DiskModel();
+      disk1.name = "d" + generateString(5);
+      disk1.size = diskSize;
+      disk1.mountpoint = "/newDisk1";
```
AhmedHanafy725, line 122: this disk should be removed

packages/grid_client/scripts/mass_deployments.ts
```diff
@@ -3,107 +3,201 @@ import {
...
   MachineModel,
   MachinesModel,
   NetworkModel,
   randomChoice,
+  TwinDeployment,
 } from "../src";
 import { config, getClient } from "./client_loader";
 import { log } from "./utils";
 
+async function pingNodes(grid3: GridClient, nodes: any[]): Promise<any[]> {
```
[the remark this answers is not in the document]
AlaaElattar: - u mean create interface of that ?
image

packages/grid_client/scripts/mass_deployments.ts
```diff
@@ -3,107 +3,201 @@ import {
...
+  const pingPromises = nodes.map(async node => {
+    const pingPromise = grid3.zos.pingNode({ nodeId: node.nodeId });
+    const timeoutPromise = new Promise((_, reject) =>
+      setTimeout(() => reject(new Error(`Timeout: Ping to node ${node.nodeId} took too long`)), 10000),
+    );
+
+    try {
+      const res = await Promise.race([pingPromise, timeoutPromise]);
```
AhmedHanafy725, line 24: instead of promise race, we can set the expiration to 10 sec in `pingNode` method

packages/grid_client/scripts/mass_deployments.ts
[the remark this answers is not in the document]
AlaaElattar: yes, but the results is not flattened yet also i need to check on res and error.

packages/grid_client/scripts/mass_deployments.ts
```diff
@@ -1,119 +1,211 @@
...
-  randomChoice,
+  NodeInfo,
+  TwinDeployment,
 } from "../src";
 import { config, getClient } from "./client_loader";
 import { log } from "./utils";
 
+async function pingNodes(
+  grid3: GridClient,
+  nodes: any[],
```
AhmedHanafy725, line 17:
```suggestion
  nodes: NodeInfo[],
```

packages/grid_client/src/primitives/nodes.ts
```diff
@@ -373,7 +374,7 @@ class Nodes {
       ipv6: options.accessNodeV6,
       gateway: options.gateway,
       certification_type: options.certified ? "Certified" : "",
-      farm_ids: [options.farmId],
+      farm_ids: options.farmIds ? options.farmIds : [options.farmId],
```
AhmedHanafy725, line 377: `farmId` is more specific than `farmIds`, so I think if the `farmId` is provided, we should use it and ignore the `farmIds`
```suggestion
      farm_ids: options.farmId ? [options.farmId] : ptions.farmIds,
```

packages/grid_client/src/primitives/nodes.ts
```diff
@@ -47,6 +47,7 @@ interface NodeInfo {
   updatedAt: string;
   total_resources: NodeResources;
   used_resources: NodeResources;
+  farm_ids: number[];
```
AhmedHanafy725, line 50: do we have this on the grid proxy nodes endpoint?

packages/grid_client/scripts/mass_deployments.ts
```diff
@@ -1,119 +1,211 @@
...
+  if (farms.length < 1) {
+    throw new Error("No farms found");
+  }
+
+  console.time("Total Deployment Time");
+
+  for (let batch = 0; batch < batches; batch++) {
+    console.time("Batch " + (batch + 1));
+
+    const nodesPromises = Array.from({ length: batchSize }, async () => {
```
AhmedHanafy725, line 72: do we still need to do that? we are already getting all batch nodes with on call

packages/grid_client/src/primitives/nodes.ts
AlaaElattar: you're right, it's not existing

AhmedHanafy725 (approved):

mohamedamer453 (commented):
Great work! i ran the script locally and from the workflow and both ran without any issues.

There's just a tiny comment, after the script is done the number of successful and failed deployments seems to be inaccurate. as seen from [this](https://github.com/threefoldtech/tfgrid-sdk-ts/actions/runs/7158412170/job/19490465493#step:6:5279) run on github, the number of successful deployments is 200 and the number of failed ones is 161, and as seen in the script itself the total number of deployments is 250, so i think there's an error with the counting logic.

mohamedamer453 (approved):
Verified from the latest run, the count now adds up to 250, great work!

Issue #1412, opened 2023-11-14
Error `maybe it's rented by another user or node is dedicated.` when dedploying any solution


A-Harby:
### Description

Keep getting that error when deploying any solution.
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/c0e87379-c6f1-47ab-9383-6b3aed7d2916)

AhmedHanafy725:
require: https://github.com/threefoldtech/tfgrid-sdk-go/issues/483

AlaaElattar:
### Work Completed:
- Condition updated when requesting nodes.

A-Harby:
Verified, 
Devnet, 3fc8787.

The `inDedicatedFarm` endpoint was added, and node searches are specified for shared farms only.
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/569f09fe-a18e-440e-ba0d-8bd999878efd)
And deploying is working fine too now.
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/0d149683-da53-4493-8a39-2497d58902b1)

PR #1425, opened 2023-11-16
Fix deployment error `maybe it's rented by another user or node is dedicated.`

AlaaElattar:
- Update condition of requesting nodes after being updated in gridproxy.
- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1412


packages/grid_client/src/primitives/nodes.ts
```diff
@@ -409,7 +409,7 @@ class Nodes {
   async nodeAvailableForTwinId(nodeId: number, twinId: number): Promise<boolean> {
     return send("get", urlJoin(this.proxyURL, `/nodes/${nodeId}`), "", {})
       .then(node => {
```
AhmedHanafy725, line 411: can we add the new field in the `NodeInfo` interface?

```suggestion
      .then(node: NodeInfo => {
```

AhmedHanafy725 (approved):

amiraabouhadid (approved):

PR #1417, opened 2023-11-14
Introduce error types in tfchain client

0oM4R:
### Description

introduce error types package and use it in tfchain client 

### Changes

- add types package 
- use errors from types package to throw erros in tfchain client

### Related Issues

#995

### Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation
- [x] Code format and docstrings
- [ ] Screenshots/Video attached (needed for UI changes)

PR #1433, opened 2023-11-19
Use error types in RMB direct client

0oM4R:
### Description

use types/errors in rmb direct client

### Changes

- added rmb error and connection error to types/errors
- use types/errors in rmb direct client f

### Related Issues
- #1417 
- #995 
### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation


packages/rmb_direct_client/lib/client.ts
```diff
@@ -103,16 +104,17 @@ class Client {
...
       this.twin = await this.tfclient.twins.get({ id: twinId });
       if (!twinId) {
-        throw new Error({ message: `Couldn't find a user for the provided mnemonic on this network.` });
+        throw new ValidationError(`Couldn't find a user for the provided mnemonic on this network.`);
       }
     } catch (e) {
       console.log(e);
-      throw new Error({ message: `Couldn't find a user for the provided mnemonic on this network.` });
+      //TODO is this a valid error message?
+      throw new ConnectionError(`Couldn't find a user for the provided mnemonic on this network.`);
```
AlaaElattar, line 112: I think this is a validation Error not connection. What do u think ?

AlaaElattar (changes requested):

packages/rmb_direct_client/lib/client.ts
0oM4R: i think this one is the validation related 
https://github.com/threefoldtech/tfgrid-sdk-ts/pull/1433/files/d251636ea203fb9362bd10a0efc8b6deac7d9203#diff-3290bb9d40c106d5afd7fbca7c8c3d3bc60d1025e6c5b6634f0e244138fd215cR107
 
but in catch from my usage in playground i got this error many times while network is offline so i mentioned in line 111 to check this error message

AlaaElattar: maybe we can change it to `Couldn't find a user for the provided mnemonic, check your connection`. or any msg with indication that the problem in the connection

AhmedHanafy725: IMO, this try-and-catch block was added for having a more user-friendly error. But now tfchain should throw the correct error with its type, no?

0oM4R: > IMO, this try-and-catch block was added for having a more user-friendly error. But now tfchain should throw the correct error with its type, no?

exactly but the question is what it the correct error in this case ?

0oM4R:
will draft this pr to user types in reject ;

packages/rmb_direct_client/lib/client.ts
```diff
@@ -324,7 +322,8 @@ class Client {
             }
           } else {
             this.responses.delete(requestID);
-            reject("invalid signature, discarding response");
+            //TODO what is the probate error type
```
0oM4R, line 325: @AhmedHanafy725 what do you think

AlaaElattar (approved):

AhmedHanafy725 (approved):

PR #1437, opened 2023-11-20
Fix build workflows errors

0oM4R:
change tsconfig in types package


zaelgohary (changes requested):
New Stats & Playground build still fail:

image

zaelgohary (approved):

MohamedElmdary (approved):

PR #1440, opened 2023-11-20
Use error types in graphql client 

0oM4R:
### Description

only in this cases we have invalid input so we should use ValidationError class  

### Changes

use `throw new ValidationError` instead of `panic`
> panic function is not used anymore but I will leave it in case if it needed 

### Related Issues

#995 

### Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation

### DIFF
--- a/packages/graphql_client/package.json
+++ b/packages/graphql_client/package.json
@@ -8,7 +8,8 @@
     "build": "tsc"
   },
   "dependencies": {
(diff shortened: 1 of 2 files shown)


amiraabouhadid (approved):

Mahmoud-Emad (approved):
Good job :)

Issue #1451, opened 2023-11-21
Grid tests workflow is failing after last update


A-Harby:
### Description

From the grid test [result](https://github.com/threefoldtech/tfgrid-sdk-ts/actions/workflows/grid_client_tests.yml) you can see that all the test runs are failing with the same error: `Couldn't create TFClient without store secret` 
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/084837a6-0473-484c-8938-2eaed5e68c81)

The error is usually caused when the mnemonics are not provided.

ramezsaeed:
we need to know the updates when run this manually.
do we have same results or not ?
@A-Harby

A-Harby:
> we need to know the updates when run this manually. do we have same results or not ? @A-Harby

Yes, when we run it manually or when we trigger it from the workflow, it works with no issue, and the tests are passing.
But the issue appears when the schedule runs it.

A-Harby:
Verified,

The test workflow is now running fine, and the tests are working with no errors. https://github.com/threefoldtech/tfgrid-sdk-ts/actions/workflows/grid_client_tests.yml

PR #1454, opened 2023-11-22
Development fix gird tests yml

A-Harby:
### Description

The network variable wasn't assigned any value when the workflow ran by the schedule.

https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1451

### DIFF
--- a/.github/workflows/grid_client_tests.yml
+++ b/.github/workflows/grid_client_tests.yml
@@ -2,7 +2,7 @@ name: Grid Client Tests
 
 on:
   schedule:
-    - cron: "0 0 * * *"
+    - cron: 0 0 * * *
   workflow_dispatch:
     inputs:
       network:
@@ -24,6 +24,9 @@ jobs:
       STORE_SECRET: secret
       MNEMONIC: ${{ secrets.MNEMONIC }}
     steps:
+      - name: set network for Schedule
+        if: ${{env.NETWORK}} == ""
+        run: echo "NETWORK=dev" >> $GITHUB_ENV
       - uses: actions/checkout@v4
       - name: Set up node 18
         uses: actions/setup-node@v4
--- a/.github/workflows/grid_client_tests.yml
+++ b/.github/workflows/grid_client_tests.yml
@@ -7,27 +7,26 @@ on:
     inputs:
       network:
         type: choice
-        description: 'which network to run tests on'
+        description: "Which network to run tests on"
         required: true
-        default: 'dev'
+        default: "dev"
         options: 
-        - 'dev'
-        - 'qa'
-        - 'test'
-        - 'main'
-
+        - "dev"
+        - "qa"
+        - "test"
+        - "main"
 jobs:
   tests:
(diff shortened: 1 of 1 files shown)


.github/workflows/grid_client_tests.yml
```diff
@@ -2,32 +2,34 @@ name: Grid Client Tests
...
         options: 
-        - 'dev'
-        - 'qa'
-        - 'test'
-        - 'main'
-
+        - "dev"
+        - "qa"
+        - "test"
+        - "main"
```
zaelgohary, line 17: Can we add custom option? We'll also need to provide all urls.

.github/workflows/grid_client_tests.yml
```diff
@@ -2,32 +2,34 @@ name: Grid Client Tests
...
       RMB_PROXY: true
       STORE_SECRET: secret
       MNEMONIC: ${{ secrets.MNEMONIC }}
     steps:
-      - uses: actions/checkout@v2
+      - name: set network for Schedule
+        if: ${{env.NETWORK}} == ""
+        run: echo "NETWORK=dev" >> $GITHUB_ENV
+      - uses: actions/checkout@v4
       - name: Set up node 18
```
zaelgohary, line 31: Maybe change this to Set up node version?

zaelgohary (changes requested):

.github/workflows/grid_client_tests.yml
A-Harby: Changed in [195c3d3](https://github.com/threefoldtech/tfgrid-sdk-ts/pull/1454/commits/195c3d369550071f470b1d4e66a524b74af13cc0)

.github/workflows/grid_client_tests.yml
A-Harby: The workflow is only for the default setups though, and any custom options would need to be manually run. These tests are meant to run in basic environments.

zaelgohary:
@A-Harby  I edited the workflow to test it and ran it twice and it failed both times. You can check the logs [here](https://github.com/threefoldtech/tfgrid-sdk-ts/actions/runs/6971016372/job/18971030282).

A-Harby:
> @A-Harby I edited the workflow to test it and ran it twice and it failed both times. You can check the logs [here](https://github.com/threefoldtech/tfgrid-sdk-ts/actions/runs/6971016372/job/18971030282).

The [issue](https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1451) I tried to solve was that the tests themselves didn't run, giving the error: `Couldn't create TFClient without store secret` and the cause of it was that the network wasn't assigned a value.

And from your logs, it shows that the issue is fixed because the tests started to run.

But the fail test from logs was caused by this error below, and it is fixed on https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1412, but it was merged to branch 3.13, and workflow runs on the latest development branch, which is why it's still failing.
image

zaelgohary (approved):

Issue #1203, opened 2023-10-12
New option to find a farm by its Name/ID.


A-Harby:
### Description

Add new option for the user to make it easier to find a farm by its Name or ID.

the used scenario now is that the user have to load all the farm from farm list then search for the farm name in that list, which does take a lot of time to load all the farms and waste a lot of requests just to find one farm.

zaelgohary:
I don't think manually filtering by farm name is practical as the searching purpose is to find a node to deploy on. A manual selection for a node is currently being implemented and should be done once https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1094 is finished. Do we really need to add an extra step for searching by farm name before searching by node ID?

A-Harby:
I know a manual node would be enough in most cases, but it would be a nice option for the user in case he wants to use a specific farm without regard to the node. Maybe it's his farm or a farm in his country, and searching for a specific farm on the mainnet is very time-consuming.

zaelgohary:
The only logical way to implement this is to create a manual farm selection and the user can select nodes from this specific farm. This means I'll have to remove manual node selection. Should I start working on this? @AhmedHanafy725

AhmedHanafy725:
in the automated selection, we can do the following

- allow the user to type the farm name
- disable the farm validation on resources (just make sure that this farm does exist or maybe do a quick search for the farm name to help the user with autocomplete (`name_contains` in the query would help))
- use the farm name entered to search for nodes (at this point if the farm doesn't have nodes that fit the resources, the user has to pick another farm)

zaelgohary:
Investigation and Solution:

Still discussing possible solutions with @AhmedHanafy725.

zaelgohary:
WIP:
Working on creating a request that gets called only on input in the farm field. The request should get all farms whose names match/contain the searched term.

zaelgohary:
Work completed:
Return type from proxy was different than the type from grid client. Hence, it was searching for a property that didn't exist. I edited the property name to match the one in the grid and it should be fixed. PR is ready for review.

zaelgohary:
Work completed:
Added a reload icon inside the farm field to enable the user to reset his search.

zaelgohary:
WIP:

Working on removing the reload button and reverting to automatic farm selection once the search is empty.

zaelgohary:
Work completed: 
- Removed reload button
- Old farm was always chosen if it's in the returned list so I removed this condition
[...]

PR #1461, opened 2023-11-22
Fix search farm

zaelgohary:
### Description

- Remove reload button
- Old farm was always chosen if it's in the returned list so I removed this condition

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1203

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation


packages/playground/src/components/select_farm.vue
```diff
@@ -169,20 +167,11 @@ async function loadFarms() {
   }
   farms.value = farms.value.concat(_farms);
 
-  if (oldFarm) {
-    farm.value = undefined;
-    await nextTick();
-    farm.value = farms.value.find(f => f.name === oldFarm.name);
+  farm.value = farms.value[0];
+  farmInput.value.setStatus(initialized ? ValidatorStatus.Invalid : ValidatorStatus.Init);
```
0oM4R, line 171: It looks like this caused an error.
network: main 
image

mohamedamer453 (commented):
- While searching for the farm name, the old value is kept in the field while it's loading.

   [Screencast from 27-11-23 11:29:39.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/101194226/1cb8a452-3749-4684-befc-5b620046b086)

- If the field was cleared and then the user exits the field it won't make the request and the field stays empty.

   [Screencast from 27-11-23 11:34:52.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/101194226/e94860b1-0950-4b83-8a04-e2d79aee3dd6)

- Sometimes the request isn't made when the field is cleared.

   [Screencast from 27-11-23 11:37:23.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/101194226/7a095d1c-f23a-4e77-8086-52006c7e75e6)

0oM4R:
also i don't think the best UX is to type and wait then make the request, 
on main net i have to wait about 20s before it start loading
also on load more farms it add the same farm 
[Uploading Screencast from 27 نوف, 2023 EET 11:57:39 ص.webm…]()

zaelgohary:
> also i don't think the best UX is to type and wait then make the request, on main net i have to wait for few seconds before it start loading with no data available on the list before start loading also on load more farms it add the same farm
> 
>  Screencast.from.27.2023.EET.11.57.39.webm

Should be fixed.

PR #1443, opened 2023-11-20
use error classes grid client

0oM4R:
### Description

Use error classes in gridclient
> will add docs for types/errors in another PR
### Changes

all errors should handled and replaced with our error types 
added some new errors in types/errors

### Related Issues 

 - #995 
 
### Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation
- [x] Code format and docstrings
- [ ] Screenshots/Video attached (needed for UI changes)

Issue #1398, opened 2023-11-10
2.2.0-rc4 dashboard not able to connect tfchain wallet


coesensbert:
### Description

When trying to connect your wallet no call is done, instead your browser returns `Error: url should be provided`
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/88312671/13968911-8619-4aee-b28d-b8b430b06ea0)

### Steps to reproduce

Navigate to https://dashboard.bknd1.dev.ava.tf/
Login with your mnemonic

### Notes

This is an environment separate from k8s and [setup only with docker-compose](https://github.com/threefoldtech/grid_deployment/blob/development/docker-compose/devnet/docker-compose.yml).

Specifically for the dashboard, these environment variables where used:
```
## Grid Dashboard
  grid_dashboard:
    depends_on:
      grid_proxy:
        condition: service_started
    container_name: grid_dashboard
    image: ghcr.io/threefoldtech/dashboard:2.2.0-rc2
    restart: unless-stopped
    networks:
      tfgrid_bknd:
    environment:
      STELLAR_NETWORK: test
      TFCHAIN_NETWORK: custom
      GRAPHQL_URL: https://graphql.bknd1.dev.ava.tf/graphql
      BRIDGE_TFT_ADDRESS: GDHJP6TF3UXYXTNEZ2P36J5FH7W4BJJQ4AYYAXC66I2Q2AH5B6O6BCFG
      GRIDPROXY_URL: https://gridproxy.bknd1.dev.ava.tf
      SUBSTRATE_URL: wss://tfchain.bknd1.dev.ava.tf
      ACTIVATION_SERVICE_URL: https://activation.bknd1.dev.ava.tf
      PLAYGROUND_URL: https://play.bknd1.dev.ava.tf
      RELAY_DOMAIN: wss://relay.bknd1.dev.ava.tf
      MANUAL_URL: https://manual.grid.tf/
      VERSION: 2.2.0-rc2
```

### Logs/Alerts

```
root@composetst:~# docker logs grid_dashboard

HINT: The default selected STELLAR_NETWORK is test and default selected TFCHAIN_NETWORK is custom.
Environment variables were exported before, if you want to change any of them maybe you have to re-export them or close the terminal window and start from scratch.

window.configs = {
  APP_API_URL: 'wss://tfchain.bknd1.dev.ava.tf',
[...]

PR #1465, opened 2023-11-23
Handle custom network in dashboard

0oM4R:
### Description

grid client uses the passed urls in case of the network is custom.
https://github.com/threefoldtech/tfgrid-sdk-ts/blob/c69cd4c1b2537935a92cf1d2e9d98073cfecb617/packages/grid_client/src/client.ts#L54
Also, the getDefaultUrls function will return undefined in case of custom network.
https://github.com/threefoldtech/tfgrid-sdk-ts/blob/c69cd4c1b2537935a92cf1d2e9d98073cfecb617/packages/grid_client/src/client.ts#L209

### Changes

add an error in grid client if there is any URL is missing and the network is custom 
> this should be changed in #1443
pass the urls in dashboard ffrom config

in playground we handle the custom network in activate account and create new account 
also we add clearError() on validating mnem
### Related Issues

- #1398 

### Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation


packages/dashboard/src/utils/grid.ts
```diff
@@ -8,6 +8,11 @@ export async function getGrid(mnemonic: string) {
     mnemonic,
     network: config.network as NetworkEnv,
     backendStorageType: BackendStorageType.tfkvstore,
+    relayURL: config.relay,
+    graphqlURL: config.graphqlUrl,
+    substrateURL: config.wsUrl,
+    activationURL: config.activationServiceUrl,
+    proxyURL: config.gridproxyUrl,
```
zaelgohary, line 15: We only need to set URLs in case the network is custom, other than that, no need to. This goes for all changes made.

zaelgohary (changes requested):

packages/dashboard/src/utils/grid.ts
0oM4R: done

MohamedElmdary (approved):

zaelgohary (approved):

Issue #1502, opened 2023-11-27
Format DAO proposals


xmonader:
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/64129/1c956fbd-c300-4328-a3d8-134e94dd794b)

A-Harby:
Next to the Yes option, the `No` option appears to be disabled.
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/ba496fb1-fbcc-4f76-a5b9-f5c3b27241c9)

maayarosama:
> Next to the Yes option, the `No` option appears to be disabled.

The No button disabled, Although it has a gray color which may lead the user to think it's disabled. we can change the color to red, what do you think @ehab-hassan? 
![Peek 2023-12-24 14-18](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/66883096/62b2ac57-6c16-4e7b-a949-a4bd9cd744db)

zaelgohary:
@maayarosama There's a [PR](https://github.com/threefoldtech/tfgrid-sdk-ts/pull/1767) where all buttons are changed. I think this one is also covered.

A-Harby:
The only comment that searches is case-sensitive and it shouldn't be.

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/b584bf7d-4c0c-4681-bd6f-be0caa628f62)
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/e84f0b91-5fe9-4d30-b1de-4e75535525f9)

amiraabouhadid:
Added case insensitive search

A-Harby:
Verified,
Devnet 33c308d.

[TC2232 - Verify Proposals Search Functionality](https://app.testlodge.com/a/26076/projects/40893/suites/234374?expand_section=422339#case_4150441)

The search has been updated to be case-insensitive.
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/32994883-5fb7-4595-be1f-51a33e8f00ee)
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/26ae4656-0b37-4d27-b9f6-bea7c8a86e96)

PR #1559, opened 2023-12-03
Formatting dao proposal's card

maayarosama:
### Description

Formatting dao proposals' cards.
Screenshot from 2023-12-03 15-44-43

### Changes

- Made titles bold
- Changed view details link to a tooltip
- Added some styling to the card
### Related Issues

#1502
### Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation

### DIFF
--- a/packages/playground/src/dashboard/dao_view.vue
+++ b/packages/playground/src/dashboard/dao_view.vue
@@ -36,28 +36,32 @@
           </v-card>
 
           <v-card class="my-3 pa-5" v-for="(proposal, i) in filteredProposals(tab.content.value)" :key="i">
-            <v-card-title class="pa-0 mb-5" v-if="proposal.action">
+            <v-card-title class="pa-0 mb-5 font-weight-bold" v-if="proposal.action">
               {{ proposal.action }}
             </v-card-title>
-            <!-- <v-card-subtitle class="pb-0">
-              <p class="font-weight-bold">
-                Proposal hash: <span class="text--secondary">{{ proposal.hash }}</span>
-              </p>
-            </v-card-subtitle> -->
+            <v-divider class="mt-1 mb-5 text-red-700" v-if="proposal.action" />
+
             <v-card-text class="pb-0">
-              <v-row>
-                <p class="font-weight-medium mr-3">Description:</p>
-                <span class="text--secondary">
-                  {{ proposal.description }}
-                  <a v-bind:href="proposal.link" v-bind:target="'blank'">More details</a>
-                </span>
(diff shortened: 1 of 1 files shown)


packages/playground/src/dashboard/dao_view.vue
```diff
@@ -36,28 +36,32 @@
...
-            <v-card-title class="pa-0 mb-5" v-if="proposal.action">
+            <v-card-title class="pa-0 mb-5 font-weight-bold" v-if="proposal.action">
               {{ proposal.action }}
             </v-card-title>
-            <!-- <v-card-subtitle class="pb-0">
-              <p class="font-weight-bold">
-                Proposal hash: <span class="text--secondary">{{ proposal.hash }}</span>
-              </p>
-            </v-card-subtitle> -->
+            <v-divider class="mt-1 mb-5 text-red-700" v-if="proposal.action" />
```
amiraabouhadid, line 42: might be better to just wrap these two components in a div with a single v-if="proposal.action"

amiraabouhadid:
The details button should always be on the right even when there is no remark. Looks odd here, no?
image

zaelgohary (changes requested):
Whenever I click on details, it redirects to the playground in a different tab as seen below.

[details_btn.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/40770501/6bdb3344-616d-4fb7-9670-dd35fd9165c2)

maayarosama:
> Whenever I click on details, it redirects to the playground in a different tab as seen below.
> 
>  details_btn.webm

As all proposals on devnet consists of summy data, the details button won't redirect to anything. However if you test against mainnet you can see that it redirects to the right link.

zaelgohary:
> > Whenever I click on details, it redirects to the playground in a different tab as seen below.
> > details_btn.webm
> 
> As all proposals on devnet consists of summy data, the details button won't redirect to anything. However if you test against mainnet you can see that it redirects to the right link.

Then the button should only be visible in mainnet only, no?
فا

maayarosama:
> > > Whenever I click on details, it redirects to the playground in a different tab as seen below.
> > > details_btn.webm
> > 
> > 
> > As all proposals on devnet consists of summy data, the details button won't redirect to anything. However if you test against mainnet you can see that it redirects to the right link.
> 
> Then the button should only be visible in mainnet only, no?

A per discussion with @AhmedHanafy725 it should be visible in all nets, The user who creates the proposal should always add accurate data.

zaelgohary (approved):

Issue #1526, opened 2023-11-28
Explorer farm filters are broken


zaelgohary:
Explorer farm filters have some issues:

- It only searches with the first character/digit of the search input
- Sometimes it returns no data available while it should not
- The correct data are being returned only if the search is pasted to the field

[explorer_farm_filters.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/40770501/534d34f6-2e59-4076-8300-20756a255c64)

A-Harby:
Verified, 
Devnet, 2d54818.

Farm ID is searching for the correct farm and waiting for one second after finishing writing to start filtering.
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/b557b06c-477b-4b13-a89c-7c3655f06f95)
Same for the farm name, and it also works for non-sensitive cases.
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/24a10c79-d06e-48a7-9a57-08044ca8a94a)

[TC2273 - Filter Farms by ID](https://app.testlodge.com/a/26076/projects/40893/suites/234374?expand_section=422340#case_4150482)
[TC2274 - Filter Farms by Farm Name](https://app.testlodge.com/a/26076/projects/40893/suites/234374?expand_section=422340#case_4150483)

ramezsaeed:
verified on:
- Devenet
- 6d0ec3d

Issue #1524, opened 2023-11-28
enhance explorer farms table


0oM4R:
- is there any need to have free, used, and total ips in the same table? we can go with only two of them, i suggest total and free
- `free` public ip we may change this to something like `available` ips

A-Harby:
Verified.
Devnet, 2d54818.
Used IPs was removed.
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/99f09908-876e-4105-a0a4-69dea0d08a11)

[TC2269 - Farms Table Information](https://app.testlodge.com/a/26076/projects/40893/suites/234374?expand_section=422340#case_4150478)

ramezsaeed:
verified on:
- Devnet
- 13a87cf

PR #1563, opened 2023-12-03
Development 313 fix farms filter & enhance farm table

amiraabouhadid:
### Changes

- Fix of farmId filter
- Use nameContains instead of name= query in farm name filter
- remove used public ips column
- rename free public ips to available public ips

### Related Issues

#1526 #1524

### DIFF
--- a/packages/playground/src/views/farms.vue
+++ b/packages/playground/src/views/farms.vue
@@ -85,7 +85,6 @@ const _getFarms = async (queries: Partial<FarmsQuery>) => {
           return {
             ...farm,
             totalPublicIp: total,
-            usedPublicIp: used,
             freePublicIp: total - used,
           };
         });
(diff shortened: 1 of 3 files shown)


packages/playground/src/views/farms.vue
```diff
@@ -145,15 +143,32 @@ const updateSorting = () => {
...
   if (options) {
     options.page = page.value;
     options.size = size.value;
     options.sortBy = sortBy.value;
   }
+  await updateFarms();
 };
-watch(mixedFarmFilters.value, updateFarms, { deep: true });
 
+watch(mixedFarmFilters.value, updateFarms, { deep: true });
```
Mahmoud-Emad, line 168: We can watch on the reference, not the value, because you used {deep: true} -> which means watch on it deep, please

```suggestion
watch(mixedFarmFilters, updateFarms, { deep: true });
```

packages/playground/src/views/farms.vue
```diff
@@ -145,15 +143,32 @@ const updateSorting = () => {
...
+watch(filterFarmInputs.value, updateQueries, { deep: true });
```
Mahmoud-Emad, line 169: Here, also

Mahmoud-Emad (changes requested):
Great work! I've added a few comments; please take a moment to review and address them if you find them relevant.

Mahmoud-Emad (approved):

zaelgohary (commented):
Everything works fine. Just one comment, filter should be opened by default when the page is loaded.

zaelgohary (approved):

Issue #1179, opened 2023-10-05
Refactor user nodes to use owned_by query


zaelgohary:
requires: https://github.com/threefoldtech/tfgrid-sdk-go/issues/417

A-Harby:
Verified,
Devnet, d9d6f02.

The request that is being called is now using ownedby https://gridproxy.dev.grid.tf/nodes?ret_count=true&page=1&size=10&owned_by=666

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/1b98cd60-538d-427c-8782-a6f6bbff9114)

[TC2212 - List Nodes](https://app.testlodge.com/a/26076/projects/40893/suites/234374?expand_section=422339#case_4150421)

A-Harby:
Devnet, 842ebef.

The request is still working with `owned_by` and with accounts that have nodes.
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/12fe0467-74ab-4782-8485-c74f1745082c)

ramezsaeed:
Verified:
- Devnet
- 842ebef

PR #1572, opened 2023-12-04
add nodesOwnedBy and refactor user_nodes

mohamedamer453:
### Description

Refactor user_nodes to use owned_by query from the gridproxy.

### Changes

- Added nodesOwnedBy to the gridproxy_client.
- Refactored user_nodes to use `nodesOwnedBy` instead of returning farms then returning the nodes that belong to the farms.

### Related Issues

https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1179

### Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation


packages/gridproxy_client/src/modules/nodes.ts
```diff
@@ -116,6 +116,11 @@ export class NodesClient extends AbstractClient<NodesBuilder, NodesQuery> {
     return res.json();
   }
 
+  public async nodesOwnedBy(twinId: number): Promise<Node[]> {
+    const res = await this.builder({}).build(`/nodes?owned_by=${twinId}`);
```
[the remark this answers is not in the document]
amiraabouhadid: Ok fair enough. It is handled in the component so looks good. Thanks

amiraabouhadid (approved):

packages/gridproxy_client/src/modules/nodes.ts
```diff
@@ -116,6 +116,11 @@ export class NodesClient extends AbstractClient<NodesBuilder, NodesQuery> {
     return res.json();
   }
 
+  public async nodesOwnedBy(twinId: number): Promise<Node[]> {
```
AhmedHanafy725, line 119: why do we create a new method for it when we can update the `list` method with one extra query?

mohamedamer453: `ownedBy` was added as an extra query parameter to `list` and the `nodesOwnedBy` function was removed in https://github.com/threefoldtech/tfgrid-sdk-ts/pull/1572/commits/79743fa92f4b25a191aae31badddf6f5e7ef052c

Issue #1570, opened 2023-12-04
Node availability check in the farms page is done twice


mohamedamer453:
### Description

The node availability check that checks the uptime of the node when the user enters the farms page is done twice. and with the second time if the node details is expanded it will close it.

[Screencast from 04-12-23 13:01:31.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/101194226/3b284073-c5dd-4578-b6e9-f8b55c582ae6)

### Steps to reproduce

- Log in to your account.
- Navigate to the dashboard from the side bar.
- Navigate to farms.
- Expand the details of any node.
- Check the console.

A-Harby:
Verified,

Node requests are only being called once.
https://gridproxy.dev.grid.tf/nodes?ret_count=true&page=1&size=10&owned_by=162
https://gridproxy.dev.grid.tf/nodes/72
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/d6842e00-9cb0-4632-8df0-1355a08519f7)

Issue #1569, opened 2023-12-04
(Mainnet Verification) Incorrect node uptime value in the `Your Nodes` table


mohamedamer453:
### Description

The current value of node uptime is incorrect in the new playground.

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/101194226/14fbb31b-bc96-413b-96e7-a9b0ea2c2577)

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/101194226/34b98316-e22f-4de9-8324-d89a9d6feab3)

This goes for all nodes either up or down as seen in the images above.

### Steps to reproduce

- Login to your account.
- Navigate to Dashboard then to Farms.
- In the `Your Nodes` table, check the uptime of any node.

### Environment 
- Network: `Devnet`

0oM4R:
investigation:
this is not calculation issue, this happens only after reloading the user nodes after a time out 
https://github.com/threefoldtech/tfgrid-sdk-ts/blob/8df79cc241b75160649caeade8e46cc2100fe3fc/packages/playground/src/dashboard/components/user_nodes.vue#L212-L214

not sure why this is happening yet

A-Harby:
Can you help on how this value `3568977` converted to `23.56%`?
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/da22383f-127f-4747-8817-526884c1c89d)
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/9632536d-b2c8-483f-abc6-b114e184e218)

A-Harby:
I was able to get the calculation but couldn't find any documents or explanations on how it was calculated.
  
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/bbd535c7-3771-4f5e-8cd9-4c65106f069a)

ramezsaeed:
@0oM4R 
please check Harby comment

0oM4R:
https://github.com/threefoldtecharchive/tfgrid_dashboard/issues/366#issuecomment-1304436906

@A-Harby  this should explain

A-Harby:
For the documents, I create an issue [here]: https://github.com/threefoldtech/info_grid/issues/354.

A-Harby:
For the documents, I created an issue [here]: https://github.com/threefoldtech/info_grid/issues/354.

A-Harby:
Verified on Devnet,

[...]

PR #1582, opened 2023-12-05
Fix uptime in dashboard nodes 

0oM4R:
### Description

- The anonymous function got called on mount automatically, which makes the nodes table reload after 10 seconds of mounting.
- Uptime was calculated on mount only, and if reload nodes were triggered, the uptime would not be valid.

### Related Issues

- #1570 
- #1569 

### Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation

### DIFF
--- a/packages/playground/src/dashboard/components/user_nodes.vue
+++ b/packages/playground/src/dashboard/components/user_nodes.vue
@@ -126,12 +126,7 @@ import { useProfileManager } from "@/stores";
 import type { NodeDetailsCard } from "@/types";
 import { createCustomToast, ToastType } from "@/utils/custom_toast";
 import { getNodeStatusColor } from "@/utils/get_nodes";
-import {
-  getNodeAvailability,
-  getNodeMintingFixupReceipts,
-  getNodeUptimePercentage,
-  type NodeInterface,
-} from "@/utils/node";
+import { calculateUptime, getNodeAvailability, getNodeMintingFixupReceipts, type NodeInterface } from "@/utils/node";
 
 import NodeMintingDetails from "./NodeMintingDetails.vue";
 import PublicConfig from "./public_config.vue";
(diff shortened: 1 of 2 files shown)


mohamedamer453 (commented):
- The nodes table still loads twice, once when the page is loaded and another after 10 seconds.

   [Screencast from 05-12-23 14:42:47.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/101194226/6bffc5d2-5a7f-4b53-8e77-36effb73eed3)

0oM4R:
[Screencast from 05 ديس, 2023 EET 03:56:33 م.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/62248851/827f46c8-1f50-44db-a8bc-ca01a21366b8)
works fine with me

AlaaElattar:
> Screencast.from.05.2023.EET.03.56.33.webm 
> works fine with me

Same, It works really fine with me.

packages/playground/src/dashboard/components/user_nodes.vue
```diff
@@ -197,21 +192,13 @@ export default {
...
-      if (nodes) {
-        await Promise.all(
-          nodes.map(async (n, _) => {
-            n.uptime = +(await getNodeUptimePercentage(n.nodeId));
-            return n.uptime;
-          }),
-        );
-      }
-    });
+    onMounted(async () => await getUserNodes());
```
Mahmoud-Emad, line 195: You can pass the function directly to the onMounted hock instead of calling it inside an arrow function.

```suggestion
    onMounted(getUserNodes);
```

Mahmoud-Emad (changes requested):
Great work!

mohamedamer453:
> * The nodes table still loads twice, once when the page is loaded and another after 10 seconds.

It appears that i didn't have the latest changes, tried it again now and it works fine.

mohamedamer453 (approved):

packages/playground/src/dashboard/components/user_nodes.vue
0oM4R: thanks

mohamedamer453 (approved):
lgtm, great work!

Mahmoud-Emad (approved):
LGTM

Issue #1568, opened 2023-12-04
Farm name maximum character limit


mohamedamer453:
### Description

The current max length of farm name in the new playground in 15 character while it's 40 characters from the chain and the old dashboard also had a max of 40 characters.

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/101194226/7df4210c-3458-48ef-974c-0afe615c510e)

also here are some of the farms that were created from the old dashboard that exceeded the 15 character limit and they were created without any issue.

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/101194226/64e08744-73b2-4db1-a373-220142ef2664)

and from the chain here's a farm that has a 40 char name that was created successfully.

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/101194226/4ede4494-4ec0-45a7-95cd-25f5d2d9fb0e)
 
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/101194226/99323756-8384-49ba-bf5d-4b148e31f019)

### Steps to reproduce

- Login to your account.
- Navigate to Dashboard then to Farms.
- Click on the `Create Farm` button.
- Enter a name that has more than 15 char

A-Harby:
Verified,
Devnet, 28d6c07.

The maximum char limit is upgraded to 40 char now.
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/fe55cf7a-bb3b-4c35-b773-708caeb6ed51)

[TC2241 - Farms Input validation](https://app.testlodge.com/a/26076/projects/40893/suites/234374?expand_section=422339#case_4150450)
[TC2205 - Create a Farm](https://app.testlodge.com/a/26076/projects/40893/suites/234374?expand_section=422339#case_4150414)

PR #1608, opened 2023-12-07
Update max farm name to be 40 chars

AlaaElattar:
### Description

The current max length of farm name in the new playground in 15 character while it's 40 characters from the chain and the old dashboard also had a max of 40 characters.
### Changes

Update validation length to be 40 not 15 chars.
### Related Issues

List of related issues
https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1568

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/packages/playground/src/dashboard/components/create_farm.vue
+++ b/packages/playground/src/dashboard/components/create_farm.vue
@@ -20,7 +20,7 @@
                   validators.required('Farm name is required.'),
                   name => validators.isAlpha('Farm name must start with an alphabet char.')(name[0]),
                   validators.minLength('Farm name minimum length is 2 chars.', 2),
-                  validators.maxLength('Farm name maximum length is 15 chars.', 15),
+                  validators.maxLength('Farm name maximum length is 40 chars.', 40),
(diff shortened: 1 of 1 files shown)


Mahmoud-Emad (approved):

0oM4R:
Did you check the client validation to see if it needed to be updated too?

AlaaElattar:
> Did you check the client validation to see if it needed to be updated too?

Fixed 
image

packages/grid_client/src/modules/models.ts
```diff
@@ -697,7 +697,7 @@ class AddStellarAddressToFarmModel {
 }
 
 class CreateFarmModel {
-  @Expose() @IsString() @IsNotEmpty() @MaxLength(NameLength) name: string;
+  @Expose() @IsString() @IsNotEmpty() @MaxLength(40) name: string;
```
0oM4R, line 700: can we define a const and use it like NameLength ? IMO it will be better

AhmedHanafy725 (approved):

Issue #1586, opened 2023-12-05
Second nav bar appears after log out 


0oM4R:
![Screenshot from 2023-12-05 14-35-52](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/62248851/83ea5a7d-777b-4380-821a-2a868d611a6a)

AlaaElattar:
### Work Completed:
- Added condition that second navbar won't appear until user logged in.

A-Harby:
Verified,
Devnet 0b2f0d3.

The second nav bar doesn't show when the user logs out.
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/7a5fa505-22b5-4faf-b52e-0b5e92feae80)

ramezsaeed:
Verified:
- Devnet
- 13a87cf

![Screenshot from 2023-12-19 11-33-37](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/16083592/aa81cbd2-5ff1-4f70-b694-4b33d1fa5cff)

PR #1623, opened 2023-12-10
Add Back Button in second navbar

AlaaElattar:
### Description

- If used logged out, the navbar would still be available
### Changes

Add Condition no navbar, won't appear unless user logged in.
### Related Issues

https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1586

[Screencast from 12-10-2023 11:14:52 AM.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/54944491/0621ad81-91f7-4290-b889-b8e04a186f34)

### Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation

### DIFF
--- a/packages/playground/src/App.vue
+++ b/packages/playground/src/App.vue
@@ -143,7 +143,13 @@
           height="50"
         >
           <v-container>
-            <v-row>
+            <v-row class="justify-end">
+              <v-btn
+                icon="mdi-arrow-left"
+                v-if="navbarConfig.back"
+                class="align-self-center mr-auto"
+                @click="navigateToPrevRoute(navbarConfig.path)"
+              ></v-btn>
               <v-breadcrumbs :items="navbarConfig.path" active-color="secondary">
                 <template v-slot:divider>
                   <v-icon icon="mdi-chevron-right"></v-icon>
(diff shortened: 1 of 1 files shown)


MohamedElmdary:
Please reverse back btn position with the path in second navbar

AlaaElattar:
> Please reverse back btn position with the path in second navbar

I don't understand your comment. Would you please explain more ?

AlaaElattar:
- Positions swapped.
image

ehab-hassan:
Why do we have 2 things with the same function? also putting Breadcrumb at the right is wired to the user, I think one is enough and to be at the left, also Breadcrumbs is better than the back button

AlaaElattar:
> Why do we have 2 things with the same function? also putting Breadcrumb at the right is wired to the user, I think one is enough and to be at the left, also Breadcrumbs is better than the back button

Qa returned it in progess and asked for back button to be added

xmonader:
For me this 
image

looks very clear as a breadcrumb (it's expected each item in that path to be clickable to get me back to where I was -except for the current page of course-

Similar designs 

image

image

image

and some use `/` instead of `>`

image

All in all, it seems to be implicit how to use breadcrumbs UX wise, even no need to underline the trail

xmonader:
I'll close this one and the other two issues (@AlaaElattar thank you so much!)

AlaaElattar:
- Button removed and breadcrumb at left

A-Harby:
I didn't say to move it to the right side what I meant is that it is kind useless to have a breadcrumb for only one page navigation and back button on the right side would be enough.

0oM4R (approved):

maayarosama (approved):

xmonader:
> I didn't say to move it to the right side what I meant is that it is kind useless to have a breadcrumb for only one page navigation and back button on the right side would be enough.

I understand now what you meant, similar to this 
image

I added this comment for documentation purposes

Issue #1600, opened 2023-12-06
The title of the info section for the solutions is clipped


mohamedamer453:
### Description

The title of the info section for the solutions appears to be clipped by another component. this applies for all solutions.

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/101194226/8d9e5a80-1c44-444c-8051-c6ef9959b196)

### Steps to reproduce

- Navigate to the dashboard and login to your account.
- From the sidebar select solutions then select any solution.
- On the top right corner click on the `i` icon.
- Check the title.

ehab-hassan:
fixed at development_tooltip__color #1626

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/10484295/d0fd2b64-a239-4cfa-94e8-550f27c5c2e1)

A-Harby:
Verified,
Devnet 0b2f0d3.

The title in all the solutions isn't clipped anymore.
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/d77f4c93-064f-46a3-b171-e92e8ca95e15)
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/1366d484-4e10-435d-b564-953be3806138)
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/b9e2630f-f1ac-48b9-81b1-56873c3d4ed6)

ramezsaeed:
verified on:
- Devnet
- 13a87cf

Issue #1591, opened 2023-12-06
Change close button UI in connect your wallet


A-Harby:
## Description

New playground,

The close button in the Connect wallet really does look like it is disabled in dark mode.
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/825e49a5-35ef-4e33-a3b7-93ac6b72e814)

ehab-hassan:
Fixed at development_tooltip__color https://github.com/threefoldtech/tfgrid-sdk-ts/pull/1626
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/10484295/b1714113-55e6-4b6f-ba76-19e7aefa3c23)

A-Harby:
2d54818 
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/46a9118a-50d6-40ce-b9f8-68d9d035d846)

ehab-hassan:
it seems my code overwritten :)
https://github.com/threefoldtech/tfgrid-sdk-ts/commit/6819bfbc2c9e70e12eb4030fde22f28693fa7bf1 
created new PR
https://github.com/threefoldtech/tfgrid-sdk-ts/pull/1634

A-Harby:
Verified, 
Devnet 911c562.

Close button UI improved.
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/225e6155-3d18-40b3-8d99-5973ce20fd53)
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/ac2e5bc4-f15d-401e-8e58-edd900bd96ef)
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/ab455aab-058b-48ad-989e-3a1b4a7c138a)
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/49111755-8d84-4c59-8827-a6f0cad1d242)

ramezsaeed:
Verified:
- Devnet
- 13a87cf

![Screenshot from 2023-12-19 14-07-42](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/16083592/db3f99a1-a70f-4d46-8492-a138bb9ec3f1)

Issue #1478, opened 2023-11-26
Dedicated nodes: difficult to read the hint on the price


xmonader:
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/64129/ec45d6d3-a559-4924-903d-a5a37825053d)

ehab-hassan:
Fixed tooltip at development__313_farm_enhancement branch ll create PR soon
![Untitled-1](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/10484295/10a8c8df-3c27-46aa-ac64-13987af617a4)

zaelgohary:
Tooltip is not fixed yet.

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/40770501/6b4c5edd-5c94-4fb1-9056-5faa6789b124)

ehab-hassan:
it seems my code was overridden :)
https://github.com/threefoldtech/tfgrid-sdk-ts/commit/6819bfbc2c9e70e12eb4030fde22f28693fa7bf1 
created new PR
https://github.com/threefoldtech/tfgrid-sdk-ts/pull/1634

A-Harby:
Verified,
- Devnet
- 13a87cf
- [TC2219 - List Dedicated Nodes](https://app.testlodge.com/a/26076/projects/40893/suites/234374?expand_section=422339#case_4150428)

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/1fe99ec9-3135-4888-ba69-21136bc6c2d9)
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/9447d2ff-59aa-4b0d-8bc6-ac59b038784c)

ramezsaeed:
Verified:
- Devnet
- 2.3.0-alpha3

PR #1634, opened 2023-12-10
tooltip color2

ehab-hassan:
fixed The title of the info section for the solutions is clipped #1600

image

fixed all hint (tooltip) Dedicated nodes: difficult to read the hint on the price #1478
Untitled-1

made it more whiter Change the close button UI in Connect your wallet #1591
image


zaelgohary:
Will close as the issue is fixed.

Issue #1668, opened 2023-12-12
Weird Behavior in `Add a public config` in my farms page  


AlaaElattar:
### Description

 There's weird behavior as its seen below in the validations.
Why when i type in IPv4 gateway validation is working ?

[Screencast from 12-12-2023 12:05:13 PM.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/54944491/4ae9ec23-4758-4bac-84f4-af4b8c73d5fe)

zaelgohary:
All validation is triggered if one input is being validated.

AlaaElattar:
- It's not actually an issue, it's implemented like this for a reason.
- ipv4 & gw4 are a must in form. I can't submit any input unless they're full
- if ipv4 triggered -> gw4 should be too and vice versa
- if ipv6 triggered -> gw6 should be too and vice versa
- domain optiona
- This explains the behavior of the form.

0oM4R:
should we trigger the validation if the other value is valid only ?

> * It's not actually an issue, it's implemented like this for a reason.
> * ipv4 & gw4 are a must in form. I can't submit any input unless they're full
> * if ipv4 triggered -> gw4 should be too and vice versa
> * if ipv6 triggered -> gw6 should be too and vice versa
> * domain optiona
> * This explains the behavior of the form.

AlaaElattar:
### Work In Progess (WIP):
- Working in that comment https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1668#issuecomment-1860253027

AlaaElattar:
### Work Completed :
- About ipv4 && gw4 as both are must i think they should be both triggered once user types at any one.
- same for ipv6 and gw6, excepty they're not a must untile they come together.
- As long as the button is not enabled until all are valid, i don't see this will make a difference https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1668#issuecomment-1860253027

A-Harby:
One case is not fixed yet: when any other input is being typed and validated and there is already a wrong validation on the other inputs, all of them will be affected and revalidated.

[Threefold Dashboard: Your Farms.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/f3e0a37e-2106-4cf1-ab0e-e1c32bcd0668)

[...]

Issue #1618, opened 2023-12-07
Can't add public configuration to my node


A-Harby:
### Description

Devnet, 2d54818.

I can't add any public configuration to my node and keep getting this error.

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/86b3436c-4ca4-46a7-a932-f712bb0f2fb1)

And also the error message must be more clearer to the user!

### Steps to reproduce

1. Connect your wallet (your account must have a node).
2. Navigate to the farms by expanding the dashboard from the sidebar.
3. From your node table, click on Add public IP from within your node.
4. Fill in the input and press save.

ramezsaeed:
Ipv4 should not be the same as the gateway

coesensbert:
and should be in the subnet mask range because of on-chain validation

AlaaElattar:
### Work Completed 
- Added condition to check they're different before sending the request.

AlaaElattar:
### Work Complete: 
- The error msg was triggered at wrong cases and that is fixed now.

A-Harby:
Verified,
Devnet 33c308d.

[TC2218 - Public Config Input Validation.](https://app.testlodge.com/a/26076/projects/40893/suites/234374?expand_section=422339#case_4150427)

The same IP and gateway validation were added.
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/ca991f29-2e24-49ee-b6d6-18af01a79bdd)
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/f9fb5cd3-c972-4530-8e08-c2543a4c574d)

After fixing the wrong input, IP was added.
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/27b83e3e-2d83-4fd9-8bc2-6e83cd09493d)

PR #1669, opened 2023-12-12
Check that ip should be different from gateway

AlaaElattar:
### Description

- Complex error when user add ip same as gateway. 
- Wierd validation also happened.
- IPv6 can be added to public config of node without Gateway IPv6

### Changes

- Check they're not the same before sending the request. 
- Fix behaviour of validation.
- Check that both IPv6 and Gateway IPv6 should be added together.
 
### Related Issues

https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1618
https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1668
https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1732

[Screencast from 12-19-2023 01:16:27 PM.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/54944491/a4f044d3-8900-454a-ad0d-1d8db55df786)

### Checklist

- [x] Tests included
- [ ] Build pass
- [ ] Documentation


zaelgohary (commented):
I think validation should happen before clicking the button and making the request. If 2 IPs are the same, the button should be disabled.

AlaaElattar:
> I think validation should happen before clicking the button and making the request. If 2 IPs are the same, the button should be disabled.

image

mohamedamer453 (commented):
- The validation isn't very accurate in some cases like the one seen below the validation is triggered with incorrect cases and in the case that should trigger it, it doesn't get triggered.

   [Screencast from 13-12-23 13:50:21.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/101194226/74a8ef11-5b23-4591-8206-57cab36c0f0a)

AlaaElattar:
> * The validation isn't very accurate in some cases like the one seen below the validation is triggered with incorrect cases and in the case that should trigger it, it doesn't get triggered.
>   
>     
>       
>       
>   
>       Screencast.from.13-12-23.13.50.21.webm

I think that's because this issue https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1668

0oM4R:
please check this comment 
https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1668#issuecomment-1860253027

amiraabouhadid (approved):

packages/playground/src/utils/validators.ts
```diff
@@ -705,6 +705,17 @@ export function isString(msg: string) {
   };
 }
 
+export function ipNotEqualGateway(ip: string, gw: string, msg: string) {
+  return (value: string) => {
+    const ipv = ip.split("/");
+    const ipvAddress = ipv[0];
+
+    if (gw === ipvAddress) {
```
zaelgohary, line 713:
```suggestion
  return () => {
    const address = (ip.split("/"))[0];

    if (gw === address) {
```

zaelgohary (commented):
All looks good! Just a small enhancement.

packages/playground/src/utils/validators.ts
AlaaElattar: should be fixed

zaelgohary (approved):

Issue #1604, opened 2023-12-07
Reset nodes table count and pages


A-Harby:
### Description

Devnet, https://github.com/threefoldtech/tfgrid-sdk-ts/commit/2d54818c4c44cb4111462b8cd3daea60c60c602a.

Filtering, applying toggles, or selecting the node status doesn't reset table counts or disable page navigation on it.

### Steps to reproduce

1. Connect your wallet.
2. Navigate to the Nodes page from the sidebar.
3. Click on any node and scroll down to the Twin Details section.

### Logs/Alerts

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/43380107-dbaa-4ae0-a510-a555e403b1f9)

[Threefold Dashboard _ Nodes.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/bdc2f0f7-e4b0-4d48-b612-ecd919bea6bc)

mohamedamer453:
Update

Work in Progress(WIP):

Still under investigation, trying to find out how to update the `item-length` property when the status change and there're no nodes available since this is the scenario where it displays the number from the previous table while there are no nodes available.

A-Harby:
Not all the cases have been reset.

[Threefold Dashboard _ Nodes (1).webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/995e5058-d5a7-41a6-9a0d-23f18044f69b)

amiraabouhadid:
Investigative Work: 
Looked for the reset triggers in nodes.vue

A-Harby:
Waiting for https://github.com/threefoldtech/tfgrid-sdk-ts/issues/770

zaelgohary:
@A-Harby Should be working fine now after https://github.com/threefoldtech/tfgrid-sdk-ts/pull/1819 is merged.

khaledyoussef24:
node count fixed to change to zero when there is no nodes:
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/293a124c-99c9-4faf-ab13-0d19df015836)
checked for down, standby and up.
![Screenshot from 2024-01-02 12-58-15](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/4e60b071-fad5-438f-a100-436156bebdf0)
[...]

PR #1682, opened 2023-12-13
set nodes count to 0 if there're no nodes to display

mohamedamer453:
### Description

Fix count  if there're no nodes to display in the Nodes table.

[Screencast from 13-12-23 11:25:18.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/101194226/15be94c2-3002-4a12-bbec-14acf005a9d1)

### Changes

- Added a condition to set the nodesCount to 0 if the count is equal to 0.

### Related Issues

https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1604

### Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation

### DIFF
--- a/packages/playground/src/views/nodes.vue
+++ b/packages/playground/src/views/nodes.vue
@@ -160,6 +160,9 @@ export default {
           if (count) {
             nodesCount.value = count;
           }
+          if (count == 0) {
+            nodesCount.value = 0;
+          }
         } catch (err) {
           console.log(err);
         } finally {


packages/playground/src/views/nodes.vue
```diff
@@ -160,6 +160,9 @@ export default {
           if (count) {
             nodesCount.value = count;
           }
+          if (count == 0) {
+            nodesCount.value = 0;
+          }
```
0oM4R, line 165: I'm worry about `NaN` issue if the count is undefined maybe 
so we can use `??`
```suggestion
            nodesCount.value = count ?? 0
```

zaelgohary (approved):

amiraabouhadid (approved):

Issue #1651, opened 2023-12-11
(Mainnet) Display Default Information for Unresponsive Nodes in New Dashboard


Mahmoud-Emad:
### Description

**Current Behavior:**
The new dashboard shows an error message with buttons to re-request the node when it doesn't respond, as seen in the provided screenshot:

![Error Message](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/57001890/a2f885c0-fef9-46f6-89c2-75d0203ceb9f)

**Desired Behavior:**
The old dashboard displayed the node details with default values for the stats and hid the node deployments/workloads and GPU cards when a node didn't respond, as shown in the provided screenshot:

![Default Node Information](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/57001890/9878bebb-6404-4e87-ba50-c0c032fca998)

Mahmoud-Emad:
It's possible that this issue arises because we currently retrieve the load statistics directly from the node in real-time through a request. If the node is unavailable, the request fails, resulting in a display of "NA." To address this, I propose that we continue with the current request approach, but in cases where the node doesn't respond, we should fallback to loading the statistics from the grid proxy instead of displaying "NA." This ensures a more seamless and reliable display of information even when individual nodes are temporarily unreachable.

Mahmoud-Emad:
### Issue Update 

Enhanced the node details page by dynamically fetching node information from the grid proxy, eliminating error displays. Additionally, introduced a warning alert to notify users when the node seems operational but is physically offline—possibly due to entering an offline mode. This enhancement aims to provide a more seamless user experience and informative feedback on the node's status.

PS: The changes related to this enhancement are included in [this pull request](https://github.com/threefoldtech/tfgrid-sdk-ts/pull/1701).

Note: Unable to reproduce the errors related to loading cards; testing is required on another node containing a GPU card that appears as 'online' but is physically 'down'.

maayarosama:
1. Work Completed:
- Removing N/A in node resources when the node is unreachable and added the resources returned from grid proxy

2. Work in Progress (WIP):
- Adding an error message and a refresh button in case of failure of retrieving gpu when node is unreachable
3. Investigation and Solution:
- when node is unreachable the offline chip and warning disappears on opening and closing node details dialogue multiple times

maayarosama:
Work Completed:
- Added an error message and a refresh button in case of failure of retrieving gpu when node is unreachable

[...]

PR #1701, opened 2023-12-14
Enhance loading node details, display the actual status, node stats as NA

Mahmoud-Emad:
…
### Description

Improved the node details page by dynamically fetching node information from the grid proxy, eliminating error displays. Additionally, introduced a warning alert to notify users when the node seems operational but is physically offline—possibly due to entering an offline mode. This enhancement aims to provide a more seamless user experience and informative feedback on the node's status.

- Modified the node details page to fetch node information from the grid proxy.
- Added a warning alert to inform users when the node appears to be up but is physically down, suggesting it may be in an offline mode.
- Improved error handling for a more user-friendly experience.

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1651

### Screenshots

- image

### Checklist

- [ ] Tests included
- [x] Build pass
- [x] Documentation

### DIFF
--- a/packages/playground/src/components/node_details.vue
+++ b/packages/playground/src/components/node_details.vue
@@ -139,6 +139,7 @@ export default {
       isError.value = false;
       isLiveStats.value = false;
       errorLoadingStatsMessage.value = undefined;
+      nodeOptions.loadStats = true;
 
       if (props.nodeId > 0) {
         try {
@@ -155,11 +156,7 @@ export default {
           errorLoadingStatsMessage.value =
             "The node appears like it's up but it is physically down maybe because it's gone to offline mode.";
           nodeOptions.loadStats = false;
-
-          if (props.filterOptions.gpu && !nodeOptions.loadGpu) {
-            nodeOptions.loadGpu = true;
-          }
-
+          nodeOptions.loadGpu = false;
           try {
             const _node: GridNode = await getNode(props.nodeId, nodeOptions);
             node.value = _node;
--- a/packages/playground/src/components/node_details.vue
+++ b/packages/playground/src/components/node_details.vue
@@ -36,7 +36,7 @@
 
     <template v-else>
       <v-card>
-        <node-resources-charts :node="node" />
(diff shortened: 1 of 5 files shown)


AlaaElattar:
if node is down the resources is shown as NAN, what if the problem in my internet connection ?? I'm confused about the difference.

mohamedamer453 (commented):
- After opening a node that's up but it's actually down and checking the message then closing the node and opening it again the message disappears.

   [Screencast from 14-12-23 17:26:48.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/101194226/fd446480-29ad-4b0a-8073-a6eb1415773e)


- After opening a node that has the stats set to `N/A` then opening other online nodes you will find that their stats is also set to `N/A`.

   [Screencast from 14-12-23 17:29:31.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/101194226/d158d399-4d54-47ba-ab3a-46d39ace839e)

maayarosama:
> if node is down the resources it's shown as NAN, what if the problem in my internet connection ?? I'm confused about the difference.

Updated Resources to not be shown as NaN, but value returned from grid proxy. 
The difference between them is in this case the node is retrieved from the grid proxy however it was unreachable by the rmb so we display the offline chip and add a warning as previously shown. However when there's a problem with the  internet connection, no node will be retrieved and this error message will be shown to the user
Screenshot from 2023-12-21 17-20-16

AlaaElattar:
- When i opened online node, i got N/A for HDD. why this happened if the node is online and the rest of resources loaded ?

image

maayarosama:
> * When i opened online node, i got N/A for HDD. why this happened if the node is online and the rest of resources loaded ?
> 
> image

In this specific scenario, The calculations done to get the percentage equals NaN as there was a division by 0. so we handle the NaN value by displaying N/A. However this's an exception as most of all other scenarios display a specific percentage 
Screenshot from 2023-12-26 13-34-21
Screenshot from 2023-12-26 13-34-40

AlaaElattar (approved):

0oM4R (approved):
image

Issue #1599, opened 2023-12-06
`Failed to get the user farms` when changing number of items per table on the farm page


A-Harby:
### Description

I got this error while changing items per table on the farm page. It occurred randomly, and the error message was unclear.
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/2a0baeb5-13cf-4bc1-b19a-13c70e44ad5d)

AlaaElattar:
###Investigation and Solution:
- I got this error in a different way, It came from `getUserNodes`, which is really wierd. If the error from there, it should be `Failed to get user nodes!`. 
- still investigating in it.

![Screenshot from 2023-12-12 10-25-55](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/54944491/e549d538-ec5e-4c02-9004-d4f5e79fe1bf)

AlaaElattar:
### Investigation and Solution:
- It occured here because of bad internet connection and there were no farms at the table.
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/54944491/f8446882-ff3f-41f4-89d4-82698175b9cc)

AlaaElattar:
> ### Description
[...]

PR #1703, opened 2023-12-14
update msg to be more friendly in farms page

AlaaElattar:
### Description
- Update error msg when there's problem getting user farms.

### Changes

Error msg updated to `Failed to get farms. Please check your connection.`
### Related Issues

https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1599
### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation


MohamedElmdary:
I think we should use `normalizeError` method and that string should be the fallback error
because now if any error happen user won't know @AlaaElattar

AlaaElattar:
> I think we should use `normalizeError` method and that string should be the fallback error because now if any error happen user won't know @AlaaElattar

yes, u're right.

AlaaElattar:
> I think we should use `normalizeError` method and that string should be the fallback error because now if any error happen user won't know @AlaaElattar

should be fixed in [29e54d8](https://github.com/threefoldtech/tfgrid-sdk-ts/pull/1703/commits/29e54d86cac98953e826b725853fd8d06b55f83e)

amiraabouhadid (changes requested):
I got this error instead 
Screenshot from 2023-12-17 14-06-49
Screenshot from 2023-12-17 14-06-09

AlaaElattar:
> I got this error msg instead Screenshot from 2023-12-17 14-06-49 Screenshot from 2023-12-17 14-06-09

Fixed. will always show friendly error msg.

zaelgohary (commented):
Are we certain that a connection error is the only potential issue? What if there are problems with the Grid Proxy itself? Shouldn't the error message be more inclusive or at least address other possible scenarios beyond network issues?

AlaaElattar:
> Are we certain that a connection error is the only potential issue? What if there are problems with the Grid Proxy itself? Shouldn't the error message be more inclusive or at least address other possible scenarios beyond network issues?


- Here i tried if i can't reach proxy and i got same logs. Do u have an idea how to differentiate the connection error from any other error ? 

image

0oM4R:
> > Are we certain that a connection error is the only potential issue? What if there are problems with the Grid Proxy itself? Shouldn't the error message be more inclusive or at least address other possible scenarios beyond network issues?
> 
> * Here i tried if i can't reach proxy and i got same logs. Do u have an idea how to differentiate the connection error from any other error ?
> 
> image

@AlaaElattar 
you can check on the error type, all errors should be an instance of a class on types package

example:

- import the type you want to check on here i will import validationError

then check on the catch error
` if (error instanceof ValidationError)`

amiraabouhadid:
found this issue while testing #1733

packages/playground/src/dashboard/components/user_farms.vue
```diff
@@ -235,7 +236,11 @@ export default {
         farmsCount.value = count || filteredFarms.length;
       } catch (error) {
         console.log(error);
-        createCustomToast("Failed to get user farms!", ToastType.danger);
+        if (error instanceof ConnectionError) {
+          createCustomToast("Failed to get farms. Please check your connection.", ToastType.danger);
+        } else {
+          createCustomToast("Failed to get farms.", ToastType.danger);
```
amiraabouhadid, line 242: i got this . can you please add "Please check your connection. " here too?
image

AlaaElattar: If i added same msg , what is the point then of checking on error type ?

MohamedElmdary: This issue should be fixed here https://github.com/threefoldtech/tfgrid-sdk-ts/pull/1735 @AlaaElattar @amiraabouhadid

zaelgohary: @amiraabouhadid Another error type not connection related might happen in the else, so I think it's better this way.

zaelgohary (dismissed):

packages/playground/src/dashboard/components/user_farms.vue
zaelgohary: I checked #1735 and it indeed fixes the issue in case no connection is needed.

AlaaElattar:
- I removed the connection check on error as when this pr https://github.com/threefoldtech/tfgrid-sdk-ts/pull/1735 merged, this will happen with any case of internet connection.

PR #1705, opened 2023-12-16
feat: add global navigation loader

MohamedElmdary:
### Description
add top global navigation loader to indicate that a page is being loading

### Changes
- Add navigation loader
- Use it in App.vue
- Add hook to simplify getting navigation status `<Loading | Failed | Success>`

https://github.com/threefoldtech/tfgrid-sdk-ts/assets/31689104/3ed5d2f9-86be-4f1d-8360-09a7d713df3f

### Issues
- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1708

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/packages/playground/src/hooks/useNavigationStatus.ts
+++ b/packages/playground/src/hooks/useNavigationStatus.ts
@@ -1,0 +1,39 @@
+import noop from "lodash/fp/noop.js";
+import { onMounted, onUnmounted, ref } from "vue";
+import { useRouter } from "vue-router";
+
+export function useNavigationStatus() {
+  const router = useRouter();
+  const status = ref<"Loading" | "Success" | "Failed">("Success");
+
+  let unSubscribeBeforeEach = noop;
+  let unSubscribeAfterEach = noop;
+
+  let currentToPath: string;
+
+  onMounted(async () => {
+    await router.isReady();
+
+    unSubscribeBeforeEach = router.beforeEach((to, _, next) => {
+      currentToPath = to.fullPath;
+      status.value = "Loading";
+
+      return next();
+    });
+
+    unSubscribeAfterEach = router.afterEach((to, _, failure) => {
+      if (status.value !== "Loading" || currentToPath !== to.fullPath) {
+        return;
+      }
+
+      status.value = failure ? "Failed" : "Success";
+    });
+  });
+
+  onUnmounted(() => {
+    unSubscribeBeforeEach();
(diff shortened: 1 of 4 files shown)


AlaaElattar:
- Please attach Screenshots/Video for the changes.

0oM4R (approved):

0oM4R:
What should happen if navigation fails?

MohamedElmdary:
> What should happen if navigation fails?

Well, this should never happen in our case
but if failed the line will return to 0 instead of complete to 1 (scale)
however handling the error isn't related to loader component

packages/playground/src/hooks/useNavigationStatus.ts
```diff
@@ -0,0 +1,39 @@
...
+
+      return next();
+    });
+
+    unSubscribeAfterEach = router.afterEach((to, _, failure) => {
+      if (status.value !== "Loading" || currentToPath !== to.fullPath) {
+        return;
+      }
+
+      status.value = failure ? "Failed" : "Success";
```
amiraabouhadid, line 29: might be a good idea to display an error message if page failed to load. otherwise the progress bar just gets stuck like this 
image

MohamedElmdary: well this is not how vue-router works ... vue-router will wait till the connection back and won't notify me that an error happen so I have no idea at that point if routing will fail or success

amiraabouhadid:
I'm not sure why but I get this error even after running lerna build 
image

MohamedElmdary:
> I'm not sure why but I get this error even after running lerna build
>
> ...
this issue is not related to my pr

packages/playground/src/hooks/useNavigationStatus.ts
amiraabouhadid: Then what is the point of setting status to Failed if no action will be taken in that case

MohamedElmdary: router might fail to load component or something else

maayarosama:
> > What should happen if navigation fails?
> 
> Well, this should never happen in our case but if failed the line will return to 0 instead of complete to 1 (scale) however handling the error isn't related to loader component

can we open an issue for this please

maayarosama (approved):

Issue #1113, opened 2023-09-15
Support hex secrets in login flow (and ed25519)


scottyeager:
The recent update dropping Polkadot extension from Dashboard login leaves us in a bad situation:

1. Many farms were created with the TF Connect app. They belong to a twin that's associated with a hex secret only, not a mnemonic
2. Hex secrets can't be used to log into the dashboard, so farm management and the bridge aren't available (these features aren't in the app)

Supporting hex secrets for dashboard login is important to providing full services to all of our users, and I think it should be added urgently.

### ed25519

It also looks like the dashboard only supports the sr25519 key type. It was previously possible to create accounts with ed25519 key type using the Polkadot extension (I have one such account). This is perhaps less urgent because I don't think there are as many of these accounts as there are accounts created through the TF Connect app, but it still represents a fraction of users who can no longer use the Dashboard features that require authentication.

maayarosama:
Work Completed:
- Added the select field for the keypair type
- Added The warning

Work in Progress (WIP):
- passing whatever is selected to the grid client in the keypairType option

Investigation and Solution:

xmonader:
I think we shouldn't allow the generation of ED keys (just allow importing them)

maayarosama:
Work Completed:
- Disabled creating nee account when using hex seed
- passed the KeypairType to the grid

Work in Progress (WIP):
- the positioning of the select field 

- saving the keypair type as a static value somewhere same as mnemonic

Investigation and Solution:

maayarosama:
Work Completed:

- Re-validationg when keyPair change
- Enhancements to warning alert
- Adding an alert in case of choosing ED key telling the user that generation of ed keys isn't supported


Work in Progress (WIP):

Investigation and Solution:

khaledyoussef24:
blocked on #1690 to verify that we can create a twin with a hex seed also

khaledyoussef24:
env used : devnet
version used : 228c723
verified now user can log in and activate account from mnemonic and hex seed
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/31b5b19f-14ff-4354-82f9-623ead179df1)
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/bbfad00c-793d-4b73-a6f9-22143d8d10f2)

test lodge : https://app.testlodge.com/a/26076/projects/40893/suites/234374?expand_section=422331#case_4161528

[...]

PR #1709, opened 2023-12-17
Adding selection field for keypair type

maayarosama:
### Description

Adding selection for key pair type in profile manager
Screenshot from 2023-12-18 16-12-50

### Changes

- Added selection for  key pair type
- Added a warning for the user
- Added key pair type in profile interface
- Updated get grid method 
### Related Issues

#1113
### Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation

### DIFF
--- a/packages/playground/src/weblets/profile_manager.vue
+++ b/packages/playground/src/weblets/profile_manager.vue
@@ -564,8 +564,9 @@ const ssh = ref("");
 const mnemonicInput = useInputRef();
 const shouldActivateAccount = computed(
   () =>
-    mnemonicInput.value?.error?.toLowerCase()?.includes("couldn't get the user twin for the provided mnemonic") ||
-    false,
+    mnemonicInput.value?.error
+      ?.toLowerCase()
+      ?.includes("couldn't get the user twin for the provided mnemonic and keytype") || false,
 );
 let sshTimeout: any;
 const isValidConnectConfirmationPassword = computed(() =>
--- a/packages/playground/src/weblets/profile_manager.vue
+++ b/packages/playground/src/weblets/profile_manager.vue
@@ -197,7 +197,7 @@
                   You will need to provide the password used while connecting your wallet.
                 </p>
               </v-alert>
-              <v-autocomplete :items="[...keyType]" return-object item-title="name" v-model="keyPairType" />
+              <v-autocomplete :items="[...keyType]" item-title="name" v-model="keyPairType" />
               <PasswordInputWrapper #="{ props: passwordInputProps }">
                 <InputValidator
                   :value="password"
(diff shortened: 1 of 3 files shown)


AhmedHanafy725 (commented):
we need to re-validate the mnemonic if the keypair is changed as most probably the account will be created with a certain keypair type

AlaaElattar:
- Validation should be reloaded.
- In that case, I've added mnemonic of type ed25519 and then chose it.

[Screencast from 12-19-2023 03:00:31 PM.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/54944491/898bdf59-d761-4992-9585-2529f0e9c9f3)

AhmedHanafy725:
not sure if we should move the new warning message to the top or keep it. But if we keep it, I think it should have some spacing after @ehab-hassan 

image

AlaaElattar:
> * Validation should be reloaded.
> * In that case, I've added mnemonic of type ed25519 and then chose it.
> 
>  Screencast.from.12-19-2023.03.00.31.PM.webm

Please consider the same for `Activate Account` and `Create Account` buttons.

packages/playground/src/weblets/profile_manager.vue
```diff
@@ -252,6 +267,9 @@
               <v-alert type="error" variant="tonal" class="mt-2 mb-4" v-if="loginError">
                 {{ loginError }}
               </v-alert>
+              <v-alert variant="tonal" type="warning" v-if="activeTab === 1">
```
AlaaElattar, line 270: It feels more like an info not a warning , what do u think ?

AlaaElattar:
Also as long as the user can't create account with ed25519 (just import pre existing account), I think w should say that somewhere.

ehab-hassan:
> not sure if we should move the new warning message to the top or keep it. But if we keep it, I think it should have some spacing after @ehab-hassan
> 
> image

yes please let's keep it down and make sure to have the same margin's first warning message had.

AlaaElattar (approved):

packages/playground/src/weblets/profile_manager.vue
```diff
@@ -462,6 +490,15 @@ watch(
...
 );
+watch(
+  () => keypairType.value,
+  async (value, oldValue) => {
+    if (value !== oldValue) {
+      mnemonicInput.value?.validate();
+    }
+  },
+  { deep: false },
+);
```
amiraabouhadid, line 501: can we please update the message on line 567 to "Couldn't get the user twin for the provided mnemonic **and keytype**".

amiraabouhadid (approved):

AhmedHanafy725 (approved):

Issue #1673, opened 2023-12-12
Grid nightly tests is failing


A-Harby:
### Description

After the script is updated in this [Issue](https://github.com/threefoldtech/tfgrid-sdk-ts/issues/950), [Grid Nightly](https://github.com/threefoldtech/tfgrid-sdk-ts/actions/workflows/grid_client_nightly.yml) has been failing for a while.

Tests themselves are failing, but the issue is that their results appear to be passing.

A-Harby:
Verified is passing with no issues and reporting a failure if exists.
The workflow is now passing on [mainnet](https://github.com/threefoldtech/tfgrid-sdk-ts/actions/runs/7497879316/job/20412141156)
And reporting a failure on [devnet](https://github.com/threefoldtech/tfgrid-sdk-ts/actions/runs/7497879316/job/20412140728) and it is reported [here](https://github.com/threefoldtech/tf_operations/issues/2196).

Issue #950, opened 2023-08-09
Update all global scripts to be structured and up to date.


Mahmoud-Emad:
### Description

The entire global script requires restructuring and updating for enhanced functionality. For those intending to adopt this script, it's essential to acknowledge that "the script should be modularized into functions." This improvement is aligned with the idea that the code needs to be compliant with linting tools in the GitHub Actions workflow.

Please refer to the example provided below for guidance:

```ts

function main() {
  const grid = new GridClient();
  grid.connect();

  // List all contracts
  listAllContracts(grid);

  // Uncomment the line below if you intend to perform contract cancellation.
  // cancel(grid);
}

function listAllContracts(grid: GridClient) {
  // Utilize the contract module within the grid client to list all available contracts.
  return grid.contracts.list();
}

function cancel(grid: GridClient, contractID: number) {
  // Utilize the contract module within the grid client to cancel a specific contract.
  return grid.contracts.cancel(contractID);
}
```

mohamedamer453:
Update:

Work Done:

- Updated the structure of 17/38 scripts. #1409 
- The new structure splits the script into functions for each functionality and calls these function in the main() function.

Work In Progress (WIP)

- Updating the rest of the scripts.

mohamedamer453:
Update:

Work Done:
- Updated the rest of the scripts to the new structure.

Work In Progress (WIP):
- A couple of scripts are still remaining, still looking into how i'll re-structure them.
- More structuring to be done for each script to be more like the [`vm_with_qsfs.ts`](https://github.com/threefoldtech/tfgrid-sdk-ts/blob/development/packages/grid_client/scripts/vm_with_qsfs.ts#L36&L90) script

mohamedamer453:
Update:

Work Done:

mohamedamer453:
Update:

Work Done:

- scripts in the `scripts/tfchain` directory were not updated yet, so i updated them with the correct structure and models.
- Did a final review for the scripts and did a final update.
- PR is ready for review #1409

A-Harby:
Verified,
On development branch, 
The script has been updated to adopt three functions in every script (deploy, get deployment, and cancel).
Other scripts have also been refactored to functions.
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/c32e2743-6147-4eaf-bed3-1400ff33a2a9)

PR #1710, opened 2023-12-17
Add exit code to in case of error in test script

A-Harby:
### Description

After the update script issue, the tests have been failing but passing in the test result because of the try and catch that have been added to the code. Removed all the try and catch from the scripts and added a checker after the delete all.

Update the yggdrasil version to the latest v0.5.4.

### Changes

Only the scripts that are included in the grid nightly workflow.

### Related Issues

https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1673 
https://github.com/threefoldtech/tfgrid-sdk-ts/issues/950

https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1756

### DIFF
--- a/.github/workflows/grid_client_nightly.yml
+++ b/.github/workflows/grid_client_nightly.yml
@@ -100,7 +100,9 @@ jobs:
 
       - name: Check if contracts are Empty
         run: |
-          arrays=("${{ steps.command_step.outputs.myOutput }}")
+          contract_json=$(echo "${{ steps.command_step.outputs.myOutput }}" | sed '1d;$d')
+          # Convert the middle line to three arrays
+          IFS=', ' read -r -a arrays <<< "$(echo "$contract_json" | awk -F':|,' '{print $2, $4, $6}')"
           # Check if all three [Name - Node - Rent] contracts are empty
           if [ -n "${arrays[0]}" ] || [ -n "${arrays[1]}" ] || [ -n "${arrays[2]}" ]; then
           echo "Deletion failed; at least one array is not empty."
--- a/.github/workflows/grid_client_nightly.yml
+++ b/.github/workflows/grid_client_nightly.yml
@@ -94,6 +94,7 @@ jobs:
       - name: Run check up - List all contracts
         id: command_step
         run: |
+          sleep 15
           output=$(yarn run ts-node --project packages/grid_client/tsconfig-node.json packages/grid_client/scripts/list_all_contracts.ts)
           echo "::set-output name=myOutput::$output"
 
--- a/.github/workflows/grid_client_nightly.yml
+++ b/.github/workflows/grid_client_nightly.yml
@@ -91,6 +91,20 @@ jobs:
         run: |
           yarn run ts-node --project packages/grid_client/tsconfig-node.json packages/grid_client/scripts/delete_all_contracts.ts
 
+      - name: Run check up - List all contracts
+        id: command_step
+        run: |
+          output=$(yarn run ts-node --project packages/grid_client/tsconfig-node.json packages/grid_client/scripts/list_all_contracts.ts)
+          echo "::set-output name=myOutput::$output"
+
+      - name: Check if contracts are Empty
+        run: |
+          arrays=("${{ steps.command_step.outputs.myOutput }}")
+          # Check if all three [Name - Node - Rent] contracts are empty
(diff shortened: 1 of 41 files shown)


zaelgohary (commented):
Maybe we can throw the error instead of logging it and exiting the process? Also, not all scripts are updated.

xmonader:
We need to make sure clean up function is always executed

AhmedHanafy725:
> Maybe we can throw the error instead of logging it and exiting the process? Also, not all scripts are updated.

let's remove the try-catch block in case we need to make it crash on failure. There is no need for it

A-Harby:
- Removed try-catch from the test scripts, then all of them.
- Update the workflow to check that all the contracts are deleted after cleaning up, and if not, the workflow will fail.
- Change the time to 4 AM to run before mass deployment by 2 hours.
- Tested both [successful](https://github.com/threefoldtech/tfgrid-sdk-ts/actions/runs/7274617307/job/19820895857#step:16:1) run and deletion failure.

.github/workflows/grid_client_nightly.yml
```diff
@@ -91,6 +91,23 @@ jobs:
...
+          sleep 15
+          yarn run ts-node --project packages/grid_client/tsconfig-node.json packages/grid_client/scripts/list_all_contracts.ts > output.txt
+
+      - name: Check if contracts are Empty
+        run: |
+          # Read the Contracts line from the file
+          contracts_json=$(awk 'NR==5' output.txt)
+
+          # Check if all three [Name - Node - Rent] contracts are empty
+          if [[ $contracts_json != *'nodeContracts: []'* ]]; then
```
mohamedamer453, line 106: The comment mentions that it checks for `[Name - Node - Rent]` while the condition itself seems to check nodeContracts only.

.github/workflows/grid_client_nightly.yml
```diff
@@ -91,6 +91,23 @@ jobs:
...
+            echo "Deletion failed; The Node Contract is not empty."
+            exit 1
+          fi
```
mohamedamer453, line 109: When the this check fails it just prints a message without mentioning the failed contracts.

image

Imo it should mention the contracts that weren't deleted if possible and in case of failure it should re-run the delete all scripts again to ensure that the clean up was sucessfull.

A-Harby: There was a test for the delete all contract; if it fails, the workflow fails, and then I added the delete checker to make sure that even if the delete all task was successful, all the contracts should be checked to make sure the test is not just passing, and if not, print the failing to delete contract.

The workflow's main focus isn't to retry deleting that many but rather to notify us if those jobs are failing and the reason (especially since we don't want the delete script to just give wrong information about a successful deletion and the contracts are deleted as said).

A-Harby:
@mohamedamer453 I fixed both comments and tested both workflows. https://github.com/threefoldtech/tfgrid-sdk-ts/actions/runs/7395367664/job/20118396746 and https://github.com/threefoldtech/tfgrid-sdk-ts/actions/runs/7395403959/job/20118505842.

Could you re-review?

mohamedamer453 (commented):
In this [run](https://github.com/threefoldtech/tfgrid-sdk-ts/actions/runs/7408499704/job/20156823808#step:16:78) you can see that the contracts were deleted successfully, however the run failed in the `Check if contracts are empty` step, i think it might be giving a false positive, the issue might be that it needs a bit of time before listing all contracts to register that the contracts were deleted, so if it's possible to increase the timeout before doing this step to ensure that it's not giving a false positive.

A-Harby:
> In this [run](https://github.com/threefoldtech/tfgrid-sdk-ts/actions/runs/7408499704/job/20156823808#step:16:78) you can see that the contracts were deleted successfully, however the run failed in the `Check if contracts are empty` step, i think it might be giving a false positive, the issue might be that it needs a bit of time before listing all contracts to register that the contracts were deleted, so if it's possible to increase the timeout before doing this step to ensure that it's not giving a false positive.

This should be fixed [here](https://github.com/threefoldtech/tfgrid-sdk-ts/actions/runs/7411818244/job/20167220899).

mohamedamer453 (approved):
After the latest [run](https://github.com/threefoldtech/tfgrid-sdk-ts/actions/runs/7459638548) the workflow seems to be working fine. great work!

AhmedHanafy725 (approved):

Issue #1686, opened 2023-12-13
Farm Name input validation


A-Harby:
### Description
Devnet, 28d6c07.

Adding farm input validation will prevent this error message.

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/836ca834-b6fa-480f-a10a-6632d6573b08)

The old input validation for Farm Name was [maximum 40 ascii characters, no spaces, only -, _ are allowed as special characters]. I don't know if it was changed or not.

zaelgohary:
Check validation in tf chain and same validation should be applied here.

amiraabouhadid:
Investigative work:
looked for validation on the tf chain

amiraabouhadid:
fixed: 
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/56790126/356d908a-2e4b-4ed4-bb4a-8de4f16e2b49)
![Screenshot from 2023-12-17 16-36-57](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/56790126/66ab1d7f-28fb-4f46-8304-bbb44907f125)
![Screenshot from 2023-12-17 14-06-09](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/56790126/812d6bc4-a657-4f3c-b196-b26c006b2459)

A-Harby:
6d0ec3d
Not released yet.
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/34157b4c-f042-4f0b-b402-1a306d7ec79d)

amiraabouhadid:
moved back to pending review, was put in verification by mistake

A-Harby:
Verified,
- Devnet
- e2580a5
- [TC2241 - Farms Input validation](https://app.testlodge.com/a/26076/projects/40893/suites/234374?expand_section=422339#case_4150450)

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/7b7100c5-a055-452f-a5fd-ace08f23f498)
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/92197312-3c6f-44c7-97e2-118cf2a1e051)
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/294e3f12-ecdc-461e-8fe1-369109948047)
[...]

PR #1721, opened 2023-12-18
fix farm name validation special chars and min length

amiraabouhadid:
### Changes

add allow only -_ as special characters, make min length 3 chars to be consistent with tfchain validation

### Related Issues

#1686

### DIFF
--- a/packages/playground/src/dashboard/components/create_farm.vue
+++ b/packages/playground/src/dashboard/components/create_farm.vue
@@ -21,7 +21,7 @@
                   name => validators.isAlpha('Farm name must start with an alphabet char.')(name[0]),
                   validators.minLength('Farm name minimum length is 3 chars.', 3),
                   validators.maxLength('Farm name maximum length is 40 chars.', 40),
-                  validators.pattern('Farm name  should not contain whitespaces.', {
+                  validators.pattern('Farm name should not contain whitespaces.', {
                     pattern: /^[^\s]+$/,
                   }),
 
--- a/packages/playground/src/dashboard/components/create_farm.vue
+++ b/packages/playground/src/dashboard/components/create_farm.vue
@@ -19,11 +19,13 @@
                 :rules="[
                   validators.required('Farm name is required.'),
                   name => validators.isAlpha('Farm name must start with an alphabet char.')(name[0]),
-                  validators.minLength('Farm name minimum length is 2 chars.', 2),
+                  validators.minLength('Farm name minimum length is 3 chars.', 3),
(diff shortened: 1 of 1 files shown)


packages/playground/src/dashboard/components/create_farm.vue
```diff
@@ -92,11 +94,17 @@ export default {
         isCreating.value = false;
       }
     }
+    function validateFarmName(name: string) {
+      if (!name.split("").every((c: string) => /[a-zA-Z0-9\-_]/.test(c))) {
+        return { message: "Invalid Farm Name" };
```
0oM4R, line 99: I suggest to make the message contain the name rules,  what characters are allowed and so on

amiraabouhadid: done, thanks

packages/playground/src/dashboard/components/create_farm.vue
```diff
@@ -19,11 +19,13 @@
...
                   validators.required('Farm name is required.'),
                   name => validators.isAlpha('Farm name must start with an alphabet char.')(name[0]),
-                  validators.minLength('Farm name minimum length is 2 chars.', 2),
+                  validators.minLength('Farm name minimum length is 3 chars.', 3),
                   validators.maxLength('Farm name maximum length is 40 chars.', 40),
                   validators.pattern('Farm name  should not contain whitespaces.', {
                     pattern: /^[^\s]+$/,
                   }),
+
+                  validateFarmName,
```
MohamedElmdary, line 28: can we use validators.pattern like the above one ?

0oM4R: and please remove the extra space in line 24 after `Farm name`

amiraabouhadid: validators.pattern checks only if pattern exists, it does not check if every character satisfies pattern.

packages/playground/src/dashboard/components/create_farm.vue
```diff
@@ -92,11 +94,17 @@ export default {
         isCreating.value = false;
       }
     }
+    function validateFarmName(name: string) {
+      if (!name.split("").every((c: string) => /[a-zA-Z0-9\-_]/.test(c))) {
+        return { message: "Farm name can only contain alphabetic letters, numbers, '-' or '_'" };
+      }
+    }
```
zaelgohary, line 101: Can we use the pattern method from validators instead?

maayarosama (approved):

AlaaElattar (approved):

0oM4R (approved):

Issue #1707, opened 2023-12-17
Add  a note on Your Farms page to explain if the user doesn't have farms


xmonader:
Show a small hint if there're no farms to the user "Can't see any of your farms?" with some more information that they needed to import the ed key or something

A-Harby:
Verified, 
- Devnet
- e2580a5
- [TC2246 - List farms](https://app.testlodge.com/a/26076/projects/40893/suites/234374?expand_section=422339#case_4150455)


![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/316aadf6-933e-4811-aa17-5095d6b1481a)

ramezsaeed:
Verified:
- Devnet
- ebf75a3

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/16083592/46e62033-9edf-46da-b523-99b5fa489d07)

PR #1722, opened 2023-12-18
add hint when no farms

amiraabouhadid:
### Description

added a hint to the user to change key types from TFChain wallet when he/she cannot see their farms. 

### Related Issues

#1707

image

### DIFF
--- a/packages/playground/src/dashboard/components/user_farms.vue
+++ b/packages/playground/src/dashboard/components/user_farms.vue
@@ -7,7 +7,7 @@
       single-line
       hide-details
     ></v-text-field>
-    <v-alert type="info" variant="tonal" v-if="farmsCount == 0">
+    <v-alert type="warning" variant="tonal" v-if="farmsCount == 0" class="my-8">
       Can't see any of your farms? Try changing your key type in your TFChain Wallet above.
     </v-alert>
     <v-data-table-server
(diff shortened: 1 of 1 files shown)


packages/playground/src/dashboard/components/user_farms.vue
```diff
@@ -7,6 +7,9 @@
       single-line
       hide-details
     ></v-text-field>
+    <v-alert type="info" variant="tonal" v-if="farmsCount == 0">
```
MohamedElmdary, line 10: what do u think about styles should we add margin bottom?
@amiraabouhadid @ehab-hassan

ehab-hassan: yes let's add some margin-bottom and I think a warning color here like this
image

amiraabouhadid: done, thanks

AlaaElattar:
- I don't understand the point of the hint. maybe the user has no farms, why to mention keytype ?

ehab-hassan (approved):
Approved at design

amiraabouhadid:
> * I don't understand the point of the hint. maybe the user has no farms, why to mention keytype ?

please refer to #1113

Issue #1760, opened 2023-12-21
Limit validation tasks' tries to 1


MohamedElmdary:
useAsync by default tries 3 times which is not needed in case of validation

khaledyoussef24:
is there a way i can verify this ?

MohamedElmdary:
> is there a way i can verify this ?

Yes basically check if network making a single request while select a node manually (if node is not valid)

khaledyoussef24:
verified 
![Screenshot from 2024-01-02 07-11-05](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/d9bfdf95-c773-4e8a-a8e3-aab100c1e2d8)
![Screenshot from 2024-01-02 07-10-59](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/e33b9402-db72-4998-b8b8-a3bffe50ce3c)

ramezsaeed:
Verified:
- Devnet
-  2.3.0-alpha3

PR #1761, opened 2023-12-21
fix: add limit 1 to validation tries

MohamedElmdary:
### Description
While working with useAsync I notice that validation is done 3 times (if valid) which not bad in case a user have bad connection but too bad if task takes long time 

### Changes
limit tries to 1 to validation tasks

### Related Issues
- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1760

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/packages/playground/src/components/node_selector/TfAutoNodeSelector.vue
+++ b/packages/playground/src/components/node_selector/TfAutoNodeSelector.vue
@@ -210,6 +210,7 @@ export default {
     const nodeInputValidateTask = useAsync<true, string, [NodeInfo | undefined]>(
       node => checkNodeCapacityPool(gridStore, node, props.filters),
       {
+        tries: 1,
         shouldRun: () => props.validFilters,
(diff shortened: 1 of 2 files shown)


AhmedHanafy725 (approved):

zaelgohary (commented):
Do we still need this check?
https://github.com/threefoldtech/tfgrid-sdk-ts/blob/52cb84ceb34ea9819e785d2f34bbaef65c83c001/packages/playground/src/hooks/useAsync.ts#L37

MohamedElmdary:
> Do we still need this check?
> 
> https://github.com/threefoldtech/tfgrid-sdk-ts/blob/52cb84ceb34ea9819e785d2f34bbaef65c83c001/packages/playground/src/hooks/useAsync.ts#L37



It should be fine to keep the check as someone might pass a wrong value e.g -1

zaelgohary (approved):

Issue #1725, opened 2023-12-18
Decimal validation on the dedicate node resources filters


A-Harby:
### Description

Devnet, 13a87cf.

Dedicate node table filters: entering a decimal number with more than 1 digit doesn't filter or give an error.
This applies to all the resource filters on the page, too.

[Threefold Dashboard: Nodes.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/1758f495-a7f1-43db-b125-13f490fd0e92)

ramezsaeed:
![Screenshot from 2023-12-21 14-22-58](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/16083592/914118d6-5088-458b-b5fe-01a250720907)

khaledyoussef24:
verified now decimal validations works well with dedicated node 
![Screenshot from 2023-12-26 13-24-05](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/1f6a131c-d3ec-4380-88ac-2374e4a4ff61)
![Screenshot from 2023-12-26 13-23-46](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/c9fb9acf-cc7e-427b-ad8c-f691661df939)

testlodge :https://app.testlodge.com/a/26076/projects/40893/suites/234374?expand_section=422335#case_4162539

ramezsaeed:
Verified:
- DEvnet
- ec2ed3d

![Screenshot from 2023-12-27 12-16-35](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/16083592/45c67255-e646-4291-b868-860cda18396c)

PR #1772, opened 2023-12-21
Fix decimal err in dedicated filters

zaelgohary:
### Description

- Used Math.floor with values as Grid Proxy doesn't accept float values
- Refactored NodesQuery
- Added filter by total resources in nodes page
- Fixed formatResources
- Fixed filter by decimal in nodes & dedicated nodes

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1725
- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1783

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation


0oM4R:
Screenshot from 2023-12-24 11-13-58

Ram field does not accept decimal number
and i don't think it's a good idea to use floor, I suggest to use ceil

amiraabouhadid:
I tried it with exact RAM value but it failed 
image
Screenshot from 2023-12-21 15-02-56

zaelgohary:
> I tried it with exact RAM value but it failed image image

The value in the table was rounded. Fixed by showing the value without any rounding.

amiraabouhadid (approved):
thanks for adding it nodes page too
image

0oM4R (approved):

Issue #1781, opened 2023-12-24
Enhance disconnect method 


0oM4R:
https://github.com/threefoldtech/tfgrid-sdk-ts/blob/832d4d9ee5e944d677db392f6e015579f55704d2/packages/grid_client/src/client.ts#L124

this line should moved to `this.disconnect` method

0oM4R:
### Investigation and Solution

- the current solution make the client disconnect twice
- will remove event listeners from grid client as each client has its own event listeners.

0oM4R:
couldn't reproduce disconnecting twice issue, will check with Hanafy later for this point

AhmedHanafy725:
the example script on the direct client is not working as the close method has been removed

AhmedHanafy725:
requested changes are done

PR #1796, opened 2023-12-25
Fix the disconnection when killing the process

AhmedHanafy725:
### Description

Fix the disconnection when killing the process

### changes

-  remove event listeners from grid client as each client has its own event listeners 
- remove `disconnectAndExit`; it has no effect as we removed the event listeners, each client now have its `process.removeAllListeners();`
- rmb direct client
  - merge `close` to `disconnect` function in rmb direct client
  - in `catch` in connect method 
    - we remove the `c.close` as it will be done in disconnect 
    - if the user doesn't have enough balance no need to disconnect  
  - in `disconnectAndExit` 
    - clear timeout 
    - close all connections 
    - remove all event listeners as mentioned earlier
- tfchain `disconnectAndExit` : we need to close all connection when killing the process, so now it loops through `QueryClient.connections` map and pass all its urls to close the connection using its `disconnectHandler` 

### Related Issues

- #1781 

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/packages/grid_client/src/client.ts
+++ b/packages/grid_client/src/client.ts
@@ -120,7 +120,6 @@ class GridClient {
     }
 
     await this.testConnectionUrls(urls);
-    await this.rmbClient.connect();
 
     if (BackendStorage.isEnvNode()) {
       process.on("SIGTERM", this.disconnectAndExit);
@@ -133,8 +132,8 @@ class GridClient {
       };
       window.onunload = this.disconnect;
     }
-
     this._connect();
+    await this.rmbClient.connect();
     await migrateKeysEncryption.apply(this, [GridClient]);
   }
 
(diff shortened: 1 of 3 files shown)


zaelgohary (dismissed):

amiraabouhadid (approved):

AlaaElattar:
- shouldn't it print `disconnecting` after `Hello 2` ??

image

AlaaElattar:
> * shouldn't it print `disconnecting` after `Hello 2` ??
> 
> image

- my bad, when i use `connect` instead of `_connect` the script worked well.

amiraabouhadid (approved):

MohamedElmdary (approved):

Issue #1730, opened 2023-12-19
Enhance the performance of validating the resource while deploying the workloads


AhmedHanafy725:
This method needs to check the resources of the nodes in parallel (use `Promise.all`) way instead of waiting for each deployment to finish validating.

https://github.com/threefoldtech/tfgrid-sdk-ts/blob/13a87cfb8fb21aced31cd2eab5aa4f597a84fc25/packages/grid_client/src/high_level/twinDeploymentHandler.ts#L319

Also, make sure to sum all the resources to be validated against each node as if there is a deployment containing a VM with a public ip, it won't be merged with the other workload in the same deployment. So we may have 2 or more different deployments on the same node.

zaelgohary:
WIP: still trying to understand the file structure and the changes needed.

AhmedHanafy725:
the required enhancement has been done

PR #1810, opened 2023-12-27
Enhance performance of workloads validating

MohamedElmdary:
### Description
Enhance the performance of validating the resource while deploying the workloads

### Changes
- refactor: use Promise.all instead of for await to parallelize execution
- fix: group all deployments with same nodeId and sum all it's resources before validating

### Related Issues
- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1730

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/packages/grid_client/src/high_level/twinDeploymentHandler.ts
+++ b/packages/grid_client/src/high_level/twinDeploymentHandler.ts
@@ -7,7 +7,6 @@ import {
   TimeoutError,
   ValidationError,
 } from "@threefold/types";
-import groupBy from "lodash/fp/groupBy.js";
 
 import { RMB } from "../clients";
 import { TFClient } from "../clients/tf-grid/client";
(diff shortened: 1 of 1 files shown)


packages/grid_client/src/high_level/twinDeploymentHandler.ts
```diff
@@ -7,6 +7,7 @@ import {
   TimeoutError,
   ValidationError,
 } from "@threefold/types";
+import groupBy from "lodash/fp/groupBy.js";
```
AhmedHanafy725, line 10: `lodash` is not in the client's dependencies and no need to add it just to group the deployments.

AhmedHanafy725 (changes requested):

packages/grid_client/src/high_level/twinDeploymentHandler.ts
MohamedElmdary: I'm just using groupBy and imported only it's file might be ~ 1 KB at most
also lodash is installed somewhere and shared by lerna 
should we implement our own groupby and avoid using it here?

AhmedHanafy725: yes, cause the client is published alone without the playground which has it.

zaelgohary (approved):

AhmedHanafy725 (approved):

Issue #1872, opened 2024-01-03
Redisign Nodes in solutions to include more details 


AlaaElattar:
### Description
- Add more details about information of each node.
- modify design to match needed resources.

A-Harby:
Verified, Devnet 21ef1d6.

[TC2461 - Node details in Load Nodes List](https://app.testlodge.com/a/26076/projects/40893/suites/234374?expand_section=422336#case_4177821)

The redesign really looks a lot better than before; all the data is visible in a good way.
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/2d9c1878-2696-419c-a90f-825d1fcfb548)
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/3d9af5c1-8554-4825-8073-aa84005b09aa)

ramezsaeed:
Verified:
- Devnet
- b43ea4c

![Screenshot from 2024-01-15 15-34-35](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/16083592/ff267dbb-1443-4772-a461-5a9ea05ac548)

Issue #1870, opened 2024-01-03
load Nodes buttun should retun an error if the selected farm doesn't have a node with the needed resources


ramezsaeed:
### Description

load nodes button doesn't retun any feedback if the selected farm doesn't have a node with the needed resources

### Steps to reproduce
- Go to solution tab
- select Micro VM
- select freefarm
- press load nodes (working fine and automatically select a node)
- go back and change the farm to select animalfarm
- press load nodes (can't find a node)

### Fix
A clear message should be shown to the user:
" this farm doesn't have a node with the needed resources.!"

### Logs/Alerts

![Screenshot from 2024-01-03 16-23-27](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/16083592/8b0f2271-a43d-4839-8be1-4ee35d0921ca)

A-Harby:
Verified, Devnet 21ef1d6.

[TC2460 - Filter a farm with no online nodes](https://app.testlodge.com/a/26076/projects/40893/suites/234374?expand_section=422336#case_4177820)

If the farm doesn't have any nodes, a red warning error will appear with the message 'No nodes were found!'.
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/dc239b5c-bcc1-43e4-b9d7-a2e08b4e5d2b)

ramezsaeed:
Verified:
- devnet
- b43ea4c

![Screenshot from 2024-01-15 15-33-06](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/16083592/939c1536-e6b6-46c2-ae6a-ef0c1aba10fc)

PR #1884, opened 2024-01-04
Improve node selector details

MohamedElmdary:
### Description
Improve node selector details

### Changes
- redesign node card
- write a selectable table (custom component) to work as VSelect

### Related Issues
- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1872
- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1870

## SS
1. Initial Loading (loading while there is no loaded nodes yet)
image

2. After first loading pick a valid node and scroll down to that node (to it's center) smoothly
image

3. Show load more nodes (if there is more nodes to load)
image

4. Ask to reload node if data changed
image

5. Show error if the provided filters is not valid
image

6. Show message while validating
image

7. Show loading linear line (if loading and there is nodes already there)
image

8. Got hint as before
image

9. Has hint to notify user that cpu might be more than 100% (usage)
image

10. some more tooltips for more details
image
image

11. Deploy button disabled
image
image

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/packages/playground/src/components/node_selector/TfAutoNodeSelector.vue
+++ b/packages/playground/src/components/node_selector/TfAutoNodeSelector.vue
@@ -1,95 +1,103 @@
 <template>
   <section>
-    <VFadeTransition>
-      <VAlert type="warning" class="mb-4 mt-1" v-if="filtersUpdated && validFilters">
-        Please press on <strong>Load Nodes</strong> button to list nodes matching your new requirements.
-      </VAlert>
-    </VFadeTransition>
+    <VBtn
+      variant="tonal"
+      color="secondary"
+      class="mb-4"
+      size="x-large"
+      block
+      @click="resetPageAndReloadNodes()"
+      :loading="pageCountTask.loading || nodesTask.loading"
+      :disabled="nodeInputValidateTask.loading || !validFilters"
+    >
+      Load Nodes
+    </VBtn>
 
-    <!-- <TfNodeDetailsCard :node="$props.modelValue" v-if="$props.modelValue" /> -->
+    <input-tooltip tooltip="Select a node ID to deploy on." align-center>
+      <div class="w-100" :style="{ position: 'relative' }">
+        <VProgressLinear
+          v-if="loadedNodes.length > 0 && (pageCountTask.loading || nodesTask.loading)"
+          indeterminate
+          :style="{
+            position: 'absolute',
+            top: 0,
+            left: 0,
+            zIndex: 9,
+            transform: 'none',
+            width: 'calc(100% - 16px)',
+          }"
+          height="2px"
+          color="primary"
+        />
 
-    <input-tooltip tooltip="Select a node ID to deploy on.">
-      <div class="d-flex w-100">
-        <VSelect
-          class="node-selector"
-          eager
-          ref="nodeInput"
-          label="Node"
-          placeholder="Select node"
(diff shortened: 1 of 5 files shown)


AlaaElattar:
i think the separator between cards should be stronger to be clear. what do u think ?

MohamedElmdary:
> i think the separator between cards should be stronger to be clear. what do u think ?

VDivider style doesn't match with outer border so I added a border bottom instead

xmonader (changes requested):
Let's not show a serial if it's default string

AhmedHanafy725 (commented):
what does it mean the shared chip on the farm? I prefer to have farm name instead.

mohamedamer453 (commented):
- After the nodes are loaded it randomly scrolls to one of the nodes, i think it should load the nodes and leave it up to the user to select the desired node.

[Screencast from 08-01-24 15:54:16.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/101194226/d35033ab-e0e5-452d-8996-aafac77be7b6)

- I believe as it was agreed upon before that the user doesn't need to know anything about the term storage pools and it would be better if the message was updated to something else like "checking if the deployment will fit in the node's disks" or something similar.

  image


- IMO it would look better if the node id was written in the same way as the farm id. e.g. "Node ID: 11" instead of using brackets.

   image

- In the manual filter, when the dedicated node filter is enabled and a shared node id is added it should mention that this node is shared in the error message instead of mentioning that this node is not rented.

   image

- On mainnet, some nodes appeared without flag. e.g. node 2336

 image

even though the node has a country in its details from the gridproxy

image

- After Loading more nodes the storage pools check is done even when the same node is selected.

   [Screencast from 08-01-24 16:40:48.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/101194226/af3951f8-a3b2-400f-b590-c5a6769a1fdd)

MohamedElmdary:
> what does it mean the shared chip on the farm? I prefer to have farm name instead.

For farmName we need extra request

MohamedElmdary:
> * After the nodes are loaded it randomly scrolls to one of the nodes, i think it should load the nodes and leave it up to the user to select the desired node.

It was a requested feature to improve ux

mohamedamer453:
> * After Loading more nodes the storage pools check is done even when the same node is selected.
>   

This issue is still present

MohamedElmdary:
> > * After Loading more nodes the storage pools check is done even when the same node is selected.
> 
> This issue is still present


It's verify that node still valid

packages/playground/src/components/node_selector/TfAutoNodeSelector.vue
```diff
@@ -1,93 +1,127 @@
...
+                      ? nodeInputValidateTask.loading
+                        ? 'Pending'
+                        : nodeInputValidateTask.data
+                        ? 'Valid'
+                        : 'Invalid'
+                      : 'Init'
+                  "
+                />
+              </div>
+              <!-- <div class="border-b" :style="{ borderBottomWidth: '2px !important' }" /> -->
```
mohamedamer453, line 77: remove the commented line

mohamedamer453 (approved):

0oM4R (approved):

xmonader (approved):

Issue #1902, opened 2024-01-09
The country and farm name filter should have auto complete


AhmedHanafy725:
this should be applied on all pages

0oM4R:
this issue is blocked on #1877, need to be merged first to apply this feature

0oM4R:
Work completed:
Applied for node finder 

Investigation and Solution:
will add those fields to farm finder page

0oM4R:
### Work Completed:
we decided that the current logic will not scale as wanted, we drafted the old pr and now we are enhancing the filters, almost done.

### Investigation and solution:
- will apply the new filters to the dedicated machine page 
- add query route to the filter component

khaledyoussef24:
verified on devnet
version : 7b55dd4
farm names are auto completed and you can not insert unrecognized farm name
for the nodes the countries now are in a drop down menu
[Screencast from 23 ينا, 2024 EET 08:31:08 م.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/450c206a-5f77-47cb-80bb-ef2c8fa5e0f6)

[Screencast from 23 ينا, 2024 EET 08:37:54 م.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/cf69eeb6-1825-45ea-a7ee-0c7a4e7eb5eb)
![Screenshot from 2024-01-23 20-37-50](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/39d0187c-8f81-43e4-8c87-586b7baf4acc)

ramezsaeed:
Verified:
- Devnet
- 7b55dd4

![Screenshot from 2024-01-24 15-08-13](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/16083592/17ddaae9-2512-4da6-9b81-72e34c290462)

Issue #1901, opened 2024-01-08
Country field: to autocomplete or to be a dropdown


xmonader:
Instead of `Country full name` and to make it clear the country names in English, I suggest we show the available choices as a dropdown

0oM4R:
Work completed:
we decided to use the farm and location components, and it done at https://github.com/threefoldtech/tfgrid-sdk-ts/pull/1915

Investigation and Solution:
waiting for #1877 to avoid conflicts on nodes and farms pages

0oM4R:
Work completed:
Applied for node finder 

Investigation and Solution:
will add those fields to farm finder page

0oM4R:
### Work Completed:
we decided that the current logic will not scale as wanted, we drafted the old pr and now we are enhancing the filters, almost done.

### Investigation and solution:
- will apply the new filters to the dedicated machine page 
- add query route to the filter component

khaledyoussef24:
verified on devnet
version : 7b55dd4
countries are now drop down menu
verified scenarios, you can not insert an unknown or unrecognized name.
[Screencast from 24 ينا, 2024 EET 07:34:51 ص.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/b12745fe-c922-4197-bae8-3037cecd73e6)
![Screenshot from 2024-01-24 07-34-43](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/47ad7b18-641a-42fb-8597-cec73ba95f99)
![Screenshot from 2024-01-24 07-34-29](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/4c61ace7-34cd-433c-a27e-c6f14042a84c)

Issue #1869, opened 2024-01-03
regions don't exist in the nodes view filters


xmonader:
need to add region in the filters too

Mahmoud-Emad:
Should be implemented in https://github.com/threefoldtech/tfgrid-sdk-ts/pull/1877

Mahmoud-Emad:
## Issue Update

### Work Completed:
- Added the `region` query in the `proxyClient` package.
### Work in Progress (WIP):
Detail all ongoing efforts and remaining tasks related to this issue. Clearly outline the items that are currently being worked on and those that still need to be addressed.

### Investigation and Solution:
If there has been no work completed or work in progress, elaborate on the investigative work undertaken to address the issue. Provide insights into the problem and, if a solution was reached, include it.

A-Harby:
Can you change this input into list selection (as node status) since all the regions are already specified and can't be changed, and it's better for the UX?

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/2402e2e7-0570-4252-b81c-39724b383be4)

0oM4R:
> Can you change this input into list selection (as node status) since all the regions are already specified and can't be changed, and it's better for the UX?
> 
[...]

PR #1975, opened 2024-01-14
[Node finder] make farm name and location fields as autocomplete fields

0oM4R:
### Description

using `TFSelectLocation` and `TFSelectFarm` components to provide searching and auto completion;  

### Changes

- in `TfSelectLocation.vue` we added a prop `NodeStatus` to get the locations based on the provided node status
 - This value is used in getLocations function, so in solutions <[TfSelectionDetails.vue](https://github.com/threefoldtech/tfgrid-sdk-ts/compare/development_autoCompleteFilters?expand=1#diff-b11587292fa1b7595af2337c33cb564c177389b82ed451a4ef607d955890275a)> we are passing this prop as `NodeStatus.Up` to get only up nodes,  but in filters we need to get all locations whatever the node status is, 
- added a slot in filters to prepend the region, country and farm name fields
- the rest of changes is related to the now fields added

image

> this pr is only for node finder will create another one for Farm Finder
### Related Issues

- #1901 
- #1902
- #1869

### Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation


MohamedElmdary (approved):

packages/playground/src/components/node_selector/TfSelectLocation.vue
Mahmoud-Emad, line 1: Although Vue conventionally follows PascalCase for naming components, we have used snake_case in our project, do you think we need to change the file name in this PR?

packages/playground/src/utils/filter_nodes.ts
```diff
@@ -48,45 +45,30 @@ export type DedicatedNodeFilters = {
...
-  farmIds: {
-    label: "Farm ID",
+  nodeId: {
+    label: "Node ID",
     rules: [
       [
         isNumeric("This field accepts numbers only.", { no_symbols: true }),
-        min("The ID should be larger than zero.", 1),
-        isInt("should be an integer"),
+        min("The node id should be larger then zero.", 1),
```
Mahmoud-Emad, line 66:
```suggestion
        min("The node ID should be larger then zero.", 1),
```

packages/playground/src/utils/filter_nodes.ts
```diff
@@ -48,45 +45,30 @@ export type DedicatedNodeFilters = {
...
+        startsWith("The node id start with zero.", "0"),
```
Mahmoud-Emad, line 67:
```suggestion
        startsWith("The node ID start with zero.", "0"),
```

packages/playground/src/utils/filter_nodes.ts
```diff
@@ -48,45 +45,30 @@ export type DedicatedNodeFilters = {
...
-  },
-  country: {
-    label: "Country Full Name",
-    type: "text",
-    tooltip: "Filter by country.",
-  },
-  region: {
-    label: "Region Name",
-    tooltip: "Filter by region.",
+    tooltip: "Filter by node id.",
```
Mahmoud-Emad, line 71:
```suggestion
    tooltip: "Filter by node ID.",
```

packages/playground/src/utils/get_nodes.ts
```diff
@@ -178,9 +178,9 @@ export const getQueries = (mixedFilters: MixedFilter): Partial<NodesQuery> => {
     retCount: true,
     nodeId: +mixedFilters.inputs.nodeId.value! || undefined,
     farmIds: mixedFilters.inputs.farmIds.value,
-    farmName: mixedFilters.inputs.farmName.value,
-    country: mixedFilters.inputs.country.value,
-    region: mixedFilters.inputs.region.value,
+    farmName: mixedFilters.options.farmName,
+    country: mixedFilters.options.country,
+    region: mixedFilters.options.region,
```
Mahmoud-Emad, line 183: Why was it removed from the inputs and added to the options?

The distinction between `inputs` and `options` is as follows:

- Inputs: Any input provided by the user, such as `input`, `select`, `autocomplete`, and so on.

- Options: Filters that can be applied to the inputs, such as the page number, table count, switches, and so on.

packages/playground/src/views/nodes.vue
```diff
@@ -119,10 +168,10 @@ export default {
     const nodesCount = ref<number>(0);
     const filtering = ref(false);
     const selectedNodeId = ref<number>(0);
-
+    const farm = ref();
     const isDialogOpened = ref<boolean>(false);
     const isValidForm = ref<boolean>(true);
-
+    const location = ref();
```
Mahmoud-Emad, line 174: Please indicate the type of `farm` and `location` in the respective fields.

Mahmoud-Emad:
I think there is something unusual in this behavior. Can you please check this video?

  [Screencast from 01-15-2024 11:26:44 AM.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/57001890/b1623f95-7032-49b6-bbc3-f9ccfbeea67e)

The query params are used to save the initial state and compare it to the current state.

Issue #1970, opened 2024-01-14
[Node Finder] Clear button is not functional after opening node details dialog


0oM4R:
### Description

As we discussed earlier, Clear button should update the table if the applied fields are not in the default case  

### Steps to reproduce

- go to node finder 
- select any field in this case i will select Gpu then click on apply 
- click on clear; you will notice that the table is not changed back to the init state

[Screencast from 14 ينا, 2024 EET 02:42:00 م.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/62248851/6b6ebf83-7e9a-4baa-b7d4-f8df5871d9e2)

### Logs/Alerts

Logs/Alerts if available

0oM4R:
Work Completed:
we decided that the current logic will not scale as wanted, we drafted the old pr and now we are enhancing the filters, almost done.

Investigation and solution:
will apply the new filters to the dedicated machine page
add query route to the filter component

khaledyoussef24:
verified working fine now
env used : devnet
version :  9d076f1
[Screencast from 22 ينا, 2024 EET 11:07:48 م.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/860a3422-6771-4c11-9645-423d6e7520f1)

testlodge : https://app.testlodge.com/a/26076/projects/40893/suites/234374?expand_section=426661#case_4150464

ramezsaeed:
Verified:
- Devnet
- 7b55dd4

Issue #1957, opened 2024-01-11
Table filters should disable apply button if the values not changed


0oM4R:
> * if the filter values are different than the current state values, the `apply` button should be active.

- type in any field then click on apply
- go to another field and type any thing and delete it without clicking apply 
- the expected scenario is to disable the `Apply` btn but it was enabled

[Screencast from 11 ينا, 2024 EET 05:19:37 م.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/62248851/1d7b3b19-1d34-4040-8e26-8dd15b64bb28)

_Originally posted by @0oM4R in https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1877#issuecomment-1887427122_

0oM4R:
Work Completed:
we decided that the current logic will not scale as wanted, we drafted the old pr and now we are enhancing the filters, almost done.

Investigation and solution:
will apply the new filters to the dedicated machine page
add query route to the filter component

khaledyoussef24:
working and verified on devnet
version : 9d076f1
![Screenshot from 2024-01-23 07-22-52](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/28e5d944-42d3-463a-b8d2-deefdf34e91e)
[Screencast from 23 ينا, 2024 EET 07:22:30 ص.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/31707b1b-6024-477a-95ac-02251700b5cd)

ramezsaeed:
why I can see a red line under the farm name !!
that feels there is something wrong!

![Screenshot from 2024-01-24 15-46-15](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/16083592/1784e6b7-f7a7-4623-8fb0-ce603f263b41)

khaledyoussef24:
the redline is not visible anymore 
<img width="455" alt="Screenshot 2024-01-31 at 3 56 29 PM" src="https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/f3f8c7a7-8ccb-4c7b-93cd-f8e6a36e56cd">
<img width="408" alt="Screenshot 2024-01-31 at 3 57 09 PM" src="https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/de03738e-852c-4cf5-923f-650b4b1552f9">

ramezsaeed:
Verified:
[...]

PR #1982, opened 2024-01-15
Improve filters

MohamedElmdary:
### Description

Describe the changes introduced by this PR and what does it affect

### Changes
Rewrote filters

### Related Issues

- #1869 
- #1901 
- #1902 
- #1957 
- #1970 

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation


packages/playground/src/components/filters/TfFiltersContainer.vue
```diff
@@ -0,0 +1,135 @@
...
+              <VRow no-gutters>
+                <slot />
+              </VRow>
+
+              <VRow class="mb-4" no-gutters v-show="valid && (changed || (!loading && !empty))">
+                <VAlert type="info" variant="tonal">
+                  <span>
+                    {{ changed ? "Filtering options updated but not applied." : "" }} Click
+                    <VCard class="d-inline pa-1" v-text="changed ? 'Apply' : 'Clear'" flat />
+                    {{ changed ? "inorder to reloading your data." : "to reset your selected filters." }}
```
mohamedamer453, line 18:
```suggestion
                    {{ changed ? "Filter options were updated but not applied." : "" }} Click
                    <VCard class="d-inline pa-1" v-text="changed ? 'Apply' : 'Clear'" flat />
                    {{ changed ? "in order to reload your data." : "to reset your selected filters." }}
```

packages/playground/src/dashboard/dedicated_nodes_view.vue
```diff
@@ -89,56 +192,69 @@ const tabParams = {
...
+      page: page.value,
+      totalSru: convert(filters.value.minSSD),
+      totalMru: convert(filters.value.minRAM),
+      totalHru: convert(filters.value.minHDD),
+      totalCru: +filters.value.minCPU || undefined,
+      gpuVendorName: filters.value.gpuVendorName || undefined,
+      gpuDeviceName: filters.value.gpuDeviceName || undefined,
+      hasGpu: filters.value.gpu,
+    });
+
```
mohamedamer453, line 240: remove console log if it's not needed.

mohamedamer453 (commented):
- In  the country, region and farms filters i think it would look better if all countries, regions and farms are available by default instead of showing only the countries, regions and farms that are available at the moment.

AlaaElattar:
- The highlight on Apply or Clear button in both dark and light modes, looks weird.
- Can we make it look better ?
image

0oM4R:
> * In  the country, region and farms filters i think it would look better if all countries, regions and farms are available by default instead of showing only the countries, regions and farms that are available at the moment.

the available country and region comes from all node status; up,standby and down. if the country is not listed on all of them why would we show it to the user? 
also the farm field aims to search and auto complete, is it really needed to load all farms ?

mohamedamer453:
> the available country and region comes from all node status; up,standby and down. if the country is not listed on all of them why would we show it to the user?

If that's the case then i believe it's not working correctly, for example i'm running it locally on mainnet and in the country field when i search for egypt it doesn't show it.

image

even when there're nodes in egypt from the gridproxy. https://gridproxy.grid.tf/nodes?country=egypt

mohamedamer453:
> also the farm field aims to search and auto complete, is it really needed to load all farms ?

for the farms, i didn't notice the `Load More Farms` button, i believe that's more than enough with the search/auto complete

MohamedElmdary:
> * The highlight on Apply or Clear button in both dark and light modes, looks weird.
> * Can we make it look better ?
>   image

New Style

image
image

MohamedElmdary:
> > the available country and region comes from all node status; up,standby and down. if the country is not listed on all of them why would we show it to the user?
> 
> If that's the case then i believe it's not working correctly, for example i'm running it locally on mainnet and in the country field when i search for egypt it doesn't show it.
> 
> image
> 
> even when there're nodes in egypt from the gridproxy. https://gridproxy.grid.tf/nodes?country=egypt

Fixed now location component reload all countries

mohamedamer453 (approved):

Mahmoud-Emad:
Please fix the conflict

Mahmoud-Emad (approved):

Issue #1469, opened 2023-11-26
Health check all of the 3rd party services


xmonader:
To increase user satisfaction, we should implement basic healthchecks that periodically check the performance of all third-party services we rely on, such as RMB, TFchain, GridProxy, GraphQL, and others. These health checks should be executed at least once a minute? the intervals can be rethought , and any detected performance degradation should be promptly displayed in the user interface and trigger notifications to alert users and system administrators. By proactively monitoring and addressing performance issues, we can minimize disruptions and provide a seamless user experience.

Note: the alerting of the system administrators is going to be addressed in another issue

0oM4R:
### Work completed: 
created a monitoring package that has healthcheck classes to ping each service;
### investigation 
now we have a class for each service but i don't know how exactly will create the instance of those classes, the idea is we can't only have a method that pings the urls, some of the services will pinged through http request, polkadot  version method, and rmb ping method, so will handle the implementation of them later

0oM4R:
Work completed: 
created an interface and some classes that implements it and do some internal logic;
currently we can run this script to ping the provided services: 
```ts 

```import { servicesLiveChecker } from "../src/index";
async function HealthCheck() {
  try {
    console.log(await servicesLiveChecker("fakeURL", "", "wss://tfchain.dev.grid.tf/ws", "wss://relay.dev.grid.tf"));
    process.exit(0);
  } catch (err) {
    console.log(err);
  }
}

HealthCheck();
```

and the output:
``` bash
2024-01-21 19:35:09        API/INIT: RPC methods not decorated: transaction_unstable_submitAndWatch, transaction_unstable_unwatch
2024-01-21 19:35:09        API/INIT: Not decorating runtime apis without matching versions: TransactionPaymentApi/4 (1 known), Metadata/2 (1 known)
2024-01-21 19:35:10        API/INIT: RPC methods not decorated: transaction_unstable_submitAndWatch, transaction_unstable_unwatch
2024-01-21 19:35:10        API/INIT: Not decorating runtime apis without matching versions: TransactionPaymentApi/4 (1 known), Metadata/2 (1 known)
disconnecting
{ GraphQl: 'Down', TFChain: 'Alive', RMB: 'Alive' }
``

will enhance the code and provide a way to check the life of custom service if needed,
some enhancements to do:
- pass service urls as object;
- provide retries 
- stream errors

0oM4R:
the mentioned script is totally abstract one;
[...]

PR #2045, opened 2024-01-18
Development monitoring

0oM4R:
### Description

This package provides service monitoring; currently, check if the service is alive 

#### `ServiceMonitor` class
  - `checkServiceAliveness` : accepts array of `IServiceAliveness` and 
   Creates an instance of ServiceMonitor.
     - services - An array of services to monitor.
     - interval - The interval, in minutes, between monitoring checks (default is 2 minutes)
     -  retries - The number of retries in case a service is determined to be down (default is 2 retries).
     - retryInterval - The interval, in seconds, between retries (default is 2 seconds).
   This class have three public functions 

    1. `monitorService` : Monitors the services at a regular interval and returns a function to exit and disconnect the monitoring.
    2. `pingService`: check the liveness of the services and disconnect 
    3. `disconnect`: disconnect the services that have disconnect handler

  

#### `ILivenessChecker` Interface
each instance of the classes that implements `ILivenessChecker` interface have its own logic to create the connection and 
check the service's aliveness 

#### Evenets 
we expose an instance of `MonitorEventEmitter` that make emitting events more easier 

Screenshot from 2024-01-22 16-39-21

### Changes 

added new package `monitoring`

### Related Issues

- #1469

### Checklist

- [ ] Tests included
- [x] Build pass
- [x] Documentation


MohamedElmdary:
```ts
export abstract class MonitorBase {
  constructor(protected readonly serviceName: string) {}

  public abstract isAlive(): Promise<ServiceStatus>;
}
```

Why don't we use `MonitorBase` abstract class and use it everyWhere instead of using `serviceName` in interface?

packages/monitoring/package.json
```diff
@@ -0,0 +1,26 @@
+{
+  "name": "monitoring",
+  "version": "1.0.0",
+  "description": "Threefold monitoring package",
+  "main": "dist/index.js",
+  "types": "dist/index.d.ts",
+  "scripts": {
+    "example": "yarn ts-node ./example/index.js",
```
MohamedElmdary, line 8:
```suggestion
    "example": "yarn ts-node ./example/index.ts",
```

Example should be executed directly from typescript code

packages/monitoring/src/types/index.ts
```diff
@@ -0,0 +1,28 @@
...
+  Black = "\x1b[30m",
+  Red = "\x1b[31m",
+  Green = "\x1b[32m",
+  Yellow = "\x1b[33m",
+  Blue = "\x1b[34m",
+  Magenta = "\x1b[35m",
+  Cyan = "\x1b[36m",
+  White = "\x1b[37m",
+  Gray = "\x1b[90m",
+}
```
MohamedElmdary, line 28: Checkout https://www.npmjs.com/package/chalk
Chalk is doing the same functionality without any configuration

packages/monitoring/src/serviceMonitor/graphql.ts
```diff
@@ -0,0 +1,23 @@
...
+      return {
+        alive: true,
+      };
+    } catch (error) {
+      return {
+        alive: false,
+        error,
+      };
+    }
+  }
```
MohamedElmdary, line 22:
```ts
// utils.ts
// this is a util method which handle returned status
// so we can update how return status look from a single method
export async function resolveServiceStatus(promise: Promise<any>) {
  try {
    await promise;
    return { alive: true };
  } catch (error) {
    return { alive: false, error };
  }
}
```

```suggestion
  isAlive(): Promise<ServiceStatus> {
     return resolveServiceStatus(sendGetRequest(this.ServiceURL, "", {}))
   }
```

packages/monitoring/src/helpers/utils.ts
```diff
@@ -0,0 +1,37 @@
+import { RequestError } from "@threefold/types";
+import axios, { AxiosError } from "axios";
+
+import { TextColor } from "../types";
+
+export async function sendGetRequest(url: string, body: string, headers: Record<string, string>) {
```
MohamedElmdary, line 6: Get request doesn't have body ... also we don't need any headers here for now so we can mark headers as optional

packages/monitoring/src/types/index.ts
0oM4R: awesome , thanks

packages/monitoring/src/helpers/utils.ts
```diff
@@ -0,0 +1,40 @@
...
+    const { response } = e as AxiosError;
+    let errorMessage = (response?.data as { error: string })?.error;
+    if (!errorMessage) {
+      errorMessage = (e as AxiosError).message;
+    }
+    throw new RequestError(`HTTP request failed ${errorMessage ? "due to " + errorMessage : ""}.`);
+  }
+}
+
+export function generateString(length: number): string {
```
xmonader, line 23: Why didn't you include tests for this?

packages/monitoring/src/helpers/utils.ts
```diff
@@ -0,0 +1,40 @@
...
+import { ServiceStatus } from "src/types";
+
+export async function sendGetRequest(url: string, headers?: Record<string, string>) {
+  const options = {
+    method: "get",
+    url: url,
+    headers: headers || {},
+  };
+  try {
+    return await axios(options);
```
xmonader, line 12: why not axios.get directly?

packages/monitoring/src/helpers/utils.ts
```diff
@@ -0,0 +1,40 @@
+import { RequestError } from "@threefold/types";
+import axios, { AxiosError } from "axios";
+import { ServiceStatus } from "src/types";
+
+export async function sendGetRequest(url: string, headers?: Record<string, string>) {
```
xmonader, line 5: why didn't you default the headers to {} ?

packages/monitoring/src/helpers/utils.ts
```diff
@@ -0,0 +1,40 @@
...
+  const options = {
+    method: "get",
+    url: url,
+    headers: headers || {},
+  };
+  try {
+    return await axios(options);
+  } catch (e) {
+    const { response } = e as AxiosError;
+    let errorMessage = (response?.data as { error: string })?.error;
```
xmonader, line 15: this should have been more concise e.response?.data.error || e.message I believe

packages/monitoring/src/helpers/events.ts
```diff
@@ -0,0 +1,37 @@
+import chalk from "chalk";
+import { EventEmitter } from "events";
+
+const events = new EventEmitter();
+type ServiceStatus = { [key: string]: boolean };
+
+const serviceStatusSummary: ServiceStatus = {};
+const ALIVE = chalk.green.bold("Alive");
+const DOWN = chalk.red.bold("Down");
+function logsHandler(msg) {
```
xmonader, line 10: won't this duplicate the logs coming from the grid client? I thought we were going to create another sink

packages/monitoring/src/helpers/events.ts
```diff
@@ -0,0 +1,37 @@
...
+function printStatusSummary() {
+  const serviceNames = Object.keys(serviceStatusSummary);
+  const maxServiceNameLength = Math.max(...serviceNames.map(entry => entry.length));
+  console.log(chalk.blue.bold("Aliveness check summary:"));
+  for (const service in serviceStatusSummary) {
+    const padding = " ".repeat(maxServiceNameLength - service.length);
+    console.log(`\t${service}${padding}: ${serviceStatusSummary[service] ? ALIVE : DOWN}`);
+  }
+}
+events.addListener("logs", logsHandler);
```
xmonader, line 33: and in the dashboard package ppl should listen to these events to provide custom callbacks in the UI?

packages/monitoring/src/serviceMonitor/disconnectHandler.ts
```diff
@@ -0,0 +1,11 @@
+import { events } from "../helpers/events";
+import { RMBMonitor, TFChainMonitor } from "../index";
+import { IServiceAliveness } from "../types";
+export async function disconnectServices(services: IServiceAliveness[]): Promise<void> {
+  for (const service of services) {
+    if (service instanceof RMBMonitor || service instanceof TFChainMonitor) {
```
xmonader, line 6: The check should be if the service implements the IDisconnector

packages/monitoring/src/serviceMonitor/alivenessChecker.ts
```diff
@@ -0,0 +1,56 @@
...
+ * @throws Will throw an error if the provided services array is empty.
+ *
+ * @returns An object with a method `exitAndDisconnect` that can be used to stop the monitoring and disconnect services.
+ *
+ */
+export function monitorServiceAliveness(
+  services: IServiceAliveness[],
+  interval = 2,
+): { exitAndDisconnect: () => Promise<void> } {
+  if (!services.length) throw new Error("No services to monitor");
```
xmonader, line 20: be careful with your checks, do services.length === 0, while your code works, still better to be as explicit as possible

packages/monitoring/src/serviceMonitor/alivenessChecker.ts
```diff
@@ -0,0 +1,56 @@
+import { events } from "../helpers/events";
```
xmonader, line 1: I don't uderstand why we didn't do it like this


class ServiceMonitor {

    constructor(services: Array<IService>, interval) {
        ...
    }
    private checkLivenessOnce {

    }
    private checkHealthOnce {

    }
    monitor() {
        withInterval do your checkLivenessOnce and checkHealthOnce 
    }

}

packages/monitoring/src/helpers/events.ts
0oM4R: will add prefix to the event names

packages/monitoring/example/index.ts
```diff
@@ -0,0 +1,25 @@
+import { GridProxyMonitor, RMBMonitor, ServiceMonitor, TFChainMonitor } from "../src/";
+async function HealthCheck() {
+  try {
+    const services = [
+      new GridProxyMonitor("<FakeURL>"),
+      new TFChainMonitor("wss://tfchain.dev.grid.tf/ws", "mnemonic", "sr25519"),
+      new RMBMonitor("wss://relay.dev.grid.tf", "wss://tfchain.dev.grid.tf/ws", "mnemonic", "sr25519"),
+    ];
```
mohamedamer453, line 8: Would be great if a check for graphql was added to the example.

mohamedamer453 (approved):

packages/monitoring/src/serviceMonitor/tfChain.ts
```diff
@@ -0,0 +1,39 @@
...
+
+import { IDisconnectHandler, ILivenessChecker, ServiceStatus } from "../types";
+
+export class TFChainMonitor implements ILivenessChecker, IDisconnectHandler {
+  private name = "TFChain";
+  private url: string;
+  private tfClient: TFClient;
+  constructor(tfChainUrl: string, mnemonic: string, keypairType: KeypairType) {
+    this.url = tfChainUrl;
+    this.tfClient = new TFClient({ url: this.url, mnemonicOrSecret: mnemonic, keypairType: keypairType });
```
AhmedHanafy725, line 12: why are we using TFClient here? it requires a mnemonic and a keypair type. Instead, we can use the queryClient which requires only the url

AhmedHanafy725 (changes requested):
I am afraid about disconnecting the clients cause it's only one static instance per url on the client, so if it will be used in the dashboard it might disconnect the client while deploying something.

packages/monitoring/package.json
```diff
@@ -0,0 +1,32 @@
+{
+  "name": "monitoring",
```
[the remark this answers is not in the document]
0oM4R: i notice the license also is missing, which license should we use ?

Issue #1855, opened 2024-01-03
Adding and Removing public config take long time to reflect


A-Harby:
### Description

It take a very long time to add or remove a public configuration.
When you click remove and the tab loads, then the config input disappears, but when you reopen it, it still appears, and you have to wait a long time and a refresh for it to disappear. The same happens in the add.

### Logs/Alerts

[Threefold Dashboard: Your Farms.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/dbc37d24-56a9-4d03-a601-cfc2a31da39a)

Mahmoud-Emad:
## Issue update 

### Investigation and Solution:

- Created the config object and set the value of the node public config inside it
- Send the config object to the node as an event to update the UI
- Created a separate function for toggling adding/removing the config to make it readable and easier to code.

### Work Completed:

- The bug should be fixed at https://github.com/threefoldtech/tfgrid-sdk-ts/pull/2052

A-Harby:
- Now it sometimes doesn't let you save a configuration, even if the value is valid but the save button is disabled.

[screen-capture (3).webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/c44f61fe-81f1-4c2a-9d3e-02bb01b9d042)

- And when I change the values, despite closing the dialogue without saving the value, it's saved, and I can do it more often again. The only way is to reload the page.

[screen-capture (4).webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/c8103429-784c-4cca-9473-22e36064f284)

- And as you can see in the video, some of the values aren't valid, but the save button is still enabled (validation isn't rechecked), which will give you an error after the save button is pressed.

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/ec642367-6e0b-42c5-8eca-fdd733152fcc)

Mahmoud-Emad:
## Issue Update

### Work in Progress

Currently addressing the issue related to transitioning from the current public configuration `<Current state>` to the node configuration `<Default state>`.

For context, when the component is mounted, the node's public configuration is initially set to `null` or remains unset. This situation results in undesired behavior, as demonstrated below in a Vue.js template:

```vue
<template>
  <v-text-field v-model="config.ip4" /> <!-- User's input changes both the form and the node config -->
</template>

<script lang="ts">
[...]

PR #2052, opened 2024-01-21
Fix: Adding and Removing public config take long time to reflect.

Mahmoud-Emad:
### Description

Replaced the binding on the 'modelValue' with the initialized 'config' object.

### Changes
- Created the config object and set the value of the node public config inside it
- Send the config object to the node as an event to update the UI
- Created a separate function for toggling adding/removing the config to make it readable and easier to code.

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1855

### Screenshots/Video

  [Screencast from 01-21-2024 03:47:00 PM.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/57001890/32b2c72c-0dcd-4c01-a7dc-5e788ae65570)

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/packages/playground/src/dashboard/components/public_config.vue
+++ b/packages/playground/src/dashboard/components/public_config.vue
@@ -146,14 +146,12 @@
 </template>
 
 <script lang="ts">
-import type { GridNode, PublicConfig } from "@threefold/gridproxy_client";
+import type { PublicConfig } from "@threefold/grid_client";
 import _ from "lodash";
 import { onMounted, type PropType, ref, watch } from "vue";
 
-import { gridProxyClient } from "@/clients";
 import type { RuleReturn } from "@/components/input_validator.vue";
 import { useFormRef } from "@/hooks/form_validator";
-import type { IPublicConfig } from "@/utils/types";
 
 import { useGrid } from "../../stores";
 import { createCustomToast, ToastType } from "../../utils/custom_toast";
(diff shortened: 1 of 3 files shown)


AlaaElattar:
[Screencast from 01-22-2024 04:01:19 PM.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/54944491/fbfb0fd4-c55f-4788-885a-d89fe697337e)

mohamedamer453 (commented):
- After clearing the public config then instantly switching the page and returning to it the config will stay the same and won't be removed.

   [Screencast from 22-01-24 16:06:03.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/101194226/b60d77ad-bbfa-4d1f-8cb0-8ae648da7b2b)

- The same flow mentioned above is also reversible and can be done with adding public config.

   [Screencast from 22-01-24 16:08:26.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/101194226/a940bae7-2125-4b88-9178-73bb10cd9f8b)

Mahmoud-Emad:
> * After clearing the public config then instantly switching the page and returning to it the config will stay the same and won't be removed.
>   
>     
>       
>       
>   
>       Screencast.from.22-01-24.16.06.03.webm
> * The same flow mentioned above is also reversible and can be done with adding public config.
>   
>     
>       
>       
>   
>       Screencast.from.22-01-24.16.08.26.webm

That happened because the changes reflected in the chain in time but not in the gql, so I think using the chain directly will be better to fix this issue

mohamedamer453 (approved):

AlaaElattar (approved):

Issue #2053, opened 2024-01-21
Clicking on the Grace Period Contract doesn't show the locked TFT anymore


A-Harby:
### Description

Devnet, 5b50681.

When clicking on the grace period contract, it doesn't show me the locked TFT like before; it's just stuck doing nothing.

[screen-capture.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/62727df4-33e4-4f7c-bab0-9f24eeb8002e)

### Steps to reproduce

1. Navigate to the dashboard.
2. Connect your wallet (Twin with a grace period contract).
3. Navigate to your contract.
4. Click on the Grace Period Contract.

AlaaElattar:
### Work in Progress (WIP):
- found the issue and fixing it.

A-Harby:
Verified, Devnet 2311246.

The grace period dialogue is added back with a clarifying message and amount locked.

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/aa687e8d-12eb-4926-86de-fff02e53f110)

[TC2618 - Grace period-locked TFT amount](https://app.testlodge.com/a/26076/projects/40893/suites/234374?expand_section=430129#case_4214522)

ramezsaeed:
Verified:
- Devnet
- 48a660d

PR #2076, opened 2024-01-29
Add Dialog to show contract lock details for contracts on grace period

AlaaElattar:
### Description

When clicking on the grace period contract, it doesn't show me the locked TFT like before; it's just stuck doing nothing.

### Changes
- Dialog added to show locked tfts.

### Related Issues

#2053 

[Screencast from 01-29-2024 11:18:56 AM.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/54944491/fb0cc253-ba77-4f2c-bd91-71b86f614fa5)

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/packages/playground/src/components/contracts_list/contracts_table.vue
+++ b/packages/playground/src/components/contracts_list/contracts_table.vue
@@ -195,7 +195,7 @@ const getAmountLocked = (): number => {
 };
 
 const isNodeInRentContracts = computed(() => {
-  if (props.contractsType == "rent") {
+  if (props.contractsType == ContractType.RENT) {
     const nodeIds = contracts.value.map(contract => contract.nodeId).filter(nodeId => nodeId !== undefined) as number[];
     if (contractLocked.value && contractLocked.value.amountLocked === 0) {
       return nodeIds.includes(selectedItem.value.nodeId);


packages/playground/src/components/contracts_list/contracts_table.vue
```diff
@@ -89,6 +89,27 @@
     </template>
   </weblet-layout>
 
+  <v-dialog width="800" v-model="contractStateDialog">
+    <v-card>
+      <v-card-title class="text-h5 mt-2"> Contract lock Details</v-card-title>
+      <v-card-text>
+        <v-row class="d-flex justify-center"> Amount Locked {{ contractLocked?.amountLocked }} TFTs. </v-row>
```
mohamedamer453, line 96: Maybe it would look better if the amount was in a different color/style from the text next to it.

image

Not sure if it's the best option with balance but maybe it should be rounded to 4 decimals.

image

AlaaElattar: should be fixed

AlaaElattar:
- The reason for locked tfts to be zero, that's deployed on dediated node that's also in grace period. That's why the locked tfts are zero and it has no billing rate. 

image

packages/playground/src/components/contracts_list/contracts_table.vue
```diff
@@ -89,6 +89,38 @@
...
+    <v-card>
+      <v-card-title class="text-h5 mt-2"> Contract lock Details</v-card-title>
+      <v-card-text>
+        <v-row class="d-flex justify-center">
+          Amount Locked:
+          {{ contractLocked?.amountLocked ?? 0 > 0 ? contractLocked?.amountLocked.toFixed(3) : 0 }}
+          TFTs.
+        </v-row>
+
+        <v-alert v-if="contractLocked?.amountLocked == 0" class="ma-4" type="warning" variant="tonal"
```
AhmedHanafy725, line 102: can we be more specific with this condition? we can check if the user has a rent contract on this node or not

AlaaElattar: should be fixed.

packages/playground/src/components/contracts_list/contracts_table.vue
```diff
@@ -155,6 +190,22 @@ const props = defineProps({
...
+const rentContracts = ref<NormalizedContract[]>([]);
+
+onMounted(() => {
+  rentContracts.value = props.contracts.value.filter(contract => contract.type === ContractType.RENT);
+  console.log("Rent Contracts: ", rentContracts);
+});
+
+const isNodeInRentContracts = computed(() => {
+  if (contractLocked.value && contractLocked.value.amountLocked === 0) {
+    const nodeIdsInRentContracts = rentContracts.value.map(contract => contract.nodeId);
```
AhmedHanafy725, line 202: why do we create this list of the rent contracts each time we click on a contract?
can't we just do it once after loading the contracts?

AhmedHanafy725 (commented):
please remove the debug logs

packages/playground/src/components/contracts_list/contracts_table.vue
AlaaElattar: fixed, I moved it to be calculated on mounting.

packages/playground/src/components/contracts_list/contracts_table.vue
```diff
@@ -89,6 +89,41 @@
     </template>
   </weblet-layout>
 
+  <v-dialog width="800" v-model="contractStateDialog">
+    <v-card>
+      <v-card-title class="text-h5 mt-2"> Contract lock Details</v-card-title>
+      <v-card-text>
+        <v-row class="d-flex justify-center">
+          Amount Locked:
+          {{ contractLocked?.amountLocked ?? 0 > 0 ? contractLocked?.amountLocked.toFixed(3) : 0 }}
```
Mahmoud-Emad, line 98: Please create a function that does this logic and add only the variable/function caller in the template

```suggestion
          {{ getAmountLocked() }}
          
  const getAmountLocked = (): number => {
    const amountLocked = contractLocked?.amountLocked ?? 0;
    return amountLocked > 0 ? parseFloat(amountLocked.toFixed(3)) : 0;
  };
```

packages/playground/src/components/contracts_list/contracts_table.vue
```diff
@@ -118,8 +153,8 @@ import { ContractStates, type GridClient } from "@threefold/grid_client";
 import type { NodeStatus } from "@threefold/gridproxy_client";
 import type { ContractLock } from "@threefold/tfchain_client";
 import { DeploymentKeyDeletionError, TFChainErrors } from "@threefold/types";
-import { defineComponent, type PropType, type Ref, ref } from "vue";
-import { capitalize } from "vue";
+import { computed, defineComponent, type PropType, type Ref, ref } from "vue";
+import { capitalize, onMounted } from "vue";
```
Mahmoud-Emad, line 157: Could be used in the same line
```suggestion
import { capitalize, onMounted, computed, defineComponent, type PropType, type Ref, ref } from "vue";
```

packages/playground/src/components/contracts_list/contracts_table.vue
```diff
@@ -155,6 +190,23 @@ const props = defineProps({
   },
 });
 
+const rentContracts = ref<NormalizedContract[]>([]);
+const nodeIdsInRentContracts = ref<number[]>([]);
+
+onMounted(() => {
+  rentContracts.value = props.contracts.value.filter(contract => contract.type === ContractType.RENT);
```
Mahmoud-Emad, line 197: We are already passing the contracts based on their types, namely [name, rent, node]. Instead of individually creating tables for each contract type using a `for loop`, you can directly render them using a single loop within the template. This eliminates the need for multiple instances of the `contract-table` component.

For example:

```vue
<template>
  <div>
    <contract-table v-for="(contractType, index) in contracts" :key="index" :contracts="contractType" />
  </div>
</template>

<script>
import { ref, onMounted } from 'vue';

export default {
  setup() {
    const contracts = ref([]);

    onMounted(() => {
      contracts.value = getContracts(); // Returns an array of contracts grouped by type [NameContract[], NodeContract[], RentContract[]]
    });

    return {
      contracts,
    };
  },
};
</script>

**Review by Mahmoud-Emad — CHANGES_REQUESTED** [review]
Good job :)

File: packages/playground/src/components/contracts_list/contracts_table.vue [hunk 7]
```diff
@@ -155,6 +189,21 @@ const props = defineProps({
   },
 });
 
+const getAmountLocked = (): number => {
+  const amountLocked = contractLocked?.value?.amountLocked ?? 0;
+  return amountLocked > 0 ? parseFloat(amountLocked.toFixed(3)) : 0;
+};
+
+const isNodeInRentContracts = computed(() => {
+  if (props.contractsType == "rent") {
```
Mahmoud-Emad, line 198: please use `ContractType.RENT` instead

Mahmoud-Emad (changes requested):

Mahmoud-Emad (approved):
Thanks Alaa, Good job :)

mohamedamer453 (approved):

Issue #1388, opened 2023-11-08
integrate perf tests results in the UI


xmonader:
TBD

zaelgohary:
WIP: 
Currently working on implementing getPerfTest in grid client. Once done, will apply it in Dashboard.

zaelgohary:
WIP:
Added perf tests to grid client. Currently working on displaying them in node details.

zaelgohary:
WIP: currently working on changing the iperf card style.

zaelgohary:
Removed to in progress because perf tests are not shown in case of grid is not defined.

A-Harby:
Devnet, 2311246.
On some nodes, the tests are repeated.
eg. Node ID 171.
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/b742439d-d3c6-4e7f-abdd-782d191803a7)

zaelgohary:
Work completed:

Added a check to slice the returned result if its length > 4.

zaelgohary:
The current behavior in zos returns tests done against gateway nodes. As a temporary fix, I sliced the array to return only results of one node. Opened a new [issue](https://github.com/threefoldtech/tfgrid-sdk-ts/issues/2231) so we can edit this once it's changed from zos side.

khaledyoussef24:
verified on devnet
version : 75e5a1b
![Screenshot 2024-02-26 at 11 51 01 AM](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/ac839668-3dda-433e-9f2d-04f599011d87)
![Screenshot 2024-02-26 at 11 51 10 AM](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/b80075fe-720c-4e9a-a975-de22fe2de749)
![Screenshot 2024-02-26 at 11 51 22 AM](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/ac1ddc1f-acf3-4ee9-8dca-a1d060d0976e)
![Screenshot 2024-02-26 at 11 51 28 AM](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/fa5f5158-26f3-48a0-a42c-59495d2bedbc)
[...]

PR #2097, opened 2024-01-31
Add node perf tests

zaelgohary:
### Description

- Add perf tests to grid client
- Add perf tests to node details (will be shown only if node is up)
- Edit formatResources to include smaller values such as MB & KB

image

### Changes

- Added perf tests in client and updated zos script

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1388

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation


xmonader:
Change IPerf test in the UI to Network Speed Test and, not sure why there is workloads field shown as part of the CPU Benchmarks

mohamedamer453 (commented):
The dashboard keeps loading and doesn't start, with the following error in the console.

image

```
models.js:2144 Uncaught ReferenceError: Cannot access 'CPUBenchmark' before initialization
    at models.js:2144:31
```

packages/playground/src/components/node_details.vue
```diff
@@ -70,6 +77,7 @@ import { type PropType, ref, watch } from "vue";
 import { useRoute } from "vue-router";
 
 import CountryDetailsCard from "@/components/node_details_cards/country_details_card.vue";
+import cpuBenchmarkCard from "@/components/node_details_cards/cpu_benchmark.vue";
```
Mahmoud-Emad, line 80:
```suggestion
import CPUBenchmarkCard from "@/components/node_details_cards/cpu_benchmark_card.vue";
```

packages/playground/src/components/node_details.vue
```diff
@@ -82,6 +90,7 @@ import type { FilterOptions } from "@/types";
 import { type GridProxyRequestConfig, nodeInitializer } from "@/types";
 import { getNode, getNodeStatusColor } from "@/utils/get_nodes";
 
+import IPerfCard from "./node_details_cards/iperf_details.vue";
```
Mahmoud-Emad, line 93:
```suggestion
import IPerfCard from "./node_details_cards/iperf_details_card.vue";
```

Mahmoud-Emad (changes requested):
Good job ya zuzu :), small comments :)

mohamedamer453 (commented):
Great work! everything is working fine but i noticed some weird behavior in the network speed. for all nodes the network speed is within `1023.xx KB` range.

image

this range is present for all nodes that are available on devnet, so i think there might be something wrong either with the received data or with the calculation.

packages/playground/src/components/node_details.vue
zaelgohary: I didn't name it that way because if renamed to this, it would be used in the template as <c-p-u-benchmark />

zaelgohary:
> Great work! everything is working fine but i noticed some weird behavior in the network speed. for all nodes the network speed is within `1023.xx KB` range.
> 
> image
> 
> this range is present for all nodes that are available on devnet, so i think there might be something wrong either with the received data or with the calculation.

I double-checked and the data is fine. Each node gets tested against nodes 11 & 14 and I'm showing the results of the first. I added more details to the card so it would be more clear.

mohamedamer453 (approved):

Mahmoud-Emad (approved):

Issue #2112, opened 2024-02-01
All routes are broken on development and staging


Mahmoud-Emad:
### Description

All routes are broken on development and staging

Must include any relevant identifiers like:

- **Dev & Staging**:

### Steps to reproduce

- Go to the dashboard
- Go to deploy tab then applications
- Click on any application

### Screenshots

- ![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/57001890/f43dac5d-5233-47c3-9f19-ebf6ff371678)

Mahmoud-Emad:
## Issue Update

### Work Completed

- Updated the router, adding the 'solutions/' route.
- Implemented an enum (ApplicationRoute) to assign variables to all routes, improving code quality.

### Work in Progress (WIP)

None.

### Investigation and Solution

- Investigated the need for a router update and identified the 'solutions/' route addition.
- Implemented an enum to enhance code reliability and eliminate typos in the hard-coded approach

khaledyoussef24:
verified routes are fixed and reached all solutions and applications



![Image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/16346046-25a7-45f1-9c75-23097c82bb5e)
![Image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/6a3a05f6-4330-447a-9f9d-8ae7b2cb8aec)
![Image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/b11b502d-328c-4910-9cd5-79314d63c41b)
![Image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/7cab676e-ff12-4955-8938-89a2cf6a213e)
![Image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/06c704a7-0f8e-4532-9b05-f51cdb0da10b)
![Image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/3ae188d9-eda5-40ec-a7a6-32f28186008c)
![Image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/f38395a4-d51e-47cd-94b8-2d6632fe4943)

PR #2115, opened 2024-02-01
Fix: All routes are broken on development and staging.

Mahmoud-Emad:
### Description

- Updated the router, adding the 'solutions/' route.
- Implemented an enum (ApplicationRoute) to assign variables to all routes, improving code quality.

### Changes

- Updated the router, adding the 'solutions/' route.
- Implemented an enum (ApplicationRoute) to assign variables to all routes, improving code quality.

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/2112

### Screenshots

- image

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/packages/playground/src/router/index.ts
+++ b/packages/playground/src/router/index.ts
@@ -183,7 +183,7 @@ const router = createRouter({
             },
 
             {
-              path: "peertube",
+              path: "/solutions/peertube",
               component: () => import("../views/peertube_view.vue"),
               meta: {
                 title: "Peertube",
@@ -205,7 +205,7 @@ const router = createRouter({
               },
             },
             {
-              path: "funkwhale",
+              path: "/solutions/funkwhale",
               component: () => import("../views/funkwhale_view.vue"),
               meta: {
                 title: "Funkwhale",
@@ -227,7 +227,7 @@ const router = createRouter({
               },
             },
             {
-              path: "mattermost",
+              path: "/solutions/mattermost",
               component: () => import("../views/mattermost_view.vue"),
               meta: {
                 title: "Mattermost",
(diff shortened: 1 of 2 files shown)


packages/playground/src/router/index.ts
```diff
@@ -491,7 +509,7 @@ const router = createRouter({
               },
             },
             {
-              path: "freeflow",
+              path: ApplicationRoute.freeflow,
```
mohamedamer453, line 512: i believe this route should be commented since the freeflow solution is currently hidden from the applications but with this present in the routes it can be accessed.

image

Mahmoud-Emad: Yeah indeed but from where do you know that we have a freeflow solution? but yeah you are right I'll comment it out good catch.

mohamedamer453 (approved):

MohamedElmdary:
Why do we need `ApplicationRoute` enum? that's not the issue we only need to update the path in  solution_view page

packages/playground/src/views/solutions_view.vue
```diff
@@ -2,7 +2,7 @@
   <view-layout>
     <v-row>
       <v-col sm="12" md="6" lg="4" v-for="card in cards" :key="card.title">
-        <router-link :to="'/solutions' + card.route">
+        <router-link :to="card.route">
```
MohamedElmdary, line 5:
```suggestion
        <router-link :to="'/deploy/applications' + card.route">
```

this should fix the issue without introducing any new line of code which add up complexity for no reason

Mahmoud-Emad: You have a point, actually I did what you meant in the first commit, then I noticed that we push the route to the router when clicking on the solution card, we added it hard-coded, so I wonder if we faced the same issue by mistake of someone, I decided that the enum would be a good idea to enhance our readability, what do you think?

MohamedElmdary (approved):

Issue #1981, opened 2024-01-15
Grid Capacity cards on the homepage 


A-Harby:
### Description

https://dashboard.bknd1.ninja.tf/#/, 2.3.0-alpha6

Where is this data coming from, and how is it calculated?
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/c70c9424-699e-43ea-95fd-d82aafd2a95c)

And why is there a difference between it and the stats site?
If it is data from all the networks, the data should be bigger than the statistics site, which only has 3 networks.
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/a183915d-d802-4e3a-809f-fe23d26fb0ec)

And if it is only for each network, why is that not the same as the stats page on the same network?
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/defe353c-7940-40c8-98da-69250db68c67)

A-Harby:
The stats page [issue](https://github.com/threefoldtech/tfgrid-sdk-ts/issues/2002) was fixed and updated, and it's calculated correctly.
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/40c5a949-b90b-426e-863f-be7e8b755206)



But still, capacity, nodes, and cores are showing varying numbers.
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/e4cfab0a-9951-48c9-b429-e82fb1bd7216)
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/1c6f55b4-d67f-4a78-863f-1f4ea7b08d77)

0oM4R:
@samaradel  is the home page comes from threefold.io ? 
if yes, is those card represents all networks, or just main net ?

samaradel:
> @samaradel  is the home page comes from threefold.io ? 
> if yes, is those card represents all networks, or just main net ?

Yes, all envs

0oM4R:
actually the cards data is hard coded on dashboard 
https://github.com/threefoldtech/tfgrid-sdk-ts/blob/e0fb9c4024a4a4c6f628cf8946204d4a16170769/packages/playground/src/components/logged_in_landing.vue#L156-L177


i suggest to fetch only the current net, stats

0oM4R:
### Investigation and Solution:
as we use those stats in two pages, we have to store the data.
[...]

PR #2123, opened 2024-02-05
add stats summary endpoint

0oM4R:
### Description

Using nginx javascript we can run a script whenever the user hit the target endpoint, the mentioned endpoint is `/api/stats-summary`, this will reduce code replication and make data consistent over different websites; threefold.io, dashboard 
also we are caching the data for one day, if any errors happen while updating it, we are returning the outdated data with flag `outdated` if there is no cached data too or if any error happen while reading it, we return a dummy data,
### Changes

- DockerFile: 
   - update the nginx version: njs require the latest nginx image
   - copy nginx files to the image
- Nginx : 
   - in nginx.conf I just loaded the needed js module and  set the resolver:
     the actual changes on the default nginx.conf is :
      ```
     load_module modules/ngx_http_js_module.so;
      js_path "/etc/nginx/njs/";
  
      resolver 8.8.8.8;
      js_import main from stats.js;
     ```
   - add `/summary` endpoint to server in `default.conf
   - add njs script

### Related Issues

 - #1981 

### Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation

### DIFF
--- a/packages/new_stats/Dockerfile
+++ b/packages/new_stats/Dockerfile
@@ -11,4 +11,3 @@ RUN yarn lerna run build --no-private && yarn workspace @threefold/newstats buil
 FROM nginx:alpine
 COPY --from=build /app/packages/new_stats/dist /usr/share/nginx/html
 COPY --from=build /app/packages/new_stats/nginx /etc/nginx
-COPY --from=build /app/packages/new_stats/tmp /tmp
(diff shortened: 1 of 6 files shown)


0oM4R:
the verify option is disabled here as we <currently> don't have ssl cert for the image;
Screenshot from 2024-02-05 12-41-06

 https://github.com/threefoldtech/tfgrid-sdk-ts/blob/0e5faac8d2d3c49ffcf907240ed0905ae486791c/packages/new_stats/nginx/njs/stats.js#L15

MohamedElmdary:
Do we need to implement retries 3 times and/or return back default values if it failed? @AhmedHanafy725

MohamedElmdary:
1. [ ] Let's invest if this script is used as server or just fire whenever we call route ... in case it works as server we can implement a very simple in-memory cache (last for 1 day)

2. [ ] Add 3 reties for request (better to call the failed urls only `if possible`)

AhmedHanafy725 (changes requested):

AhmedHanafy725 (approved):

MohamedElmdary (approved):

Issue #1790, opened 2023-12-25
Farm name is case sensitive, which allow to create two farms with the same name


ramezsaeed:
Create farm in your farms tab is case senestive;
which means the one user can crerate mutltiple farms with the same name !
and when come to search by name (its not case sensitive)

rtest1
Rtest1
rTest1

all these farms have the same name but different in upper/lower cases

![Screenshot from 2023-12-25 14-19-03](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/16083592/fe09b4b3-7d63-4b2c-aae6-50dd5ea010bd)

xmonader:
This is allowed on the chain @sameh-farouk

sameh-farouk:
> This is allowed on the chain @sameh-farouk

Yes!

https://github.com/threefoldtech/tfchain/blob/36c891c02a1758a6bd6769357fdef230886940ff/substrate-node/pallets/pallet-tfgrid/src/farm.rs#L31-L34

https://github.com/threefoldtech/tfchain/blob/36c891c02a1758a6bd6769357fdef230886940ff/substrate-node/pallets/pallet-tfgrid/src/farm.rs#L55

xmonader:
can be limited from the UI for now (if the user doesn't have multiple farms with different cases)

khaledyoussef24:
verified on devnet
version : 
the user can not create farms with the same name using any combination of lower and upper cases.


![Image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/7b651b41-836f-4fcf-8e6d-e7d6841ba11a)
![Image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/1a6e232f-359f-480e-9ef6-5b0bde8e5db2)

https://app.testlodge.com/a/26076/projects/40893/suites/234374?expand_section=426668#case_4150450

PR #2156, opened 2024-02-13
Disable creating 2 farms with same name

AlaaElattar:
### Description

Create farm in your farms tab is case senestive;
which means the one user can crerate mutltiple farms with the same name !
and when come to search by name (its not case sensitive)

### Changes

- added function returns all farms names', then check the entered name in lower case doesn't exist.
### Related Issues

https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1790

[Screencast from 02-13-2024 11:28:35 AM.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/54944491/43d4701f-b80b-4bcc-8c01-62de2478b282)

### Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation

### DIFF
--- a/packages/playground/src/components/manage_gateway_dialog.vue
+++ b/packages/playground/src/components/manage_gateway_dialog.vue
@@ -227,6 +227,7 @@
 </template>
 
 <script lang="ts">
+import { FailedToFreeIPs } from "@threefold/types/dist/errors/tfchain/smart_contract";
 import { onMounted, type PropType, ref } from "vue";
 
 import { useProfileManager } from "../stores";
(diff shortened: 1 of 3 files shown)


0oM4R (commented):
I suggest to move this step to be a validation function, we already have the farm names and there is no requests to wait for, also the error is created by our side, I believe it will be great to make it validator role, what do you think ?

AlaaElattar:
> I suggest to move this step to be a validation function, we already have the farm names and there is no requests to wait for, also the error is created by our side, I believe it will be great to make it validator role, what do you think ?

should be done

Screenshot from 2024-02-14 12-05-29

0oM4R (approved):
Thanks ya Alaa great job!

Issue #2117, opened 2024-02-02
Add searchbar in the solutions component


xmonader:
To have a generic search functionality to filter out solutions

zaelgohary:
Work completed: 

Added search input in applications. Now the cards list will be filtered based on the search term.

khaledyoussef24:
version : 385928e
network : devnet
search in now added and functional 

![Screenshot 2024-03-25 at 10 41 32 AM (2)](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/3012cf6c-7a54-4b45-9eb2-4db46d520ecb)
![Screenshot 2024-03-25 at 10 41 39 AM (2)](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/673a7f0c-ec77-494d-94ad-ff5e47278b7f)
![Screenshot 2024-03-25 at 10 41 44 AM (2)](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/e9053824-42a9-47da-b578-d82721fe1183)
![Screenshot 2024-03-25 at 10 41 52 AM](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/36f1f2ee-8227-446c-891d-3d642c246fae)
https://app.testlodge.com/a/26076/projects/40893/suites/234374?expand_section=422332#case_4246719

ramezsaeed:
Verified:
- Devnet
- d2d2622

![Screenshot from 2024-04-01 00-55-18](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/16083592/af0490f3-43bb-410f-80ea-fd23ae48ab73)

PR #2190, opened 2024-02-18
Add search in solutions

zaelgohary:
### Description

- Add search in Application page

### Changes

- Filter applications based on the search item

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/2117

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation


AlaaElattar:
- when i searched then clicked, it didn't open directly.

[Screencast from 02-18-2024 02:16:02 PM.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/54944491/27af54da-02d1-472e-9c99-fb833b2ba47b)

MohamedElmdary:
> * when i searched then clicked, it didn't open directly.
> 
>  Screencast.from.02-18-2024.02.16.02.PM.webm

I don't think we can use combox because it's default searching will be remove after any `click` event no matter we stop propagation or not because it's not about `click` but `blur` event

So I'd recommend using `v-text-frield` instead as we don't need to select from list but search with some text

packages/playground/src/views/solutions_view.vue
```diff
@@ -1,12 +1,6 @@
 <template>
   <view-layout>
-    <v-autocomplete
-      label="Search Applications"
-      v-model="searchItem"
-      :items="titles"
-      class="mb-5"
-      @update:search="filterCards"
-    ></v-autocomplete>
+    <v-text-field label="Search Applications" v-model="searchItem" class="mb-5" @input="filterCards"></v-text-field>
```
MohamedElmdary, line 3: I think it should be clearable

zaelgohary: what do you mean?

MohamedElmdary: clearable prop in vuetify `v-text-field`

AlaaElattar (approved):

packages/playground/src/views/solutions_view.vue
zaelgohary: Added.

packages/playground/src/views/solutions_view.vue
```diff
@@ -1,7 +1,14 @@
...
+    <v-text-field
+      label="Search Applications"
+      v-model="searchItem"
+      class="mb-5"
+      @input="filterCards"
+      clearable
+    ></v-text-field>
     <v-row>
-      <v-col sm="12" md="6" lg="4" v-for="card in cards" :key="card.title">
+      <v-col sm="12" md="6" lg="4" v-for="card in filteredCards" :key="card.title">
```
MohamedElmdary, line 11: here we can update filter directly so we can eliminate the entire logic

packages/playground/src/views/solutions_view.vue
```diff
@@ -161,11 +170,25 @@ export default {
         route: DashboardRoutes.Applications.Wordpress,
       },
     ];
+    const filteredCards = ref(cards);
+    const titles = cards.map(({ title }) => ({
+      title,
+    }));
```
MohamedElmdary, line 176: If we applied the above suggestion we can remove this code
```suggestion
```

packages/playground/src/views/solutions_view.vue
```diff
@@ -161,11 +170,25 @@ export default {
...
     const baseURL = import.meta.env.BASE_URL;
+    const searchItem = ref("");
+    function filterCards() {
+      filteredCards.value = cards.filter(n =>
+        n.title.toLocaleLowerCase().includes(searchItem.value.toLocaleLowerCase()),
+      );
+    }
```
MohamedElmdary, line 183: If we applied the above suggestion we can remove this code
```suggestion
```

MohamedElmdary:
Also if a user wrote a search query which doesn't  match any application we should show a message with some text message (e.g `No solution was found with the provided search query`)

packages/playground/src/views/solutions_view.vue
```diff
@@ -1,7 +1,14 @@
 <template>
   <view-layout>
+    <v-text-field
+      label="Search Applications"
+      v-model="searchItem"
+      class="mb-5"
+      @input="filterCards"
+      clearable
+    ></v-text-field>
```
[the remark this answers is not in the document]
zaelgohary: Replaced function logic w computed variable.

MohamedElmdary (approved):

Issue #2179, opened 2024-02-15
Update new stats chart to allow CORS


0oM4R:
### Description

accessing https://stats.dev.grid.tf/api/stats-summary from playground throws cors errors 

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/62248851/d412c778-def7-46b5-ab40-573d18b80e7f)

need to update ingress values to allow get method from all origins

PeterNashaat:
- you can update the Nginx Ingress template to include CORS-related annotations. This will allow requests from all origins. Here's an example of how you might configure it : 
```
  annotations:
    nginx.ingress.kubernetes.io/enable-cors: "true"
    nginx.ingress.kubernetes.io/cors-allow-methods: "GET, OPTIONS"
    nginx.ingress.kubernetes.io/cors-allow-origin: "*"
    nginx.ingress.kubernetes.io/cors-allow-headers: "DNT,User-Agent,X-Requested-With,If-Modified-Since,Cache-Control,Content-Type,Authorization"
s
```
- or it can be 
```
    nginx.ingress.kubernetes.io/cors-allow-methods: "PUT, GET, POST, OPTIONS"
```

PeterNashaat:
- [Cors-allow-origin](http://nginx.ingress.kubernetes.io/cors-allow-origin) by default eq "*"

MohamedElmdary:
@PeterNashaat 

I tried to request /state-summary but didn't get back `Access-Control-Allow-Origin` header back in response header
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/31689104/74e6bc20-4445-456a-8a1f-21a12b631432)

A-Harby:
Verified, Devnet.
The errors is not shown on the login page or stats page.
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/2d29284f-45b8-40d1-8826-3b5dcde196ad)

PR #2201, opened 2024-02-19
Use stats summary on grid capacity cards

0oM4R:
### Description

use stats summary endpoint to fetch the landing page cards data, those data will be stored in stats store to access them from different pages 

### Changes

  - use stats-summary endpoint to fetch grid stats data
  - add stats store to store stats summary
  - make explore grid redirect to stats website
  - add stats url to build-env script and configs, also added to
  - enable CORS from all origins on stats summary endpoint

### Related Issues
- #1981 
- #2179 

### Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation

### DIFF
--- a/packages/playground/src/stores/stats.ts
+++ b/packages/playground/src/stores/stats.ts
@@ -2,7 +2,7 @@ import { defineStore } from "pinia";
 import { computed } from "vue";
 
 import { useAsync } from "@/hooks";
-const url = window.env.STATS_URL || "https://stats.grid.tf/";
+const url = window.env.STATS_URL || "https://stats.grid.tf";
 
 export const useRequestStore = defineStore("request-store", () => {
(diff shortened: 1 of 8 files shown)


packages/playground/src/stores/stats.ts
```diff
@@ -0,0 +1,13 @@
...
+import { useAsync } from "@/hooks";
+const url = window.env.STATS_URL || "https://stats.grid.tf";
+
+export const useRequestStore = defineStore("request-store", () => {
+  const res = useAsync(() => fetch(url + "/api/stats-summary").then(resp => resp.json()), {
+    init: true,
+  });
+  const data = computed(() => res.value.data);
+  return { data };
+});
```
MohamedElmdary, line 13: `useRequestStore` is a very generic name shouldn't we call it something like `useStatsSummary`

also we are computing `stats` array in 2 places why don't we return it here directly?

AlaaElattar (approved):

Issue #1848, opened 2024-01-02
Trigger Minting hash validation before clicking view


zaelgohary:
The validation for minting should be initiated before clicking the view button.

Not only should the validation check the hash's length, but also if the hash is valid or not.

ramezsaeed:
look at the difference betweeen use the last char as "c" or "k"

![Screenshot from 2024-01-03 11-52-48](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/16083592/87e2712a-e4d3-4479-87ee-42c6c52e7810)
![Screenshot from 2024-01-03 11-52-32](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/16083592/da0f0445-ad13-4b99-a059-ffeb4aa180d1)
![Screenshot from 2024-01-03 11-52-26](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/16083592/dfeac06a-dcb1-454b-97fd-22bf319b76a4)
![Screenshot from 2024-01-03 11-52-13](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/16083592/b7aa31e7-97f1-40ec-a032-9ecb7d4f90df)

AlaaElattar:
### Work In Progress(WIP):
- Working on validating if the hash exists or not before clicking on view button.

PR #2207, opened 2024-02-19
Trigger Minting hash validation before clicking view 

AlaaElattar:
### Description

The validation for minting should be initiated before clicking the view button.

Not only should the validation check the hash's length, but also if the hash is valid or not.
### Changes

- Make all validations done before clicking view.
- Found that if hash valid will appear automatically, then there's no need for view button.
### Related Issues

#1848 

[Screencast from 02-19-2024 02:01:36 PM.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/54944491/2c401fe4-d4a0-4cc0-8bb0-944bed67d2bd)

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/packages/playground/src/views/minting_view.vue
+++ b/packages/playground/src/views/minting_view.vue
@@ -4,7 +4,7 @@
     <v-alert variant="tonal" type="warning" class="mb-4 pa-2">
       <p>All data subject to change</p>
     </v-alert>
-    <v-form @submit.prevent="mintingHash()" class="d-inline-flex w-100">
+    <v-form class="d-inline-flex w-100">
       <FormValidator v-model="isValidForm">
         <InputValidator
           :value="receiptHash"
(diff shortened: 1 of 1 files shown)


0oM4R (changes requested):
I think one of the two error msgs here will be better 
image

0oM4R (changes requested):

packages/playground/src/views/minting_view.vue
```diff
@@ -207,17 +205,30 @@ function reset() {
...
+
+  return asyncValidator;
 }
+
+watch(
+  () => receiptHash.value,
+  () => {
+    reset();
+  },
+);
```
0oM4R, line 231: we can move this to the component 
`@update:model-value="reset"`

0oM4R (approved):
great work thanks ya Alaa

amiraabouhadid (approved):

PR #2226, opened 2024-02-22
Slice Iperf array to contain only 4 elemnts max

zaelgohary:
### Description

Slice Iperf array to contain only 4 elements max

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1388

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/packages/playground/src/components/node_details_cards/iperf_details_card.vue
+++ b/packages/playground/src/components/node_details_cards/iperf_details_card.vue
@@ -61,10 +61,11 @@ export default {
     const getNodeIPerfCard = async (): Promise<NodeDetailsCard[]> => {
       const res = await gridStore.grid.zos.getNodeIPerfTest({ nodeId: props.node.nodeId });
       // filter the returned result to show node other than the one being tested against
-      let array = res.result
+      const array = res.result
         .filter(
           (node: any) => node.download_speed && node.upload_speed && !node.error && node.node_id !== props.node.nodeId,
         )
+        .slice(0, 4)
         .map(node => ({
           name: node.test_type.toLocaleUpperCase(),
           type: isIPv4(node.node_ip) ? "IPv4" : "IPv6",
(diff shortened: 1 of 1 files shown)


packages/playground/src/components/node_details_cards/iperf_details_card.vue
```diff
@@ -61,7 +61,7 @@ export default {
     const getNodeIPerfCard = async (): Promise<NodeDetailsCard[]> => {
       const res = await gridStore.grid.zos.getNodeIPerfTest({ nodeId: props.node.nodeId });
       // filter the returned result to show node other than the one being tested against
-      const array = res.result
+      let array = res.result
         .filter(
           (node: any) => node.download_speed && node.upload_speed && !node.error && node.node_id !== props.node.nodeId,
         )
```
MohamedElmdary, line 67:
```suggestion
      const array = res.result
        .filter(
          (node: any) => node.download_speed && node.upload_speed && !node.error && node.node_id !== props.node.nodeId,
        ).slice(0, 4)
```

We can slice 0 ~ 4 (4 index is excluded) if array has less than 4 element it will return them only

packages/playground/src/components/node_details_cards/iperf_details_card.vue
```diff
@@ -71,6 +71,10 @@ export default {
           downloadSpeed: format(node.download_speed),
           uploadSpeed: format(node.upload_speed),
         }));
+
+      if (array.length > 4) {
+        array = array.slice(4);
+      }
```
MohamedElmdary, line 77:
```suggestion
```

If we add `.slice(0, 4)` we won't need check anymore

0oM4R:
Can you please explain why the first 4, i can't understand this point.

zaelgohary:
> Can you please explain why the first 4, i can't understand this point.

It matters only in case of nodes 11 & 14 as only 4 objects will be returned, hence the first 4.

0oM4R (approved):
works fine thank you
image

MohamedElmdary (approved):

Issue #2234, opened 2024-02-23
Wrong tft value conversion on dashboard contract


Mik-TF:
# Situation

As reported kindly by Hristo D on telegram (see [here](https://t.me/threefoldtesting/14264)), the conversion on the dashboard contracts section seems off:

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/77026219/66c62ad6-3495-489a-ad2e-fe747de907ee)

# Fix

It seems it's the tft to usd value that is off.

khaledyoussef24:
verified on devnet
version : 7b43120


now the total conversion is multiplied by .017 to give a reasonable exchange rate.
![Image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/66ca8311-9f7c-46e1-afc6-6e0ed914d24a)


![Image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/f9ef3f0a-74a9-466b-ae2c-c883372966a1)


https://app.testlodge.com/a/26076/projects/40893/suites/234374?expand_section=430129#case_4161190

ramezsaeed:
Verified:
- QAnet
- 2.3.0-alpha12

PR #2236, opened 2024-02-25
fix tft to USD conversion should be multiplied not divided

AlaaElattar:
### Description

The conversion was wrong
### Changes

- Instead of dividing by the value of tft in usd, it should be multiplied
### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/2234
### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/packages/playground/src/weblets/tf_contracts_list.vue
+++ b/packages/playground/src/weblets/tf_contracts_list.vue
@@ -126,7 +126,7 @@ async function onMount() {
         nodeStatus.value = await getNodeStatus(nodeIDs.value);
         totalCost.value = getTotalCost(contracts.value);
         const TFTInUSD = await queryClient.tftPrice.get();
-        totalCostUSD.value = totalCost.value / TFTInUSD;
+        totalCostUSD.value = totalCost.value * TFTInUSD;
       } catch (error: any) {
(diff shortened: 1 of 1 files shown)


packages/playground/src/weblets/tf_contracts_list.vue
```diff
@@ -126,7 +126,7 @@ async function onMount() {
         nodeStatus.value = await getNodeStatus(nodeIDs.value);
         totalCost.value = getTotalCost(contracts.value);
         const TFTInUSD = await queryClient.tftPrice.get();
-        totalCostUSD.value = totalCost.value / TFTInUSD;
+        totalCostUSD.value = totalCost.value * TFTInUSD;
```
AhmedHanafy725, line 129: this is correct, but we multiply the result by 1000 somewhere so the value is still wrong

AhmedHanafy725 (changes requested):

packages/playground/src/weblets/tf_contracts_list.vue
AlaaElattar: should be fixed

AhmedHanafy725 (approved):

Issue #2267, opened 2024-02-27
Add new environment variable to identify the network in case custom network is used


AhmedHanafy725:
The target is to enable/disable the features that depend on the network environment (dev, qa, ...) for example:

- get TFT button
- minting receipts

maayarosama:
As per discussion with @AhmedHanafy725,  the user is now restricted to using only the following four networks ```main, dev, qa and test```. 

However the user can set any of the following urls:
-   GRAPHQL_URL
-  GRIDPROXY_URL
-  SUBSTRATE_URL
-  ACTIVATION_SERVICE_URL
-  RELAY_DOMAIN
-  BRIDGE_TFT_ADDRESS
-  STATS_URL
-  STELLAR_NETWORK

If not set the network's default urls will be used.

A-Harby:
Can you also update the docs
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/b8b3348c-3090-4faa-bd74-7d4d973be449)

hossnys:
stats url is hard coded here https://github.com/threefoldtech/tfgrid-sdk-ts/blob/v2.3.0-rc1/packages/playground/scripts/build-env.sh#L9 as it overrides stats passed env variable

hossnys:
in network inspect of https://dashboard.02.dev.grid.tf it shows mixed connection to two different gridproxy instances , and two relays , which maybe happened because urls passed to the grid client but it's not to gridproxy or graphql clients  , so please check that also .

A-Harby:
Verified, 
Docs.
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/31b4be83-f33e-42e4-823c-a0ef6171efcb)
Testnet
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/42265215-0ffd-41aa-b483-663d74d1038e)
02dev
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/4900c4ae-789c-4813-a46a-0c73728b741a)

PR #2279, opened 2024-02-28
Using default urls if env vars are'nt avaliable

maayarosama:
### Description

If all required env vars are missing, we'll use default urls
### Changes

- Adding condition to check if env vars are empty, if so the use default urls in build-env bash in playground and in client in grid_client
- When connecting to the grid always send the env vars regardless of network

### Related Issues

#2267
### Checklist

- [ ] Tests included
- [x] Build pass
- [x] Documentation

### DIFF
--- a/packages/grid_client/src/client.ts
+++ b/packages/grid_client/src/client.ts
@@ -67,11 +67,21 @@ class GridClient {
       backendStorageType: clientOptions.backendStorageType ? clientOptions.backendStorageType : BackendStorageType.auto,
       deploymentTimeoutMinutes: clientOptions.deploymentTimeoutMinutes ? clientOptions.deploymentTimeoutMinutes : 10,
     };
-    this.clientOptions.substrateURL = clientOptions.substrateURL;
-    this.clientOptions.proxyURL = clientOptions.proxyURL;
-    this.clientOptions.graphqlURL = clientOptions.graphqlURL;
-    this.clientOptions.activationURL = clientOptions.activationURL;
-    this.clientOptions.relayURL = clientOptions.relayURL;
+
+    if (
+      this.clientOptions.network === NetworkEnv.dev ||
+      this.clientOptions.network === NetworkEnv.qa ||
+      this.clientOptions.network === NetworkEnv.test ||
+      this.clientOptions.network === NetworkEnv.main
+    ) {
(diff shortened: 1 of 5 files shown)


packages/grid_client/src/client.ts
```diff
@@ -177,30 +158,38 @@ class GridClient {
...
-      urls.substrate = "wss://tfchain.grid.tf/ws";
-      urls.graphql = "https://graph.grid.tf/graphql";
-      urls.activation = "https://activation.grid.tf/activation/activate";
+    if (
+      !this.clientOptions.proxyURL ||
+      !this.clientOptions.relayURL ||
+      !this.clientOptions.substrateURL ||
+      !this.clientOptions.graphqlURL ||
+      !this.clientOptions.activationURL
+    ) {
```
AhmedHanafy725, line 167: this means that I have to provide all the links, but we might change one of them and need the rest to be the default, you should use the value in the `clientOptions` (if any) or the default value.

AhmedHanafy725 (changes requested):

0oM4R (changes requested):

AhmedHanafy725 (approved):

0oM4R:
can you please update the related docs ? 
i think that is : https://github.com/threefoldtech/tfgrid-sdk-ts/blob/development/packages/playground/docs/build.md

packages/playground/docs/build.md
```diff
@@ -14,10 +14,10 @@ bash ../scripts/build-env.sh
 - By default, it runs on dev mode. the values already sat on the config file. if you want to change the mode
 
   ```bash
-  export MODE=dev | qa | test | main | custom
+  export MODE=dev | qa | test | main
   ```
 
-- In case you chose `custom` you will need to provide all the needed values which is
+- In case the user wants to use different urls, he can chose change any of the following urls, In case a url wasn't provided by the user, the default network's url will be used
```
AhmedHanafy725, line 20:
```suggestion
- In case the user wants to use different urls, he can change any of the following urls, In case a url wasn't provided by the user, the default network's url will be used
```

MohamedElmdary:
@AhmedHanafy725 @maayarosama should we outsource urls into it's own file?

0oM4R (approved):

PR #2328, opened 2024-03-04
Removing hard coded stats url from build-env script and passing urls to grid proxy and graphql client

maayarosama:
### Description

Removing hard coded stats url from build-env script and passing urls to grid proxy and graphql client

### Changes
-  passing urls to grid proxy and graphql client instead of network
- In playground passed the appropriate url as a parameter to both grid proxy and graphql client , instead of passing the network and urls if set
- In new stats passed networks as string array instead of network enum
-  Adding relayUrl to GridClientConfig
- Updating config file in playground's public folder

### Related Issues

#2267
#2329
### Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation


packages/new_stats/src/components/network_filter.vue
```diff
@@ -85,12 +84,12 @@ const networks = ref([
 ]);
 
 // initial state; all network selected
-const selectedNetworks = ref([Network.Dev, Network.Main, Network.Test]);
+const selectedNetworks = ref(["dev", "main", "test"]);
```
AlaaElattar, line 87: why u didn't add `qa` ?

maayarosama: I'm not entirely sure why qa wasn't added from the start, @0oM4R do you have any idea why?

0oM4R: Yes we don't have to include qa net on our stats, the required network is only dev, test and main

AlaaElattar: Thank youu, i didn't know that.

AlaaElattar (approved):

Issue #2282, opened 2024-02-28
(QA/Testnet Verification) Tooltip in the stellar bridge address is only pointing to the dev address despite changing the network


A-Harby:
### Description

Qanet, 2.3.0-alpha12.

A clear description of the bug is:

The tooltip in the bridge stellar address is only pointing to the dev address and doesn't change when the network changes. 
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/a955739d-b498-4e0a-95ee-cf4edf15a9ee)

The link in the tooltip is: https://stellar.expert/explorer/testnet/account/GDHJP6TF3UXYXTNEZ2P36J5FH7W4BJJQ4AYYAXC66I2Q2AH5B6O6BCFG

A-Harby:
Verified, Mainnet 2.3.0.

The link is changed to the correct network address.
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/f99bfb54-620b-413d-9a29-ef1285c0921c)

ramezsaeed:
Verified:

- Testnet

![Screenshot from 2024-03-13 18-38-17](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/16083592/abf531db-7dc7-43b6-97fa-f3c1847c4ab1)

- Mainnet
![Screenshot from 2024-03-13 18-37-01](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/16083592/1f8e6bf4-36ea-44c5-b6df-bdbb6cecf4cc)

PR #2340, opened 2024-03-05
change stellar link for test and mainnet 

AlaaElattar:
### Description

- Wrong link was used for testnet and mainnet.
### Changes

- added condition to use right link in case of testnet and mainnet.
### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/2282
### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation


MohamedElmdary (approved):
Good job
however `computed` is not needed here as `window.env` isn't reactive value so we can stick to regular if check e.g `const x = y ? 'if' : 'else' ` 
but it's fine ... Thanks

amiraabouhadid (approved):

Issue #2357, opened 2024-03-06
Stats number aren't display on the cards


A-Harby:
### Description

A clear description of the bug is:

The stats number is not shown on the cards anymore.
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/0cca5d2a-f4f6-4181-9152-51b3024a45a9)

Must include any relevant identifiers, like:

- **Network**: Testnet

- **Version**: 2.3.0-rc3

A-Harby:
Also on the landing page, so before and after login.
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/bdb33921-461e-427b-be94-08c89420e6bd)

A-Harby:
Verified, 9e745f8.

The numbers are now displayed.

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/1c2f8849-66d5-480f-8125-590641fd1cc7)
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/e0834b2e-adac-4ce8-8fcd-3e7f3c47faca)

[TC2457 - Homepage before login](https://app.testlodge.com/a/26076/projects/40893/suites/234374?expand_section=422331#case_4177817)
[TC2458 - Homepage after login](https://app.testlodge.com/a/26076/projects/40893/suites/234374?expand_section=422331#case_4177818)

ramezsaeed:
Verified:
- Mainnet
- 2.3.0
![Screenshot from 2024-03-13 18-43-16](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/16083592/9a228900-5d34-4a32-8f15-04cdcd549f7c)

PR #2382, opened 2024-03-10
Fix stats summary endpoint delay 

0oM4R:
### Description

If a request takes too long, it adversely impacts the responsiveness of the playground UI. Therefore, we must expedite the process of obtaining stats. While timeouts are an option, we discovered that even after returning, promises persist, leading to potential duplication of time if we attempt to call the summary endpoint while a previous request is still pending.

Due to the limited functionality of NJS, we are unable to utilize AbortController, even with the [NJS node module](https://nginx.org/en/docs/njs/node_modules.html). To address this issue, the proposed solution involves implementing two separate endpoints to mitigate the aforementioned problem.

Additionally, the cached stats summary can be updated by calling the stats-updated endpoint. We can incorporate this step into the nightly workflow or a similar process to ensure the stats are regularly refreshed without impacting the UI's responsiveness.

### Changes
make the stats-summary always return the cached data 
add  new endpoint stats-update, that updates the cache

### Related Issues

#2357

### Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation

### DIFF
--- a/packages/new_stats/nginx/conf.d/default.conf
+++ b/packages/new_stats/nginx/conf.d/default.conf
@@ -14,6 +14,11 @@ server {
         default_type application/json;
         js_content main.getStats;
     }
+        location /api/stats-update {
+        add_header 'Access-Control-Allow-Origin' '*';
+        default_type application/json;
+        js_content main.updateStats;
+    }
     #error_page  404              /404.html;
 
     # redirect server error pages to the static page /50x.html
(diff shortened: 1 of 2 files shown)


amiraabouhadid (approved):
LGTM

Mahmoud-Emad (approved):
Gamed ya Omar

PR #2424, opened 2024-03-13
Merge development_314 branch in the development branch

AhmedHanafy725:


packages/playground/src/utils/nodeSelector.ts
```diff
@@ -199,6 +199,30 @@ export function loadNodes(gridStore: ReturnType<typeof useGrid>, filters: Filter
...
+  gridStore: ReturnType<typeof useGrid>,
+  node: NodeInfo | undefined | null,
+): Promise<true> | never {
+  if (!node || !node.nodeId) {
+    throw "Node ID is required.";
+  }
+
+  try {
+    const contractInfo = await gridStore.grid.contracts.get({ id: node.rentContractId });
+    if (contractInfo.state.gracePeriod) {
```
0oM4R, line 212: This makes all deployments fail.

AhmedHanafy725: We can skip this check if the rent contract is 0 as the node in this case is not rented.

Mahmoud-Emad: Its should be fixed in https://github.com/threefoldtech/tfgrid-sdk-ts/pull/2431/files#diff-c5cb2a07f18e9933e21237d4b7caeac5ea18ab5a451b4cd1d5cf73472dffaa4fL83-R83

Mahmoud-Emad (approved):

zaelgohary (approved):

Issue #1882, opened 2024-01-04
Enhancement Request: Implementing Multiple SSH Keys with Enhanced Management Features


Mahmoud-Emad:
### Description

The current SSH key page offers limited functionality with support for only a single SSH key. To improve user experience and flexibility, I propose the implementation of a feature that allows users to manage multiple SSH keys.

**Enhancements Proposed**

1. **Multiple SSH Keys:**
   - Users should be able to save and manage more than one SSH key.
   - The option to activate only one SSH key at a time should be available.

2. **Key Activation:**
   - Users can easily switch between SSH keys and activate/deactivate one.
   - Each SSH key should have a generated fingerprint, and users must enter the fingerprint when activating a key.

3. **Adding SSH Keys:**
   - Users should have the ability to add multiple SSH keys with no set limitations.

4. **Deletion Options:**
   - Users can delete individual SSH keys.
   - Batch deletion should be supported, allowing users to select multiple keys for deletion.

**Benefits:**
- Enhanced flexibility in managing multiple SSH keys.
- Improved security with unique fingerprints for each key.
- Streamlined user experience with the ability to add, activate, and delete keys conveniently.

**Additional Context:**
This enhancement aims to provide users with a more robust and user-friendly interface for managing their SSH keys, catering to scenarios where multiple keys are used for different purposes.

Here are some designs we can stand for generating our design by @ehab-hassan 

- ![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/57001890/d06144ca-87ee-4ba8-bc44-ba968ceb1253)
- ![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/57001890/cc94ff19-d460-47b6-bd5b-8e269732c673)

AhmedHanafy725:
after trying to deploy a vm with more than one ssh key, it works fine. Just set the sshkey env var with all required sshkeys separated by `\n`.

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/41957921/f922f623-b626-4f1b-98e1-2c241c38cdc3)

Mahmoud-Emad:
## Issue update 

### Work in progress

Working on the design of adding the multiple ssh-keys feature

[Screencast from 03-18-2024 04:59:50 PM.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/57001890/d273ed5c-8e4c-4668-b08f-caf7fbe02cd9)

[...]

PR #2463, opened 2024-03-18
Implementing Multiple SSH Keys with Enhanced Management Features

Mahmoud-Emad:
### Description

Implemented multiple SSH key feature and SSH key management 

### Changes 
- Implemented a helper function to generate a name for the generated key to be saved with the key.
- Added activation and deletion logic for SSH keys.
- Worked on the feature to delete all selected keys, preparing to delete keys based on types like active or all keys.
- Unified UI elements by removing the tonal theme from cards to maintain consistency with the contracts list view.
- Enhanced UI by adding another dialog for importing SSH keys and updated the colors of the table.
- Calculated SSH key fingerprints, consolidated form dialogs, and introduced a key details dialog while removing redundant dialog.
- Refined the copy input wrapper to conditionally display the copy icon based on field data presence and improved the functionality of the generate dialog.
- Migrated old SSH keys to the new SSH key type and set them as active keys.
- Fixed the issue when saving SSH keys by loading them from the chain instead. Work in progress on updating the chain when pressing on active/inactive or deleting keys.
- Enhanced the way of saving/removing the keys and improved the import key form by adding a rule to validate the entered SSH key.
- Merged SSH key management functionality into all applications.

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1882

### Screenshots

- image
- image
- image
- image
- image
- image
- image
- image
- image

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation


packages/playground/src/components/ssh_keys/ManageSshDeployemnt.vue
MohamedElmdary, line 1: why are we using v2 syntax?

Mahmoud-Emad: Would you like to use V3 syntax? i see it's more readable also it I noticed that it used in many different places

MohamedElmdary: can  you mention where it's used so we make issue for updating it? @Mahmoud-Emad

amiraabouhadid (commented):
It says that this sshkey was created today. However, this sshkey was created long before today. 

image

MohamedElmdary:
Deploy button need to be disable if no key is select

### How to do that?
basically, use our validator form api but manually
example for such solution can be found here https://github.com/threefoldtech/tfgrid-sdk-ts/blob/development/packages/playground/src/components/node_selector/TfSelectionDetails.vue#L144

Mahmoud-Emad:
> It says that this sshkey was created today. However, this sshkey was created long before today.
> 
> image

i have keys created from 31/3

image

MohamedElmdary (approved):
Good job

amiraabouhadid (approved):

Issue #1691, opened 2023-12-14
enhance list requests in data tables 


0oM4R:
### Description

Currently, we are adding `ret_count` to every request on the table, which effects the performance, so if there is no change in the parameters, there is no change in the total count, and there is no need to make the requests heavier with ret_count.
  

### Implementation

we can get the total count on_mount and on any change of filters or delete\add actions

khaledyoussef24:
verified on devent dashboard node finder that the request is made.
verified also on gridproxy using swagger `ret_count` is found and enabled as a default(set as true).

swagger:
![Image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/088962c3-8d62-469f-a5de-818829b3471c)
![Image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/2e6f8806-a83c-4ea2-92c5-d74efbb46628)

[gridproxy](https://gridproxy.dev.grid.tf/nodes?page=1&size=20&ret_count=true&sort_by=status&sort_order=asc):


![Image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/5003e8c7-9818-4ac3-af7b-236fd6a2c674)



dashboard:
![Image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/6943e218-73f1-43e4-99f2-c88cf4e5130f)

PR #2488, opened 2024-03-21
enhance list requests in nodes and farms tables

0oM4R:
### Description

get the total count on_mount and on any change of filters or delete\add actions

### Changes

in general, the tf filter component handle the changes on the params and it also handle on mounted and clear by calling `apply` 
so in case of calling apply, we are passing ret_count flag to be true, the default value of the flag is false
if the flag passed, that's mean we are changing all the table data so we need to reset page to 1 and return count.

### Related Issues

- #1691 

### Documentation PR

For UI changes, Plaese provide the Documetation PR on [info_grid](https://github.com/threefoldtech/info_grid)

### Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation

### DIFF
--- a/packages/playground/src/views/farms.vue
+++ b/packages/playground/src/views/farms.vue
@@ -1,6 +1,6 @@
 <template>
   <view-layout>
-    <TfFiltersContainer @apply="loadFarms" class="mb-4" :loading="loading">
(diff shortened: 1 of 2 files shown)


MohamedElmdary (approved):

zaelgohary (approved):
Good job ya Omar.

Issue #2220, opened 2024-02-21
Make the manual url configurable 


AhmedHanafy725:
- Use the staging manual for devnet, qa, and testnet. (staging's url: https://www2.manual.grid.tf/)
- Update all docs urls in the dashboard to use the configured manual url.

zaelgohary:
Work completed:

Configured [manual URL](https://www.manual.grid.tf) as a global prop and used it in all playground pages.

khaledyoussef24:
verified on devnet
version : 5c12959



![Image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/0c52b77a-964f-4681-a417-6929e4237081)
![Image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/1ffa3db2-e71b-4208-a0d2-29c6f12561e5)

verified that all networks now leads to the [same link](https://www.manual.grid.tf/)

PR #2495, opened 2024-03-24
Configure manual as a global prop

zaelgohary:
### Description

- Added MANUAL_LINK to the build script so it can be added before deployment
- Configure manual as a global prop
- Update config.js

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/2220

### Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation


MohamedElmdary (approved):

AhmedHanafy725 (commented):
please run the script from the `public` directory to have this link in the config.js so the auto-update workflow can have this change

amiraabouhadid (approved):

packages/playground/scripts/build-env.sh
```diff
@@ -6,8 +6,7 @@ STELLAR_NETWORK="${STELLAR_NETWORK:=test}"
 TIMEOUT="${TIMEOUT:=10000}"
 PAGE_SIZE="${PAGE_SIZE:=20}"
 MINTING_URL="https://alpha.minting.tfchain.grid.tf"
-STATS_URL="https://stats.grid.tf"
-
+MANUAL_URL="https://www.manual.grid.tf"
```
AhmedHanafy725, line 9: this will always overwrite the value in the environment.
```suggestion
MANUAL_URL="${MANUAL_URL:-https://www.manual.grid.tf}"
```

AhmedHanafy725 (changes requested):

packages/playground/scripts/build-env.sh
zaelgohary: There's no value in the environment. Only one value for all nets.

AhmedHanafy725: okay, but this value can't be changed. If the manual link gets changed and exported, it won't reflect in the script.

AhmedHanafy725 (approved):

Issue #2479, opened 2024-03-20
(Mainnet Verification) Node uptime remains at 0% despite being initiated a few days ago - Hristo Damianov (hristo.damianov@gmail.com)


Cameron1986labeyrie:
Uptime stays on 0% since they were started a few days ago 

Node:6539 ( Created on : 3/15/24, 11:12 PM) 
Node:6540 ( Created on : 3/16/24, 12:43 AM)

Node:6539 Last uptime recorded: Wed Mar 20 2024 09:30:36 GMT+0000
Node:6540 Last uptime recorded: Wed Mar 20 2024 09:31:30 GMT+0000

![image_efn949](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/152964575/11eaf1ae-6997-4d15-b052-75bc671db3fc)
![image_von1m3](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/152964575/903cde40-d084-4542-ae80-002b72bd52d9)

0oM4R:
Update: 

Investigation and Solution: 
the issue is with `getNodeMintingFixupReceipts`, the request fails and this is not handled
need to move this function into  inner try block 

```

        
       GET https://alpha.minting.tfchain.grid.tf/api/v1/node/6539 404 (Not Found)
```

khaledyoussef24:
verified and working on mainnet 
V 2.4.2
minting requests response are working now for the nodes giving the same starting time 
https://alpha.minting.tfchain.grid.tf/api/v1/node/6539
https://alpha.minting.tfchain.grid.tf/api/v1/node/6540

![Image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/0af74861-e4d9-4aed-836a-98e940fc3a54)


![Image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/bb8c66d4-69c5-4649-9aea-e1aab9d831ec)
![Image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/feecc049-6298-4be1-9950-cf530dc68536)
![Image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/4021369e-d00d-4778-9c5d-92e99b2f7b0d)


@Cameron1986labeyrie reply and reopen if the issue is still there but for now the minting for the nodes are working fine

Cameron1986labeyrie:
Thank you for the update @khaledyoussef24

PR #2497, opened 2024-03-26
Fix: node uptime

0oM4R:
### Description

for the mentioned nodes, they were created early this month, so there will be no minting receipts for them until the next month, 
so getting the reports was failing and exiting the function, The next few lines in the function is related to up-time calculation, which was not reached, so the result is 0%

### Changes

add getting minting reports to try block to make sure the function will continue even if there is an error with fetching receipts
### Related Issues

- #2479 

### Documentation PR

For UI changes, Plaese provide the Documetation PR on [info_grid](https://github.com/threefoldtech/info_grid)

### Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation


packages/playground/src/dashboard/components/user_nodes.vue
```diff
@@ -204,7 +204,7 @@ export default {
     async function getUserNodes() {
       try {
         loading.value = true;
-        const twinId = profileManager.profile!.twinId;
+        const twinId = 11261;
```
MohamedElmdary, line 207: why did we use `11261` as twinId instead of logged in user's one?

MohamedElmdary (approved):

samaradel (approved):
Good job!

Issue #2537, opened 2024-04-07
🐞 [Bug]: Playground build fails due to incompatible data types


amiraabouhadid:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### which package/s did you face the problem with?

Dashboard

### What happened?

Development branch fails because uid is a number that is passed to an argument that accepts it as a string. 

### Steps To Reproduce

1. just look at the development branch build log

### which network/s did you face the problem on?

Dev

### version

n/a

### Relevant screenshots/screen records

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/56790126/8db243c2-6155-442b-8703-424b7582adae)

### Relevant log output

```console
please refer to above screenshot
```

A-Harby:
verified, the build is passing after the fix: https://github.com/threefoldtech/tfgrid-sdk-ts/actions/runs/8585875597.
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/e9a860f3-7a88-41d3-b696-eee0ca4e37a2)

PR #2538, opened 2024-04-07
covert uid from number to stringer when passed as argument

amiraabouhadid:
### Description

Converted UID from number to string before passing to functions

### Related Issues
#2537


zaelgohary (approved):
LGTM

MohamedElmdary (approved):

Issue #2543, opened 2024-04-08
New Ssh-Keys page enhancements


Mahmoud-Emad:
### Which package/s are you suggesting this feature for?

Dashboard

### Is your feature request related to a problem? Please describe

It's not a problem

### Describe the solution you'd like

1. There Should be only one table containing both keys active and inactive 
2. The 'Manage Ssh Keys' button in the deployment needs some enhancement
3. Update the activation `Checkbox` to only active/inactive the keys and `don't open the details dialogue`
4. The SSH-Keys IDs should start from 1
5. Generate a name for the old key if there is no name

Mahmoud-Emad:
this needs to happen directly after login

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/57001890/b7631a98-3ef2-4fad-97ea-9131620a197b)

Mahmoud-Emad:
this looks very odd

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/57001890/45754df9-a4f7-44c7-beaa-2a4336ad5ebd)

Mahmoud-Emad:
Should be just one table contains all keys
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/57001890/55226ec4-87e7-4305-8758-a38f0ebb3a25)

zaelgohary:
Section titles are bigger than the ones on other pages. Maybe we can use the same font size used in them.

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/40770501/ac0071b2-22ac-4f44-a531-40173c981d8e)

zaelgohary:
The ID in tables starts with 2.

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/40770501/2e021c56-61dc-4d30-b029-2a00563a48e1)

samaradel:
@ehab-hassan can you provide better enhancement suggestions for ```2. The 'Manage Ssh Keys' button in the deployment needs some enhancement``` please

it looks 
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/11662805/0a42b01f-433c-4da3-a41a-10768ec612f5)

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/11662805/4ff404ac-ac43-43ab-a87b-0d2b04182995)

ehab-hassan:
[...]

PR #2554, opened 2024-04-14
Update sshkey page

samaradel:
### Description
- Remove Active keys table
- Add all keys into one table
- generate a name for keys that haven't a name
- Reduce the size of titles
- prevent opening the dialog of ssh key details by clicking on the active checkbox
- Generate name when the key is imported without name

### Changes

image
### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/2543
- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1882

### Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation


MohamedElmdary (approved):

AlaaElattar (approved):
Great Job

Mahmoud-Emad (changes requested):
Good job ya Samar, as we agreed we have to migrate the old key once the user login in background.

image

MohamedElmdary (approved):

MohamedElmdary (approved):

zaelgohary (commented):
I was able to save the same key more than once. (same name and SSH)

image

packages/playground/src/views/sshkey_view.vue
0oM4R, line 1: looks like we didn't handle the insufficient balance case
image

zaelgohary (commented):
When deleting a key, table IDs won't update.

image

packages/playground/src/views/sshkey_view.vue
0oM4R: we need to add error handler also, it keep loading forever

packages/playground/src/components/ssh_keys/SshTable.vue
```diff
@@ -214,20 +206,26 @@ export default defineComponent({
...
     };
 
-    const activateKey = (key: SSHKeyData) => {
-      if (key.isActive) {
-        emit("inactive", key);
-      } else {
-        emit("active", key);
-      }
+    const toggleKeyActivation = (key: SSHKeyData) => {
+      console.log(key.isActive);
```
0oM4R, line 210: we should remove this line

zaelgohary:
Two buttons have the same label, can we change the first one to 'Export All' instead?

image

zaelgohary:
Header font size doesn't match with other headers.

image

image

Mahmoud-Emad:
> When deleting a key, table IDs won't update.
> 
> image

Shouldn't change, we save it into a database and it's a valid behavior, if we changed the keys I'd each time we delete it'll be a big traffic.

packages/playground/src/utils/validators.ts
```diff
@@ -733,16 +733,6 @@ export function pattern(msg: string, config: RegexPattern) {
...
-/**
- * Checks if a given string is a valid SSH key.
- * @param key - The SSH key string to be validated.
- * @returns True if the key is a valid SSH key, false otherwise.
- */
-export function isValidSSHKey(key: string): boolean {
-  const sshKeyRegex = /^(ssh-rsa|ssh-dss|ecdsa-[a-zA-Z0-9-]+|ssh-ed25519)\s+(\S+)\s+(\S+)/;
-  return sshKeyRegex.test(key);
-}
-
```
0oM4R, line 745: why did we move the validation, Shouldn't all validators be here ?

zaelgohary:
Can we change the info to be "Manage SSH keys easily, switch between them, and activate or deactivate keys as needed for accessing deployed machines. Simplify key distribution and effectively manage access to nodes."

https://github.com/threefoldtech/tfgrid-sdk-ts/blob/9ce65149d66df3ae7d1baa8bafe6c6eb8f2b3256/packages/playground/src/views/sshkey_view.vue#L9-L12

zaelgohary:
The lower export button doesn't work as expected.

[export_ssh.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/40770501/720409d2-6e1b-45cf-bc97-70020dd7dcbf)

packages/playground/src/utils/validators.ts
Mahmoud-Emad: It moved to be inside the class management, I want to group all ssh-related things in one sope

packages/playground/src/views/sshkey_view.vue
Mahmoud-Emad: Resolved

packages/playground/src/components/ssh_keys/SshTable.vue
Mahmoud-Emad: Removed

zaelgohary (approved):
Good job, Emad!

0oM4R (changes requested):
I still have an unhanded errors on low balance account 
image

packages/playground/src/utils/ssh.ts
```diff
@@ -0,0 +1,232 @@
...
+      return;
+    }
+
+    const balance = await loadBalance(grid!);
+    if (balance.free === balance.locked || balance.free < this.updateCost) {
+      throw new InsufficientBalanceError(
+        "Your wallet balance is insufficient to save your SSH key. To avoid losing your SSH key, please recharge your wallet.",
+      );
+    }
+
```
0oM4R, line 97: I think we should wait for the transaction to throw this error and handle it, 
this step should be useful only if we need to disable the btns if the balance is not enough.
what i mean, we should check for the errror if it is an instance of InsufficientBalanceError then create a toast with readable error

0oM4R (changes requested):
On low balance, an error is handled, but the status has changed!
After reloading the page, the status went back to active. 
also i think the toast should have more friendly message 
[Screencast from 21 أبر, 2024 EET 11:21:09 ص.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/62248851/8fb38930-44ac-41d1-8426-7fb8a328c075)

packages/playground/src/utils/ssh.ts
```diff
@@ -0,0 +1,232 @@
...
+      return;
+    }
+
+    const balance = await loadBalance(grid!);
+    if (balance.free === balance.locked || balance.free < this.updateCost) {
+      throw new Error(
+        "Your wallet balance is insufficient to save your SSH key. To avoid losing your SSH key, please recharge your wallet.",
+      );
+    }
+
```
0oM4R, line 97: I think there is no need to check for balance and throw errror here, the client already do that for us, 
and this is the idea of using types package and errror classes, client throw an error  then we customize the displayed error massage for the user based on the thrown error class.

packages/playground/src/views/sshkey_view.vue
```diff
@@ -115,258 +100,196 @@
...
+  sshKeysManagement.export(exportKeys);
+  isExporting.value = false;
+};
+
+const updateActivation = async (key: SSHKeyData) => {
+  activating.value = key.activating = true;
+
+  try {
+    await sshKeysManagement.update(allKeys.value);
+  } catch (e: any) {
```
0oM4R, line 194: in this catch, we should look up for "insufficientBalanceError" and create a toast with friendly error message, 
and the else block should handle other cases with something like " something went wrong"

packages/playground/src/utils/ssh.ts
Mahmoud-Emad: Yeah, you have a point, errors should be handled outside the management class, However, it is only used in the SSH component and not planned to be used somewhere else, I think we can keep all SSH-related in one place.

packages/playground/src/views/sshkey_view.vue
Mahmoud-Emad: The ssh script handle this.

packages/playground/src/utils/ssh.ts
Mahmoud-Emad: Yeah, indeed.

0oM4R (approved):

0oM4R (approved):
Thanks ya Omda great job

Mahmoud-Emad (approved):

Issue #2471, opened 2024-03-19
Show number of workers on top of weblet


0oM4R:
### Which package/s are you suggesting this feature for?

Dashboard

### Is your feature request related to a problem? Please describe

adding workers is not clear, how many workers i just added.
![Screenshot from 2024-03-19 11-25-35](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/62248851/28e49249-5cb4-4388-ad41-44e603103795)
in this picture i added about 2 workers but if want to make sure, i have to scroll down to count them

### Describe the solution you'd like

show number of workers beside `Worker` tab title or beside add button

zaelgohary:
Work completed:

Added workers optional prop in tabs & applied change in k8s & caprover.

khaledyoussef24:
verified on devnet 
version : e1cdba0
now no. of workers is shown in the top of the dashboard

![Image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/04eaf51c-f8c3-4daa-9a6c-c13af38dba04)
![Image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/d67ff84f-9283-4532-bf72-9fe8dd2ac22b)
![Image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/25cdb149-382e-408e-8f61-bf2a1197a68b)

user can easily know the no. of added workers
tried also to delete workers after addition to make sure it decrease as well

PR #2565, opened 2024-04-15
Add workers number chip in k8s & caprover

zaelgohary:
### Changes

- Add workers optional prop in tabs
- Apply change in k8s & caprover

image

image

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/2471

### Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation

### DIFF
--- a/packages/playground/src/components/dynamic_tabs.vue
+++ b/packages/playground/src/components/dynamic_tabs.vue
@@ -13,6 +13,7 @@
         }"
       />
       {{ tab.title }}
+      <v-chip color="info" v-if="tab.title === 'Workers'" class="ml-1">{{ tab.workers }}</v-chip>
       <v-chip color="info" v-if="forms[tabs.indexOf(tab)]?.pending" class="ml-1">
         Validating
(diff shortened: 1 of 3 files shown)


0oM4R (changes requested):
can we hide the chip if there is no workers added ?

packages/playground/src/components/dynamic_tabs.vue
```diff
@@ -49,6 +50,7 @@ export interface Tab {
   value: string;
   icon?: string;
   imgPath?: string;
+  workers?: number;
```
0oM4R, line 53: can we make the name of the property more generic to refer to number of disks later if we need to ?
what do you think ?

zaelgohary: Maybe we can add another prop for the disks as well if needed later on.

AlaaElattar (approved):

0oM4R (approved):

Issue #2489, opened 2024-03-21
Allow to create contract to unlock grace period


Mik-TF:
### Which package/s are you suggesting this feature for?

Dashboard

### Is your feature request related to a problem? Please describe

When you are in grace period, and fill it your account, it takes some time until you can access your deployment. Maybe this isn't well suited for someone who has a very important workload: they would want to fill the wallet and get the workload back on track as soon as possible.

### Describe the solution you'd like

Allow the user to query the grid/tfchain to recheck the status of the workload and re-activate the contract and then the workload is accessible quickly.

### Note

We could charge a fee for this, not just the contract price. It's the "price" to pay for the user's mistake of not funding the wallet. But we allow the possibility to have the workload up and running quickly, given a reasonable fee.

ramezsaeed:
@sameh-farouk 
can you please check if we can add a trigger for this.

sameh-farouk:
Easy peasy, can be fully implemented from the front end.

By showing a button to bill the due amount and restore workload when the dashboard detects that the user has some contracts in the grace period and a reasonable TFT amount in his wallet.

The button would just call ʼ bill_contract_for_blockʼ externsic with the contract ID.
it's a signed transaction and basically user will have to pay only the the transaction fee and have his contact billed immediately and consequently revive his workload.

0oM4R:
### issue update 

##### work completed: 
added the required functions to tfchain client and grid client

##### WIP 
use  the functions in dashboard 

##### investigation
 not sure what is the best ux in this scenario, will check with Ehab after create the required logic on dashboard

0oM4R:
### Work completed 
add unlock button to the ` Contract lock Details` dialog that appears only if the user has the required balance
![Screenshot from 2024-04-21 17-13-04](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/62248851/fe141751-ea66-4796-81a6-dc3d77667c1c)

### Work in progress 

adding a button to the table to allow multiple contracts unlock

0oM4R:
@ehab-hassan  please suggest a flow for this issue, and please check https://github.com/threefoldtech/tfgrid-sdk-ts/pull/2582#pullrequestreview-2016824310

sameh-farouk:
https://github.com/threefoldtech/tfgrid-sdk-ts/pull/2582#pullrequestreview-2016824310
[...]

PR #2582, opened 2024-04-18
Unlock grace period contracts

0oM4R:
##### Update 
- now user can select contracts that are on rented node, and the coast of related rent contract is added to require funds.
- fixed free balance issue, now we are using usable balance (free -lock)
- add loading and error handling, dialog was opened before loading the details.
- update lock details by opening dialog, to avoid mismatched lock details

### Description

add button to manually unlock grace period contract, instead of waiting for the next billing cycle 
image

### Changes

- add required methods to call `billContractForBlock`  
- support unlock grace period contracts with 3 flows
 1. single contract 
Add an unlock button to the Contract lock details dailog
image
if the contract is on rented node, there is no unlock button, user should unlock the rent contract and the alert is shown
 2. multiple contracts
   selected contracts from the table could be unlocked by unlock button in each table
image
  3. unlock all contracts
  image

###  note 
we can't guarantee that the contracts will be moved to the created state, as the chain method `billContractForBlock` do not return an error if the balance is not enough, but it only updates the locked amount, 
Also, we are updating the contracts table after 30 seconds from calling the method

Screenshot from 2024-05-07 16-38-29
Screenshot from 2024-05-07 16-51-44
### Related Issues

- #2489 
- #2670
- #2687
### Documentation PR

### Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation


zaelgohary (commented):
I noticed that we let the user unlock the contract manually. What if we unlock contract/contracts automatically if the user has enough balance? This would save us any new changes in the UI.

packages/grid_client/src/modules/contracts.ts
```diff
@@ -271,6 +272,40 @@ class Contracts {
...
+    const contracts = await this.listMyContracts({ state: [ContractStates.GracePeriod] });
+    if (contracts == undefined) return LockedContracts;
+    const contractTypes = ["nameContracts", "nodeContracts", "rentContracts"];
+    for (const type of contractTypes) {
+      for (const contract of contracts[type]) {
+        LockedContracts[type][parseInt(contract.contractID)] = await this.contractLock({
+          id: parseInt(contract.contractID),
+        });
+        LockedContracts.totalAmountLocked += LockedContracts[type][parseInt(contract.contractID)].amountLocked;
+      }
```
Mahmoud-Emad, line 293:
```suggestion
      for (const contract of contracts[type]) {
          const contractID = parseInt(contract.contractID);
          const contractLockDetails = await this.contractLock({ id: contractID });
          LockedContracts[type][contractID] = contractLockDetails;
          LockedContracts.totalAmountLocked += contractLockDetails.amountLocked;
      }
```

packages/grid_client/src/modules/contracts.ts
```diff
@@ -271,6 +272,40 @@ class Contracts {
...
+  @validateInput
+  async unlockContract(ids: number[]) {
+    return await this.client.contracts.batchUnlockContracts(ids);
+  }
+
+  @expose
+  @validateInput
+  async unlockMyContracts() {
+    return await this.client.contracts.unlockMyContracts(this.config.graphqlURL);
+  }
```
Mahmoud-Emad, line 308: Let's document these functions

packages/playground/src/components/contracts_list/contracts_table.vue
```diff
@@ -119,6 +128,29 @@
           <v-btn variant="outlined" color="anchor" class="mr-2 px-3" @click="contractStateDialog = false">
             Close
           </v-btn>
+          <v-tooltip
+            :text="
+              freeBalance < getAmountLocked
+                ? `You don't have enough balance to unlock this contract`
+                : `Get your contract ready again`
```
Mahmoud-Emad, line 135:
```suggestion
                : `Unlock this contract from the grace period state.`
```

packages/playground/src/components/contracts_list/contracts_table.vue
```diff
@@ -145,11 +177,68 @@
...
+          <p class="mb-9 mt-4">
+            Unlock the following Contract<span v-if="selectedContracts.length > 1">s</span>
+          </p></v-toolbar
+        >
+      </v-card-title>
+      <v-card-text>
+        <v-alert class="my-4" type="warning" variant="tonal">
+          <div>
+            Selected contract<span v-if="selectedContracts.length > 1">s</span> Locked amount:
+            <span class="font-weight-bold">{{ lockedAmount.toFixed(4) }} </span> TFTs
```
Mahmoud-Emad, line 194: We normally use `toFixed(3)` to wrap any TFT value, why here you used `toFixed(4)`?

packages/playground/src/components/contracts_list/contracts_table.vue
```diff
@@ -145,11 +177,68 @@
...
+          {{ c.contractId }}
+        </v-chip>
+        <v-card-actions>
+          <v-spacer></v-spacer>
+          <v-btn color="anchor" variant="outlined" @click="unlockDialog = false"> Cancel </v-btn>
+          <v-tooltip
+            :text="
+              freeBalance < lockedAmount
+                ? `You don't have enough balance to unlock your contract${selectedContracts.length > 1 ? `s` : ``}`
+                : `Get your contracts ready again`
```
Mahmoud-Emad, line 214:
```suggestion
                : `Unlock your contracts from the grace period state.`
```

packages/tfchain_client/src/contracts.ts
```diff
@@ -348,6 +348,12 @@ class Contracts extends QueryContracts {
     );
     return this.client.patchExtrinsic<ServiceContract>(extrinsic);
   }
+
+  @checkConnection
+  async unlock(contractId: number) {
+    const extrinsic = await this.client.api.tx.smartContractModule.billContractForBlock(contractId);
+    return this.client.patchExtrinsic<Contract>(extrinsic);
+  }
```
Mahmoud-Emad, line 356: Let's document this function

Mahmoud-Emad (changes requested):
Good job ya Omar :)

0oM4R:
- now user can select contracts that are on rented node, and the coast of related rent contract is added to require funds.
- fixed free balance issue, now we are using usable balance (free -lock)
- add loading and error handling, dialog was opened before loading the details.
- update lock details by opening dialog, to avoid mismatched lock details

packages/grid_client/src/clients/tf-grid/contracts.ts
```diff
@@ -325,6 +335,25 @@ class TFContracts extends Contracts {
...
+  }
+  async unlockMyContracts(graphqlURL: string) {
+    const contracts = await this.listMyContracts({
+      stateList: [ContractStates.GracePeriod],
+      graphqlURL,
+    });
+    const ids: number[] = [];
+    for (const contract of contracts.nameContracts) ids.push(parseInt(contract.contractID));
+    for (const contract of contracts.nodeContracts) ids.push(parseInt(contract.contractID));
+    for (const contract of contracts.rentContracts) ids.push(parseInt(contract.contractID));
```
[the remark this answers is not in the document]
MohamedElmdary: @0oM4R @Mahmoud-Emad 
We could also do a nested loop to avoid multi loops and copying data again

```ts
// Example

const ids: number[] = [];
for (const contractType in contracts) {
  for (const { contractID } of contracts[contractType]) {
    ids.push(contractID)
  }
}
```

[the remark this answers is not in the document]
0oM4R: @MohamedElmdary @Mahmoud-Emad please check the updated code, and tell me if it need to be enhanced 
https://github.com/threefoldtech/tfgrid-sdk-ts/blob/27421b70f6ffa394919c96f4d217aa0a681c0f9f/packages/grid_client/src/clients/tf-grid/contracts.ts#L345-L352

amiraabouhadid (changes requested):
when i try to deploy a vm i get the following error
image

0oM4R:
@amiraabouhadid 

> when i try to deploy a vm i get the following error image

image

just deployed a vm, please make sure to build first

amiraabouhadid (dismissed):

MohamedElmdary (approved):

maayarosama:
I think it might look better if we unseleceted the selected contracts after they're unlocked as they stay selected as follows

Peek 2024-05-14 17-14

packages/grid_client/src/clients/tf-grid/contracts.ts
```diff
@@ -325,6 +335,24 @@ class TFContracts extends Contracts {
...
+    for (const id of ids) {
+      extrinsics.push(await this.unlock(id));
+    }
+    return this.client.applyAllExtrinsics(extrinsics);
+  }
+  async unlockMyContracts(graphqlURL: string) {
+    const contracts = await this.listMyContracts({
+      stateList: [ContractStates.GracePeriod],
+      graphqlURL,
+    });
```
sameh-farouk, line 349: Can we filter contracts with zero amounts locked to reduce network load and user fees?

sameh-farouk (changes requested):

packages/grid_client/src/clients/tf-grid/contracts.ts
0oM4R: the idea is we are listing contracts using gql and it doesn't return any billing info.
so we have to make another call to get billing info. this we increase the network load, what do you think?
image

Mahmoud-Emad (approved):
Good job ya 3areeees <3

packages/grid_client/src/clients/tf-grid/contracts.ts
sameh-farouk: Querying a value from the chain is generally much less computationally expensive than sending an extrinsic. When querying the storage, it's like retrieving/reading a value from the db of any of the RPC nodes, which is fast and requires no propagation across the network, multiple executions on all validators, or inclusion in a new block on the chain. On the other hand, sending extrinsic would require all of these, thus necessitating the sender to pay fees for the consumed resources.

sameh-farouk:
What are your thoughts on using the term 'unlock'? Would using something like 'resume' or 'restore' be more friendly since the related workload is paused?

amiraabouhadid:
> What are your thoughts on using the term 'unlock'? Would using something like 'resume' or 'restore' be more friendly since the related workload is paused?

resume sounds better

packages/grid_client/src/clients/tf-grid/contracts.ts
```diff
@@ -336,12 +336,19 @@ class TFContracts extends Contracts {
   }
 
   async batchUnlockContracts(ids: number[]) {
-    const extrinsics: ExtrinsicResult<number>[] = [];
+    console.log({ ids });
```
Mahmoud-Emad, line 339: Do we need to print out the ids?

packages/grid_client/src/clients/tf-grid/contracts.ts
```diff
@@ -336,12 +336,19 @@ class TFContracts extends Contracts {
...
+    const billableContractsIDs: number[] = [];
     for (const id of ids) {
+      if ((await this.contractLock({ id })).amountLocked > 0) billableContractsIDs.push(id);
+    }
+    console.log(billableContractsIDs);
```
Mahmoud-Emad, line 344: same here

Mahmoud-Emad (approved):

sameh-farouk (approved):

sameh-farouk:
Can we add a note that in some cases the actual deducted amount when contracts are resumed could be less than estimated here?

AhmedHanafy725:
> Can we add a note that in some cases the actual deducted amount when contracts are resumed could be less than estimated here?

I prefer to not add it as it's not harmful. Cause sometimes when we add many notes, the user becomes too lazy to read them.

zaelgohary (approved):
@0oM4R Good job! The feature works perfectly. as seen below.

[unlock_contract.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/40770501/a9e74c03-232a-4b4d-a3be-056fbad69a4a)

Just one comment. It might not be clear to the user that this feature exists. I suggest to add an alert before the tables which contains the ```grace period``` chip and how to unlock the contract. Create an [issue](https://github.com/threefoldtech/tfgrid-sdk-ts/issues/2760) for it.

Issue #2215, opened 2024-02-20
Add network usage in pricing calculator


Mik-TF:
## Description

Proposition: add network usage in pricing calculator, as explained here by @mohamedamer453: https://gist.github.com/mohamedamer453/6ac5cdac2fb35bbf3d1462e89c8bb1e7

### Implementation

Default value could be 0. Users can put the data usage to add in the total cost.

### More

Work in relation with the current manual [issue on info_grid](https://github.com/threefoldtech/info_grid/issues/424) by @A-Harby.

Please remove if duplicate, but I haven't seen it in any issues search nor on the devnet version of the dashboard.

maayarosama:
1. Work Completed:
Adding calculate nu method in calculator model, this method should be added to the pricing method and final calculations

PR #2657, opened 2024-05-07
Adding network usage to resource calculator

maayarosama:
### Description

Added a bandwidth field in pricing calculator 
### Changes

- Added an nu variable in calculator model
- Added  a calcNU method

### Related Issues

- #2215

### Checklist

- [ ] Tests included
- [x] Build pass
- [x] Documentation

### DIFF
--- a/packages/grid_client/src/modules/calculator.ts
+++ b/packages/grid_client/src/modules/calculator.ts
@@ -4,7 +4,7 @@ import { TFClient } from "../clients/tf-grid/client";
 import { GridClientConfig } from "../config";
 import { expose } from "../helpers/expose";
 import { validateInput } from "../helpers/validator";
-import { CalculatorModel, CUModel, SUModel } from "./models";
+import { CalculatorModel, CUModel, NUModel, SUModel } from "./models";
 
 export interface PricingInfo {
   dedicatedPrice: number;
@@ -49,6 +49,13 @@ class Calculator {
   async calSU(options: SUModel) {
     return options.hru / 1200 + options.sru / 200;
   }
+
+  @expose
+  @validateInput
+  async calNU(options: NUModel) {
+    // what is this number
+    return (options.nu * 1000) / 1e7;
+  }
   async getPrices() {
     const pricing = await this.client.pricingPolicies.get({ id: 1 });
     return pricing;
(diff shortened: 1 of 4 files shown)


AlaaElattar (approved):

packages/grid_client/src/modules/calculator.ts
```diff
@@ -49,6 +49,13 @@ class Calculator {
   async calSU(options: SUModel) {
     return options.hru / 1200 + options.sru / 200;
   }
+
+  @expose
+  @validateInput
+  async calNU(options: NUModel) {
+    // what is this number
```
zaelgohary, line 56: Please, remove this comment.

packages/playground/src/utils/pricing_calculator.ts
```diff
@@ -19,6 +19,13 @@ export const cruRules = _applyRules([
   max("Maximum allowed cpu cores is 256.", 256),
 ]);
 
+export const nuRules = _applyRules([
+  required("Bandwidth is required."),
+  isInt("Bandwidth must be a valid integer."),
+  min("Minimum allowed bandwidth is 0", 0),
```
zaelgohary, line 25:
```suggestion
  min("Minimum allowed bandwidth is 0.", 0),
```

zaelgohary (commented):
The UI seems a bit off. Can we replace the switches next to the Bandwidth field?

image

zaelgohary (changes requested):
Since network depends on public IP, I think the bandwidth field should be disabled by default and enabled only if the public IP switch is on. Also, the filed tooltip should explain this to the user.

packages/grid_client/src/modules/calculator.ts
```diff
@@ -49,6 +49,13 @@ class Calculator {
...
+    return (options.nu * 1000) / 1e7;
```
zaelgohary, line 57:
```suggestion
    return options.nu / 1e4;
```

xmonader: @zaelgohary no, keep it as * 1000 / 1e7 (this explains the flow)

xmonader:
Please move the balance field next to the bandwidth field

Issue #2684, opened 2024-05-12
Get TFT button is not working


zaelgohary:
> @zaelgohary After investigating in `get tft` button I found that issue related to this PR https://github.com/threefoldtech/tfgrid-sdk-ts/pull/2563/files#diff-fce083955bd2accb8ca8a9c35f2f4c1915853c058393a1b14e1ae26dad9aa2c0R43-R45

zaelgohary:
Work completed:

Reverted grid store changes in funds card.

khaledyoussef24:
verified and fixed on devnet and working fine now 
version : 85a6fa4



https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/af5c78f2-c5e3-4a63-b00b-d9bb65646fe1

PR #2563, opened 2024-04-15
Use grid store instead of getGrid

zaelgohary:
### Changes

- Use grid store instead of getGrid in all Dashboard pages excepts for profile
- Remove unused code

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1427

### Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation
- [x] Code format and docstrings
- [ ] Screenshots/Video attached (needed for UI changes)

PR #2686, opened 2024-05-13
Fix get TFT button

zaelgohary:
### Description

 Issue happened because of using grid store. Fixed by reverting changes in funds card and used get grid instead.

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/2684

### Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation

### DIFF
--- a/packages/playground/src/components/funds_card.vue
+++ b/packages/playground/src/components/funds_card.vue
@@ -25,23 +25,21 @@
 </template>
 
 <script lang="ts">
-import type { GridClient } from "@threefold/grid_client";
 import { ref } from "vue";
 
 import { manual } from "@/utils/manual";
 
 import { useProfileManagerController } from "../components/profile_manager_controller.vue";
-import { useGrid } from "../stores";
+import { useProfileManager } from "../stores";
 import { createCustomToast, ToastType } from "../utils/custom_toast";
-import { updateGrid } from "../utils/grid";
+import { getGrid } from "../utils/grid";
 
 export default {
   name: "FundsCard",
   setup() {
     const loadingAddTFT = ref(false);
     const ProfileManagerController = useProfileManagerController();
-    const gridStore = useGrid();
-    const grid = gridStore.client as GridClient;
+    const profileManager = useProfileManager();
 
     const addTFT = async () => {
       if (window.env.NETWORK !== "dev" && window.env.NETWORK !== "qa") {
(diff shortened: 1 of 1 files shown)


amiraabouhadid (approved):

packages/playground/src/components/funds_card.vue
```diff
@@ -25,31 +25,29 @@
...
 import { ref } from "vue";
 
 import { manual } from "@/utils/manual";
 
 import { useProfileManagerController } from "../components/profile_manager_controller.vue";
-import { useGrid } from "../stores";
+import { useProfileManager } from "../stores";
 import { createCustomToast, ToastType } from "../utils/custom_toast";
-import { updateGrid } from "../utils/grid";
+import { getGrid } from "../utils/grid";
```
Mahmoud-Emad, line 35: Why did we switch from `getGrid` to `useGrid` in [this PR](https://github.com/threefoldtech/tfgrid-sdk-ts/pull/2563) and then switch back here?

packages/playground/src/components/funds_card.vue
```diff
@@ -25,31 +25,29 @@
...
 
 export default {
   name: "FundsCard",
   setup() {
     const loadingAddTFT = ref(false);
     const ProfileManagerController = useProfileManagerController();
-    const gridStore = useGrid();
-    const grid = gridStore.client as GridClient;
+    const profileManager = useProfileManager();
```
Mahmoud-Emad, line 42: Do we use profileManagerController now instead?

Mahmoud-Emad (changes requested):
Good job ya Zainab

packages/playground/src/components/funds_card.vue
zaelgohary: Profile manager controller instead of what? The profile manager and profile manager controller have 2 different functionalities.

packages/playground/src/components/funds_card.vue
zaelgohary: We were supposed to use the store to reduce the number of grid instances in the Dashboard. This produced multiple issues like when listing deployments or when using the get TFT. That's why we're reverting back these changes.

Mahmoud-Emad: What kind of issues?

zaelgohary: It broke the functionalites.

zaelgohary: You can revert back to the mentioned PR. it has a log with the comments for more clarification.

Mahmoud-Emad: okay, I will approve but we need to take the know what should we use the `gridStore` or `getGrid`

Mahmoud-Emad (approved):

Issue #1868, opened 2024-01-03
allow sorting on the usd price in dedicated nodes


xmonader:
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/64129/0a60ce48-eed2-4c67-bfef-8fa83ef3216f)

are these prices before or after applying or discounts as well?

AhmedHanafy725:
There are 2 options to do that
1- Get all nodes and do the sorting and filtering on the client side
2- Make gridproxy do the calculations. https://github.com/threefoldtech/tfgrid-sdk-go/issues/608

AlaaElattar:
- Blocked on https://github.com/threefoldtech/tfgrid-sdk-go/issues/608

0oM4R:
blocked on #2285

A-Harby:
Devnet, 0d1e63f.

The node 161 has a value of 0.11 USD per month, and it is between 113 and 72 USD.
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/6138215e-0357-4b9a-8bc8-57e371292839)

So some cases are now covered maybe.

MohamedElmdary:
@A-Harby it's 0.11 USD/hour not month
so 0.11 * 24 *30 = 79.2 USD/month so it's correct

A-Harby:
https://github.com/threefoldtech/tfgrid-sdk-ts/issues/2773

> @A-Harby it's 0.11 USD/hour not month so 0.11 * 24 *30 = 79.2 USD/month so it's correct

khaledyoussef24:
verified on devnet
version: 17d861b


User should be able to sort nodes by price from high to low and low to high



![Image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/e44bc178-b88b-4adb-b543-a078e38f7bf2)
![Image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/990fdd3d-28be-46dc-80af-2559f9b6f2ca)

PR #2689, opened 2024-05-13
Support sorting nodes based on usd_price

MohamedElmdary:
### Description
Support sorting nodes based on usd_price on node finder

### Changes
- feat: Allow to modify height of table
- feat: Update table height to be as long as the filters
- feat: Add sorting in node finder nodes based on usd_price
image
image
image
image

### Related Issues
- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1868

### Documentation PR

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/packages/playground/src/views/nodes.vue
+++ b/packages/playground/src/views/nodes.vue
@@ -392,6 +392,17 @@
         </TfFiltersContainer>
       </template>
 
+      <v-row>
+        <v-spacer />
+        <v-col :style="{ maxWidth: '350px' }">
+          <v-select hide-details label="Sort By" clearable :items="sortItems" v-model="sortItem" return-object>
+            <template #item="{ props, item }">
+              <v-list-item v-bind="props" :title="item.title" :prepend-icon="item.raw.icon" />
+            </template>
+          </v-select>
+        </v-col>
+      </v-row>
+
       <div class="nodes">
         <div class="nodes-inner">
           <v-row>
@@ -399,6 +410,7 @@
               <div class="table">
                 <nodes-table
                   v-model="nodes"
+                  height="675px"
                   :size="size"
                   @update:size="
                     size = $event;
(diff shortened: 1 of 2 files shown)


amiraabouhadid (changes requested):
please disable sorting if dedicated nodes filter is not enabled as i got this strange result 
image

MohamedElmdary:
> please disable sorting if dedicated nodes filter is not enabled as i got this strange result image

I thought it's wrong at first but
0.11 * 30 * 24 ~= 79$ > 73$ but I think we still should disable it if dedicated not selected

packages/playground/src/views/nodes.vue
```diff
@@ -487,21 +511,18 @@ export default {
     });
 
     const loading = ref<boolean>(true);
-    const nodes = ref<GridNode[]>([]);
+    const _nodes = ref<GridNode[]>([]);
+
+    const sortItem = ref<{ title: string; icon: string; value: (nodes: GridNode[]) => GridNode[] }>();
```
Mahmoud-Emad, line 516: It is recommended to use `reactive` for objects and `ref` for primitives.

packages/playground/src/views/nodes.vue
```diff
@@ -487,21 +511,18 @@ export default {
     });
 
     const loading = ref<boolean>(true);
-    const nodes = ref<GridNode[]>([]);
+    const _nodes = ref<GridNode[]>([]);
```
Mahmoud-Emad, line 514: It is recommended to use `reactive` for objects and `ref` for primitives.

packages/playground/src/views/nodes.vue
MohamedElmdary: nothing special about reactive.

from dev io website
The significant difference between ref() and reactive() is that the ref() function allows us to declare reactive state for primitives and objects, while reactive() only declares reactive state for objects.

from vue docs (it's recommended to use `ref`)
https://vuejs.org/guide/essentials/reactivity-fundamentals#ref

amiraabouhadid (approved):

Mahmoud-Emad (approved):

Issue #1815, opened 2023-12-28
Add currency abstraction


xmonader:
We keep doing the conversions manually, and .toFixed calls way too often, let's introduce some sort of Tft class or Currency that can handle the logic of conversion, hourly, monthly, yearly calculations, and the string representation, .. etc

zaelgohary:
WIP:

Added the TFT class which contains conversions to/from USD alongside monthly and yearly conversions. Currently working on implementing the script to use the mentioned functions.

zaelgohary:
WIP:

Used decimal in the abstraction and updated the script. Currently working on adding unit tests.

zaelgohary:
WIP: 

Edited the TFT module and updated its script and added some tests. However, some of the tests are not working and I'm working on fixing them.

zaelgohary:
Work completed:

Changed the class name and removed the calculator completely and got TFT directly from TFChain instead. Applied changes in the script and its tests. Also, added more tests.

zaelgohary:
Changed the previous flow so the user would need to get the rate himself before creating a currency instance (Now the Grid Client is not needed to create the instance). Also added the possibility for the user to change the decimal value for all the method results.

PR #2707, opened 2024-05-15
Add currency abstraction

zaelgohary:
### Changes

- Add currency abstraction
- Add currency script
- Add currency unit tests
- Edit validateInput fn to return sync code instead of async

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1815

### Checklist

- [x] Tests included
- [x] Build pass
- [ ] Documentation


packages/grid_client/src/modules/models.ts
```diff
@@ -832,6 +832,14 @@ class GetActiveContractsModel {
   @Expose() @IsInt() @IsNotEmpty() @Min(1) nodeId: number;
 }
 
+class USDModel {
```
xmonader, line 835: Replace both with CurrencyModel with amount field

packages/grid_client/src/modules/tft.ts
```diff
@@ -0,0 +1,53 @@
...
+  QA = "wss://tfchain.qa.grid.tf/ws",
+  TEST = "wss://tfchain.test.grid.tf/ws",
+  MAIN = "wss://tfchain.grid.tf/ws",
+}
+
+class TFT {
+  public SUBSTRATE_URL: string;
+  private calculator: Calculator;
+
+  constructor(SUBSTRATE_URL: string) {
```
xmonader, line 19: You either pass a price or an object that you can get the price from, your code shouldn't know anything about substrate URL or how to initialize the calculator

xmonader: with this both can be easily injected/mocked

packages/grid_client/src/modules/tft.ts
```diff
@@ -0,0 +1,53 @@
...
+import { TFTModel, USDModel } from "./models";
+
+enum substrateURL {
+  DEV = "wss://tfchain.dev.grid.tf/ws",
+  QA = "wss://tfchain.qa.grid.tf/ws",
+  TEST = "wss://tfchain.test.grid.tf/ws",
+  MAIN = "wss://tfchain.grid.tf/ws",
+}
+
+class TFT {
```
xmonader, line 15: Let's make it TFTUSDConversionService

packages/grid_client/src/modules/tft.ts
```diff
@@ -0,0 +1,53 @@
...
+    this.calculator = new Calculator(new QueryClient(this.SUBSTRATE_URL));
+  }
+
+  async price() {
+    return await this.calculator.tftPrice();
+  }
+
+  @expose
+  @validateInput
+  async fromUSD(options: USDModel) {
```
xmonader, line 30: that's any object that implements `{amount: } `, I don't feel we need USDMdodel, but will see through the review 

Why didn't you add the return type? let's change the name to `convertUSDtoTFT`

xmonader: no need for async, make this code pure without side effect so either pass the price service or the price value and set it on the conversion service (and specify the direction that is tftInOneUSD for instance

packages/grid_client/src/modules/tft.ts
```diff
@@ -0,0 +1,53 @@
...
+    return new Decimal(options.usd / (await this.price())).toFixed(2);
+  }
+
+  @expose
+  @validateInput
+  async toUSD(options: TFTModel) {
```
xmonader, line 36: rename to convertTFTtoUSD, add return type, and apply the comments on async from before

packages/grid_client/src/modules/tft.ts
```diff
@@ -0,0 +1,53 @@
...
+    return new Decimal(options.tft * (await this.price())).toFixed(2);
+  }
+  @expose
+  @validateInput
+  toMonth(options: TFTModel) {
```
xmonader, line 41: change to `monthlyTFT` and change the parameter name to `hourlyTFT` and don't miss the return type

packages/grid_client/src/modules/tft.ts
```diff
@@ -0,0 +1,53 @@
...
+  DEV = "wss://tfchain.dev.grid.tf/ws",
+  QA = "wss://tfchain.qa.grid.tf/ws",
+  TEST = "wss://tfchain.test.grid.tf/ws",
+  MAIN = "wss://tfchain.grid.tf/ws",
+}
+
+class TFT {
+  public SUBSTRATE_URL: string;
+  private calculator: Calculator;
+
```
xmonader, line 18: add field to represent number of decimals to be passed to the toFixed

packages/grid_client/src/modules/tft.ts
xmonader: and in the comments point out that it will be calculated in 30 days and let's add dailyTFT

packages/grid_client/src/modules/tft.ts
```diff
@@ -0,0 +1,53 @@
...
+  }
+  @expose
+  @validateInput
+  toMonth(options: TFTModel) {
+    return new Decimal(options.tft * 24 * 30).toFixed(2);
+  }
+
+  @expose
+  @validateInput
+  toYear(options: TFTModel) {
```
xmonader, line 47: same comments as in in toMonth but for a year

packages/grid_client/scripts/tft.ts
```diff
@@ -0,0 +1,28 @@
+import { substrateURL, TFT } from "../src";
```
Mahmoud-Emad, line 1: Please check the other scripts and let's do the same here, we need to define a main function, separate each block into its function, and use a block logger

packages/grid_client/src/modules/models.ts
Mahmoud-Emad: let's provide a docString for each newly implemented `class` and `methods`, and try to describe what it can do, what parameters are needed, and what it returns

packages/grid_client/scripts/tft.ts
```diff
@@ -0,0 +1,28 @@
...
+import { getClient } from "./client_loader";
+
+const substrate = "wss://tfchain.dev.grid.tf/ws" as substrateURL.DEV;
```
Mahmoud-Emad, line 4: why do you redefine it again?

```ts
const substrate = substrateURL.DEV;
```

packages/grid_client/tests/modules/tft.test.ts
```diff
@@ -0,0 +1,98 @@
...
+  await grid.disconnect();
+});
+
+describe("Testing TFT module", () => {
+  test("Create TFT instance in all networks.", () => {
+    // Test Data
+    const SUBSTRATE_DEV = "wss://tfchain.dev.grid.tf/ws";
+    const SUBSTRATE_QA = "wss://tfchain.qa.grid.tf/ws";
+    const SUBSTRATE_TEST = "wss://tfchain.test.grid.tf/ws";
+    const SUBSTRATE_MAIN = "wss://tfchain.grid.tf/ws";
```
Mahmoud-Emad, line 37: if you already created the enum, you can use it here instead of re-creating the values

Mahmoud-Emad (changes requested):
Good job ya Zainab :)

packages/grid_client/tests/modules/tft.test.ts
zaelgohary: Removed it as it's not used anywhere.

packages/grid_client/src/modules/tft.ts
zaelgohary: Removed the calculator completely and changed the constructor to get the grid client config instead then I get the price from the tfchain directly.

packages/grid_client/scripts/tft.ts
zaelgohary: Edited.

packages/grid_client/scripts/tft.ts
zaelgohary: Not needed anymore. Please, check the changes.

packages/grid_client/src/modules/tft.ts
zaelgohary: Added.

Mahmoud-Emad:
Please resolve the conflicts

zaelgohary:
> Please resolve the conflicts

Resolved.

AlaaElattar (approved):
Great Job ya zuzu

packages/grid_client/src/modules/tft.ts
```diff
@@ -0,0 +1,83 @@
...
+import { GridClientConfig, TFClient } from "..";
+import { expose, validateInput } from "../helpers";
+import { CurrencyModel } from "./models";
+
+class TFTUSDConversionService {
+  private decimals = 2;
+  private tfclient: TFClient;
+  TFTPrice: number;
+
+  constructor(public config: GridClientConfig) {
```
xmonader, line 12: should take in the decimals and default it to 2

packages/grid_client/src/modules/tft.ts
```diff
@@ -0,0 +1,82 @@
...
+import { GridClientConfig, TFClient } from "..";
+import { expose, validateInput } from "../helpers";
+import { CurrencyModel } from "./models";
+
+class TFTUSDConversionService {
+  private tfclient: TFClient;
+  private rate: number;
+
+  constructor(public config: GridClientConfig, public decimals = 2) {
+    this.decimals = decimals;
```
xmonader, line 12: you don't need this

packages/grid_client/src/modules/tft.ts
```diff
@@ -0,0 +1,82 @@
...
+    this.tfclient = config.tfclient;
+    this.tfclient.tftPrice.get().then(res => {
```
xmonader, line 14: what happens in case of error?

zaelgohary: Removed the client completely.

packages/grid_client/src/modules/tft.ts
```diff
@@ -1,20 +1,11 @@
...
-
-  constructor(public config: GridClientConfig, public decimals = 2) {
-    this.decimals = decimals;
-    this.tfclient = config.tfclient;
-    this.tfclient.tftPrice.get().then(res => {
-      this.rate = res;
-    });
-  }
+  // TFT rate: 1 tft = x USD
+  constructor(public rate: number, private decimals = 2) {}
```
xmonader, line 8: I'd also protect the rate and allow its access via a getter too

xmonader: consider that to be the last comment resolve and merge once done 💯

xmonader (approved):

Issue #2485, opened 2024-03-20
🐞 [Bug]: Can't delete big number of contracts with one batch


AhmedHanafy725:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### which package/s did you face the problem with?

tfchain_client

### What happened?

when I tried to delete 1902 contracts using [delete my contract script](https://github.com/threefoldtech/tfgrid-sdk-ts/blob/development/packages/grid_client/scripts/delete_all_contracts.ts), it failed

### Steps To Reproduce

1- create 2000 contracts
2- try to delete them with the above script.

### which network/s did you face the problem on?

Dev

### version

development

### Relevant screenshots/screen records

-

### Relevant log output

```console
RpcError: 1010: Invalid Transaction: Transaction would exhaust the block limits
    at /sandbox/code/github/threefoldtech/tfgrid-sdk-ts/packages/tfchain_client/dist/node/client.js:363:23
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async TFClient.applyExtrinsic (/sandbox/code/github/threefoldtech/tfgrid-sdk-ts/packages/tfchain_client/dist/node/client.js:372:22)
    at async TFContracts.batchCancelContracts (/sandbox/code/github/threefoldtech/tfgrid-sdk-ts/packages/grid_client/src/clients/tf-grid/contracts.ts:325:5)
    at async TFContracts.cancelMyContracts (/sandbox/code/github/threefoldtech/tfgrid-sdk-ts/packages/grid_client/src/clients/tf-grid/contracts.ts:316:5)
    at async Contracts.cancelMyContracts (/sandbox/code/github/threefoldtech/tfgrid-sdk-ts/packages/grid_client/src/modules/contracts.ts:232:23)
    at async Contracts.descriptor.value (/sandbox/code/github/threefoldtech/tfgrid-sdk-ts/packages/grid_client/src/modules/utils.ts:11:12)
    at async Contracts.descriptor.value (/sandbox/code/github/threefoldtech/tfgrid-sdk-ts/packages/grid_client/src/helpers/validator.ts:22:12)
    at async deleteAllContracts (/sandbox/code/github/threefoldtech/tfgrid-sdk-ts/packages/grid_client/scripts/delete_all_contracts.ts:5:15)
    at async main (/sandbox/code/github/threefoldtech/tfgrid-sdk-ts/packages/grid_client/scripts/delete_all_contracts.ts:15:3) {
[...]

PR #2728, opened 2024-05-19
Adding a batchSize and applying extrinsics per batch

maayarosama:
### Description

Adding a batch size to batchCancelContracts method. so instead of canceling all contracts at once, they get canceled in batches 
### Changes

- Adding a batch size and applyAllExtrinsics per batch
### Related Issues

- #2485

### Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation


AhmedHanafy725 (changes requested):
This should be applied to all extrinsic methods not only canceling contracts. I suggest it to be done in utility methods (batch, batchAll) [here](https://github.com/threefoldtech/tfgrid-sdk-ts/blob/development/packages/tfchain_client/src/utility.ts)

packages/grid_client/src/clients/tf-grid/contracts.ts
```diff
@@ -323,6 +323,7 @@ class TFContracts extends Contracts {
       extrinsics.push(await this.cancel({ id }));
     }
     await this.client.applyAllExtrinsics(extrinsics);
+
```
MohamedElmdary, line 326: @maayarosama  remove this file as it got no changes

packages/tfchain_client/src/utility.ts
```diff
@@ -2,6 +2,7 @@ import { Client } from "./client";
 import { ExtrinsicResult } from "./types";
 import { checkConnection } from "./utils";
 
+const batchSize = 400;
```
AhmedHanafy725, line 5: constants should be uppercase
```suggestion
const BATCH_SIZE = 400;

```

packages/tfchain_client/src/utility.ts
```diff
@@ -13,9 +14,16 @@ class Utility {
...
-      const { resultSections, resultEvents } = this.extractResultSectionsAndEvents(extrinsics);
-      const batchExtrinsic = await this.client.api.tx.utility.batch(extrinsics);
-      return this.client.applyExtrinsic<T[]>(batchExtrinsic, resultSections, resultEvents);
+      const result: T[] = [];
+      for (let i = 0; i < extrinsics.length; i += batchSize) {
+        const batch = extrinsics.slice(i, i + batchSize);
+        const { resultSections, resultEvents } = this.extractResultSectionsAndEvents(batch);
+        const batchExtrinsic = await this.client.api.tx.utility.batch(batch);
+        const res = await this.client.applyExtrinsic<T[]>(batchExtrinsic, resultSections, resultEvents);
+        result.concat(res);
```
AhmedHanafy725, line 23: we should overwrite the variable again as `concat` returns the concatenation result and doesn't change the list.
image

```suggestion
        result = result.concat(res);

```

AhmedHanafy725 (changes requested):

AhmedHanafy725 (approved):

amiraabouhadid (approved):

Issue #1841, opened 2024-01-01
Test Automation for All available solutions on the new dashboard


ramezsaeed:
Add automated tests for all solutions in dashboard.
- Tests must include ssh to the created solution.
- successful login if the solution has UI interface

khaledyoussef24:
Wip
- added a new solutions page to the selenium 
- Added a separate page for solutions (full vm and micro vm)
- On going automated test for deploying still working on verifying the deployment and asserting then will go to deleting.

xmonader:
it should be scripted first (and added to integration tests)

khaledyoussef24:
WIP 
started in two automated tests but stuck in making them passing
(peertube , funk whale)

khaledyoussef24:
work done:
- added scripts to all of the 12 soltuions 
- domain generating still needs some fixing (tested on public ips)
- finished 5 solutions but could not test them freely due to lack of public ips and the planetary ip current issue and development to use mycellium.

WIP:
finishing the rest of the tests.
adding to ping the solutions through ip with time out to assert that the websites made.
tests left to be done

- [ ] owncloud
- [ ] presearch
- [ ] subsquid
- [ ] casperlabs
- [ ] node pilot
- [ ] umbrel
- [ ] wordpress

khaledyoussef24:
the new branch is ready for review 
solved all the conflicts

khaledyoussef24:
work done :
- added a http request to the test to check if the gateway deployed right.
- added mycelium flag to the tests and scripts.
- the pr is ready for review.

khaledyoussef24:
tried scripts and working fine.
tests don't work with consistency, but still work.
 PR is ready for review.

khaledyoussef24:
some changes are made so scripts needs to be updated.
comments are resolved but still searching why scripts are not working.

khaledyoussef24:
Work done
scripts for solutions are added.
the vms are working fine with flists waiting to be tested with public ips then will move the pr from drafts.
(created a new branch because the old branch had so many conflicts.)
will continue the tests after the pr gets merged.

khaledyoussef24:
estimated time for the tests 3 days after the test run of testnet andthe script branch gets merged.

A-Harby:
https://github.com/threefoldtech/tfgrid-sdk-ts/pull/2743

A-Harby:
> #2743

[...]

Issue #2580, opened 2024-04-18
🐞 [Bug]: can not access funkwhale vm while the solution is already working


khaledyoussef24:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### which package/s did you face the problem with?

Dashboard

### What happened?

deployed a funkwhale vm but could not access it using mycelium or yggdrasil ip

### which network/s did you face the problem on?

Dev, Main

### version

latest on each network

### Relevant screenshots/screen records

![Screenshot 2024-04-18 at 11 50 18 AM](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/e80bd969-81db-4c46-b8d8-3471ed008f46)
it shows the trial of both ips (mycelium binary is running in the background)
![Screenshot 2024-04-18 at 11 51 34 AM (2)](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/36942d20-9a4c-447a-a7b2-e25d8466aed8)

### Relevant log output

```console
na
```

Mahmoud-Emad:
Blocked on https://github.com/threefoldtech/tf_operations/issues/2473

maayarosama:
###Work In Progress
- The flist used for funkwhale is the old That doesn't support ssh
- I tried deploying a machine using the new flist, it gives me bad gateway and I can't ssh
![Screenshot from 2025-01-14 17-22-27](https://github.com/user-attachments/assets/65407595-6b7e-4ed9-8d94-5fc6bcd3acf8)
![Screenshot from 2025-01-14 17-22-45](https://github.com/user-attachments/assets/c35dac56-091a-4882-a7b1-80b41bd1cf9a)
![Screenshot from 2025-01-14 17-23-07](https://github.com/user-attachments/assets/2ed2eb30-c81a-43db-a1f2-e2fcc9e89de7)

maayarosama:
### Work In Progress
- The flist used for funkwhale is the old That doesn't support ssh
- I tried deploying a machine using the new flist, it gives me bad gateway and I can't ssh
![Screenshot from 2025-01-14 17-22-27](https://github.com/user-attachments/assets/65407595-6b7e-4ed9-8d94-5fc6bcd3acf8)
![Screenshot from 2025-01-14 17-22-45](https://github.com/user-attachments/assets/c35dac56-091a-4882-a7b1-80b41bd1cf9a)
[...]

PR #2743, opened 2024-05-20
Add the peertube, mattermost, funkwhale, and casperlabs scripts to deploy an instance from the GridClient.

A-Harby:
### Description

Add the scripts for the applications (peertube, mattermost, funkwhale, and casperlabs).

The scripts are ready and working. 

### Related Issues

https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1841

### Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation

### DIFF
--- a/packages/grid_client/scripts/applications/casberlabs.ts
+++ b/packages/grid_client/scripts/applications/casberlabs.ts
@@ -46,12 +46,13 @@ async function main() {
   const name = "newcasperlabs";
   const grid3 = await getClient(`casperlabs/${name}`);
   const subdomain = "cl" + grid3.twinId + name;
+  const instanceCapacity = { cru: 1, mru: 2, sru: 15 }; // Update the instance capacity values according to your requirements.
 
   //VMNode Selection
   const vmQueryOptions: FilterOptions = {
-    cru: 2,
-    mru: 4,
-    sru: 100,
+    cru: instanceCapacity.cru,
+    mru: instanceCapacity.mru,
+    sru: instanceCapacity.sru,
     availableFor: grid3.twinId,
     farmId: 1,
   };
@@ -78,7 +79,7 @@ async function main() {
         disks: [
           {
             name: "wedDisk",
-            size: 100,
+            size: instanceCapacity.sru,
             mountpoint: "/data",
           },
         ],
@@ -86,8 +87,8 @@ async function main() {
         public_ip: false,
         public_ip6: false,
         mycelium: false,
-        cpu: 2,
-        memory: 1024 * 4,
+        cpu: instanceCapacity.cru,
+        memory: 1024 * instanceCapacity.mru,
         rootfs_size: 0,
         flist: "https://hub.grid.tf/tf-official-apps/casperlabs-latest.flist",
         entrypoint: "/sbin/zinit init",
--- a/packages/grid_client/scripts/applications/casberlabs.ts
+++ b/packages/grid_client/scripts/applications/casberlabs.ts
@@ -8,40 +8,41 @@ async function pingNodes(client, nodes) {
       await client.zos.pingNode({ nodeId: node.nodeId });
       return node.nodeId;
     } catch (error) {
-      console.log("node " + node.nodeId + " is not responding, trying different node.");
+      throw new Error("node " + node.nodeId + " is not responding, trying different node.");
     }
   }
   throw new Error("No avaiable nodes");
 }
 
-async function deploy(client, vms) {
-  const res = await client.machines.deploy(vms);
+async function deploy(client, vms, subdomain, gatewayNode) {
+  const resultVM = await client.machines.deploy(vms);
   log("================= Deploying VM =================");
-  log(res);
+  log(resultVM);
   log("================= Deploying VM =================");
-}
 
-async function deployGW(client, gw) {
-  const res = await client.gateway.deploy_name(gw);
+  const vmPlanetary = (await client.machines.getObj(vms.name))[0].planetary;
+  //Name Gateway Model
+  const gw: GatewayNameModel = {
+    name: subdomain,
+    node_id: gatewayNode.nodeId,
+    tls_passthrough: false,
+    backends: ["http://[" + vmPlanetary + "]:80"],
+  };
+
+  const resultGateway = await client.gateway.deploy_name(gw);
   log("================= Deploying name gateway =================");
-  log(res);
+  log(resultGateway);
   log("================= Deploying name gateway =================");
 }
 
-async function getDeployment(client, vms) {
-  const res = await client.machines.getObj(vms);
+async function getDeployment(client, vms, gw) {
+  const resultVM = await client.machines.getObj(vms.name);
+  const resultGateway = await client.gateway.getObj(gw);
   log("================= Getting deployment information =================");
-  log(res);
+  log(resultVM);
+  log(resultGateway);
+  log("https://" + resultGateway[0].domain);
   log("================= Getting deployment information =================");
-  return res;
-}
-
-async function getDeploymentGW(client, gw) {
-  const res = await client.gateway.getObj(gw);
-  log("================= Getting deployment information =================");
-  log(res);
-  log("================= Getting deployment information =================");
-  return res;
 }
 
 async function cancel(client, vms, gw) {
@@ -110,28 +111,10 @@ async function main() {
   };
 
   //Deploy VMs
-  await deploy(grid3, vms);
+  await deploy(grid3, vms, subdomain, gatewayNode);
 
   //Get the deployment
-  const result = await getDeployment(grid3, name);
-
-  //Name Gateway Model
-  const gw: GatewayNameModel = {
-    name: subdomain,
-    node_id: gatewayNode.nodeId,
-    tls_passthrough: false,
-    backends: ["http://[" + result[0].planetary + "]:80"],
-  };
-
-  //Deploy gateway
-  await deployGW(grid3, gw);
-
-  //Get the deployment
-  const gatewayResult = await getDeploymentGW(grid3, gw.name);
-
-  //Get the site link
-  const domain = "https://" + gatewayResult[0].domain;
-  log(domain);
+  await getDeployment(grid3, vms, subdomain);
 
   //Uncomment the line below to cancel the deployment
   // await cancel(grid3, { name }, { name: subdomain });
--- a/packages/grid_client/scripts/applications/casberlabs.ts
+++ b/packages/grid_client/scripts/applications/casberlabs.ts
@@ -46,7 +46,7 @@ async function main() {
   const name = "newcasperlabs";
   const grid3 = await getClient(`casperlabs/${name}`);
   const subdomain = "cl" + grid3.twinId + name;
-  const instanceCapacity = { cru: 1, mru: 2, sru: 15 }; // Update the instance capacity values according to your requirements.
+  const instanceCapacity = { cru: 2, mru: 4, sru: 100 }; // Update the instance capacity values according to your requirements.
 
   //VMNode Selection
   const vmQueryOptions: FilterOptions = {
--- a/packages/grid_client/scripts/applications/casberlabs.ts
+++ b/packages/grid_client/scripts/applications/casberlabs.ts
@@ -1,0 +1,142 @@
+import { FilterOptions, GatewayNameModel, MachinesModel } from "../../src";
+import { config, getClient } from "../client_loader";
+import { log } from "../utils";
+
+async function pingNodes(client, nodes) {
+  for (const node of nodes) {
+    try {
+      await client.zos.pingNode({ nodeId: node.nodeId });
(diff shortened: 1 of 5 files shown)


Mahmoud-Emad:
After running the script, I got this error 
image

the error message: `Node 11 doesn't have enough resources: sru=50, mru=2 .` while you filter with the required resources, is this issue reported?

packages/grid_client/scripts/applications/casberlabs.ts
```diff
@@ -0,0 +1,142 @@
...
+import { config, getClient } from "../client_loader";
+import { log } from "../utils";
+
+async function pingNodes(client, nodes) {
+  for (const node of nodes) {
+    try {
+      await client.zos.pingNode({ nodeId: node.nodeId });
+      return node.nodeId;
+    } catch (error) {
+      console.log("node " + node.nodeId + " is not responding, trying different node.");
```
Mahmoud-Emad, line 11: throw error instead of logging

packages/grid_client/scripts/applications/casberlabs.ts
```diff
@@ -0,0 +1,142 @@
...
+  return res;
+}
+
+async function getDeploymentGW(client, gw) {
+  const res = await client.gateway.getObj(gw);
+  log("================= Getting deployment information =================");
+  log(res);
+  log("================= Getting deployment information =================");
+  return res;
+}
```
Mahmoud-Emad, line 45: as you did in `cancel` you delete both in a single function, you can do the same here and in deploy, you can pass a param that tells you what unit you want to deploy/get

packages/grid_client/scripts/applications/casberlabs.ts
```diff
@@ -0,0 +1,142 @@
...
+  const vms: MachinesModel = {
+    name,
+    network: {
+      name: "wedtest",
+      ip_range: "10.249.0.0/16",
+    },
+    machines: [
+      {
+        name: "casperlabs",
+        node_id: +(await pingNodes(grid3, await grid3.capacity.filterNodes(vmQueryOptions))),
```
Mahmoud-Emad, line 85: adding the `(await pingNodes(grid3, await grid3.capacity.filterNodes(vmQueryOptions)))` to another file will be better

```suggestion
        const nodeID = await pingNodes(grid3, await grid3.capacity.filterNodes(vmQueryOptions))
```

Mahmoud-Emad (changes requested):
Good job ya 7arbii

A-Harby:
> After running the script, I got this error
> the error message: `Node 11 doesn't have enough resources: sru=50, mru=2 .` while you filter with the required resources, is this issue reported?

https://github.com/threefoldtech/tfgrid-sdk-ts/issues/2750

packages/grid_client/scripts/applications/casberlabs.ts
A-Harby: Throwing the error will stop the script from running on the first unresponsive node (which happens a lot), and it's kind of neglecting the try and catch.

A-Harby:
> Good job ya 7arbii

@Mahmoud-Emad I updated the code as requested and tried to match resource on every application with one used on dashboard.
And also add pingnode function to utils.

packages/grid_client/scripts/utils.ts
```diff
@@ -3,4 +3,17 @@ import { inspect } from "util";
 function log(message) {
   console.log(inspect(message, { showHidden: false, depth: null, colors: true }));
 }
-export { log };
+
+async function pingNodes(client, nodes) {
```
Mahmoud-Emad, line 7: Please define the types of `client` and `nodes`, also, would be better if we have a `DocString` for any new function/class created to provide the user/developer some details about what is the actual job that the function does, what params it takes and what it returns.

packages/grid_client/scripts/applications/peertube.ts
```diff
@@ -0,0 +1,115 @@
...
+  const vms: MachinesModel = {
+    name,
+    network: {
+      name: "wedtest",
+      ip_range: "10.249.0.0/16",
+    },
+    machines: [
+      {
+        name: "peertube",
+        node_id: +(await pingNodes(grid3, await grid3.capacity.filterNodes(vmQueryOptions))),
```
Mahmoud-Emad, line 74: Instead of calling the function inside the object, what do you think if we stashed the value of it in a const and then used it inside the object? I think it'll enhance the code's readability.

if you agree with me, please change it in all of your scripts :)


```ts
const options = {...}
const nodes = await grid3.capacity.filterNodes(options)
const selectedNodeID = await pingNodes(nodes)

// then use it 

machines: [
  {
     node_id: selectedNodeID
  }
]
```

packages/grid_client/scripts/applications/peertube.ts
```diff
@@ -0,0 +1,115 @@
...
+        cpu: 1,
+        memory: 1024 * 2,
+        rootfs_size: 0,
+        flist: "https://hub.grid.tf/tf-official-apps/peertube-v3.1.1.flist",
+        entrypoint: "/sbin/zinit init",
+        env: {
+          SSH_KEY: config.ssh_key,
+          PEERTUBE_WEBSERVER_HOSTNAME: subdomain + "." + gatewayNode.publicConfig.domain,
+          PEERTUBE_ADMIN_EMAIL: "admin123@peer.tube", // edit me
+          PT_INITIAL_ROOT_PASSWORD: "admin123", // edit me
```
Mahmoud-Emad, line 95: The `edit me` isn't a good way to tell the user that you can edit this section if you want tbh, instead, let's give the user a clear comment.

```ts
          // edit me
```

packages/grid_client/scripts/utils.ts
```diff
@@ -3,4 +3,17 @@ import { inspect } from "util";
...
+  for (const node of nodes) {
+    try {
+      await client.zos.pingNode({ nodeId: node.nodeId });
+      return node.nodeId;
+    } catch (error) {
+      log("node " + node.nodeId + " is not responding, trying different node.");
+    }
+  }
+  throw new Error("No avaiable nodes");
```
Mahmoud-Emad, line 16:
```suggestion
  throw new Error("There are no available nodes. Please try changing the filters you have applied.");
```

packages/grid_client/scripts/utils.ts
```diff
@@ -3,4 +3,17 @@ import { inspect } from "util";
 function log(message) {
   console.log(inspect(message, { showHidden: false, depth: null, colors: true }));
 }
-export { log };
+
+async function pingNodes(client, nodes) {
+  for (const node of nodes) {
+    try {
+      await client.zos.pingNode({ nodeId: node.nodeId });
+      return node.nodeId;
```
Mahmoud-Emad, line 11: if you prefer to return the `nodeID` instead of the node interface, please define it in the function return type.

packages/grid_client/scripts/applications/peertube.ts
```diff
@@ -0,0 +1,115 @@
...
+  log("================= Getting deployment information =================");
+  log(resultVM);
+  log(resultGateway);
+  log("https://" + resultGateway[0].domain);
+  log("================= Getting deployment information =================");
+}
+
+async function cancel(client, vms, gw) {
+  const resvm = await client.machines.delete(vms);
+  const resgw = await client.gateway.delete_name(gw);
```
Mahmoud-Emad, line 38:
```suggestion
  const resultVM = await client.machines.delete(vms);
  const resultGateway = await client.gateway.delete_name(gw);
```

packages/grid_client/scripts/applications/mattermost.ts
```diff
@@ -0,0 +1,121 @@
...
+        rootfs_size: 0,
+        flist: "https://hub.grid.tf/tf-official-apps/mattermost-latest.flist",
+        entrypoint: "/sbin/zinit init",
+        env: {
+          SSH_KEY: config.ssh_key,
+          MATTERMOST_DOMAIN: subdomain + "." + gatewayNode.publicConfig.domain,
+          SITE_URL: "https://" + subdomain + "." + gatewayNode.publicConfig.domain,
+          DJANGO_SUPERUSER_EMAIL: "admin123@matter.most", // edit me
+          DB_PASSWORD: "admin123", // edit me
+          // Optional
```
Mahmoud-Emad, line 97:
```suggestion
          // Optional, if you would like to enable the access of the `SMTP Service`, you need to send these values.
```

Mahmoud-Emad (changes requested):
Small comments :)

A-Harby:
> Small comments :)

I made the requested changes, could you please review it again?

packages/grid_client/scripts/utils.ts
```diff
@@ -1,6 +1,32 @@
...
+   * Attempts to ping a list of nodes sequentially and returns the ID of the first responsive node.
+   *
+   * This function iterates through an array of nodes, pinging each one to check its responsiveness.
+   * It returns the ID of the first node that responds successfully. If no nodes respond, an error is thrown.
+   *
+   * @param {GridClient} client - An instance of the GridClient used to ping the nodes.
+   * @param {NodeInfo[]} nodes - An array of node information objects.
+   * @returns {Promise<number>} - A promise that resolves to the nodeId of the first responsive node.
+   * @throws {Error} - Throws an error if none of the nodes respond successfully.
+   */
```
Mahmoud-Emad, line 20: Move it after creating the function
```suggestion
  /**
   * Attempts to ping a list of nodes sequentially and returns the ID of the first responsive node.
   *
   * This function iterates through an array of nodes, pinging each one to check its responsiveness.
   * It returns the ID of the first node that responds successfully. If no nodes respond, an error is thrown.
   *
   * @param {GridClient} client - An instance of the GridClient used to ping the nodes.
   * @param {NodeInfo[]} nodes - An array of node information objects.
   * @returns {Promise<number>} - A promise that resolves to the nodeId of the first responsive node.
   * @throws {Error} - Throws an error if none of the nodes respond successfully.
   */
   async function pingNodes(client: GridClient, nodes: NodeInfo[]): Promise<number> {
```

Mahmoud-Emad (changes requested):

zaelgohary (commented):
@A-Harby Really nice work! Just one comment. I suggest you downgrade the resources a bit. Maybe we can use the small package ```(CPU: 1 vCores, Memory: 2 GB, SSD: 15 GB)``` as there are not enough resources in devnet ATM. We can add the previous values in an object and use them in the vms instead of using static data. This applies to all solutions in the PR.

A-Harby:
> @A-Harby Really nice work! Just one comment. I suggest you downgrade the resources a bit. Maybe we can use the small package `(CPU: 1 vCores, Memory: 2 GB, SSD: 15 GB)` as there are not enough resources in devnet ATM. We can add the previous values in an object and use them in the vms instead of using static data. This applies to all solutions in the PR.

I will add the capacity object; I really liked the idea of changing it only in one place, but for the values, I used the same values as the small capacity package for every solution, like: https://www.manual.grid.tf/documentation/dashboard/solutions/funkwhale.html.

I think it would be better to follow the recommended capacity package for every solution, so what do you think?

zaelgohary:
> I think it would be better to follow the recommended capacity package for every solution, so what do you think?

Even better. Please, do.

zaelgohary (dismissed):
Good job, @A-Harby!

zaelgohary:
I tried to ssh to the created Mattermost instance but I couldn't.

image

image

image


Also, I couldn't find the instance listed in the node contracts although it's listed in the mattermost deployment table.

image

packages/grid_client/scripts/applications/funkwhale.ts
```diff
@@ -0,0 +1,121 @@
...
+  log(resultVM);
+  log(resultGateway);
+  log("================= Canceling the deployment =================");
+}
+
+async function main() {
+  const name = "newfunkwhale";
+  const grid3 = await getClient(`funkwhale/${name}`);
+  const subdomain = "fw" + grid3.twinId + name;
+  const instanceCapacity = { cru: 1, mru: 2, sru: 15 }; // Update the instance capacity values according to your requirements.
```
zaelgohary, line 49: Smallest package in Funkwhale uses 50 sru.

image

zaelgohary (commented):
In case of Funkwhale, I was able to access the domain and login successfully. However, I couldn't ssh to the instance.

image

image

image

A-Harby:
1. For the Funkwhale SSH issue, it's an actual issue from the image itself, and it's reported [here](https://github.com/threefoldtech/tfgrid-sdk-ts/issues/2580). 

2. For listing contracts, it took a while to list, but it worked with me on both mattermost and Funkwhale.
image

3. For Mattermost SSH, it worked fine.
image

Mahmoud-Emad (approved):
Thank you Harby, Go ahead :)

zaelgohary (approved):
LGTM

PR #2784, opened 2024-05-27
Add peertube, funkwhale and casperlabs automated tests to the grid client using jest

A-Harby:
### Description

Add peertube, funkwhale and casperlabs  automated test to the grid client using Jest.

### Related Issues

https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1841

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/packages/grid_client/tests/modules/applications/casperlabs.test.ts
+++ b/packages/grid_client/tests/modules/applications/casperlabs.test.ts
@@ -5,12 +5,14 @@ import { FilterOptions, GatewayNameModel, generateString, GridClient, MachinesMo
 import { config, getClient } from "../../client_loader";
 import { bytesToGB, generateInt, getOnlineNode, log, splitIP } from "../../utils";
 
+jest.setTimeout(900000);
+
 let gridClient: GridClient;
 let deploymentName: string;
 
 beforeAll(async () => {
   gridClient = await getClient();
-  deploymentName = "cl" + generateString(15);
+  deploymentName = "cl" + generateString(10);
   gridClient.clientOptions.projectName = `casperlabs/${deploymentName}`;
   gridClient._connect();
   return gridClient;
--- a/packages/grid_client/tests/modules/applications/casperlabs.test.ts
+++ b/packages/grid_client/tests/modules/applications/casperlabs.test.ts
@@ -68,7 +68,7 @@ test("TC2683 - Applications: Deploy Casperlabs", async () => {
     farmId: 1,
     availableFor: await gridClient.twins.get_my_twin_id(),
   } as FilterOptions);
-  if (gatewayNodes.length == 0) throw new Error("no nodes available to complete this test");
+  if (gatewayNodes.length == 0) throw new Error("no gateway nodes available to complete this test");
   const GatewayNode = gatewayNodes[generateInt(0, gatewayNodes.length - 1)];
 
   //Node Selection
@@ -193,10 +193,16 @@ test("TC2683 - Applications: Deploy Casperlabs", async () => {
     const wait = await setTimeout(5000, "Waiting for gateway to be ready");
     log(wait);
 
-    axios
+    await axios
       .get(site)
-      .then(() => {
+      .then(res => {
         log("gateway is reachable");
+        log(res.status);
+        log(res.statusText);
+        log(res.data);
+        expect(res.status).toBe(200);
+        expect(res.statusText).toBe("OK");
+        expect(res.data).toContain("Your Casper node is now running succesfully on the ThreeFold Grid.");
         reachable = true;
       })
       .catch(() => {
@@ -204,20 +210,9 @@ test("TC2683 - Applications: Deploy Casperlabs", async () => {
       });
     if (reachable) {
       break;
+    } else if (i == 180) {
+      throw new Error("Gateway is unreachable after multiple retries");
     }
-  }
-
-  if (reachable) {
-    axios.get(site).then(res => {
-      log(res.status);
-      log(res.statusText);
-      log(res.data);
-      expect(res.status).toBe(200);
-      expect(res.statusText).toBe("OK");
-      expect(res.data).toContain("Your Casper node is now running succesfully on the ThreeFold Grid.");
-    });
-  } else {
-    throw new Error("Gateway is unreachable after multiple retries");
   }
 });
 
--- a/packages/grid_client/tests/modules/applications/casperlabs.test.ts
+++ b/packages/grid_client/tests/modules/applications/casperlabs.test.ts
@@ -1,0 +1,267 @@
+import axios from "axios";
+import { setTimeout } from "timers/promises";
+
+import { FilterOptions, GatewayNameModel, generateString, GridClient, MachinesModel, randomChoice } from "../../../src";
+import { config, getClient } from "../../client_loader";
+import { bytesToGB, generateInt, getOnlineNode, log, splitIP } from "../../utils";
+
+jest.setTimeout(300000);
+
+let gridClient: GridClient;
+let deploymentName: string;
+
+beforeAll(async () => {
+  gridClient = await getClient();
+  deploymentName = generateString(15);
+  gridClient.clientOptions.projectName = `casperlabs/${deploymentName}`;
+  gridClient._connect();
+  return gridClient;
+});
+
+//Private IP Regex
+const ipRegex = /(^127\.)|(^10\.)|(^172\.1[6-9]\.)|(^172\.2[0-9]\.)|(^172\.3[0-1]\.)|(^192\.168\.)/;
+
+test("TC2683 - Applications: Deploy Casperlabs", async () => {
+  /**********************************************
+     Test Suite: Grid3_Client_TS (Automated)
+     Test Cases: TC2683 - Applications: Deploy Casperlabs
+     Scenario:
+        - Generate Test Data/casperlabs Config/Gateway Config.
+        - Select a Node To Deploy the casperlabs on.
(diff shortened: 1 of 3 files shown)


packages/grid_client/tests/modules/applications/casperlabs.test.ts
```diff
@@ -0,0 +1,267 @@
+import axios from "axios";
+import { setTimeout } from "timers/promises";
+
+import { FilterOptions, GatewayNameModel, generateString, GridClient, MachinesModel, randomChoice } from "../../../src";
+import { config, getClient } from "../../client_loader";
+import { bytesToGB, generateInt, getOnlineNode, log, splitIP } from "../../utils";
+
+jest.setTimeout(1500000);
```
zaelgohary, line 8: Why the timeout?

packages/grid_client/tests/modules/applications/casperlabs.test.ts
```diff
@@ -0,0 +1,267 @@
...
+
+let gridClient: GridClient;
+let deploymentName: string;
+
+beforeAll(async () => {
+  gridClient = await getClient();
+  deploymentName = generateString(15);
```
zaelgohary, line 15: add cl prefix

packages/grid_client/tests/modules/applications/casperlabs.test.ts
```diff
@@ -0,0 +1,267 @@
...
+        - Assert that the returned domain is working
+          and returns correct data.
+    **********************************************/
+
+  //Test Data
+  const name = "gw" + generateString(10).toLowerCase();
+  const tlsPassthrough = false;
+  let cpu = generateInt(1, 4);
+  let memory = generateInt(1, 6);
+  let rootfsSize = generateInt(0, 2);
```
zaelgohary, line 51: Minimum resources should be as per the solution requirements. Please remove the generateInt and use the resources specified in the instance capacity.

packages/grid_client/tests/modules/applications/casperlabs.test.ts
```diff
@@ -0,0 +1,267 @@
...
+  let diskSize = generateInt(10, 50);
```
zaelgohary, line 52: This might fail due to invalid resources.

packages/grid_client/tests/modules/applications/casperlabs.test.ts
```diff
@@ -0,0 +1,267 @@
...
+    diskSize = generateInt(5, diskSize);
+
+    //Search for another node with lower resources.
+    nodes = await gridClient.capacity.filterNodes({
+      cru: cpu,
+      mru: memory,
+      sru: rootfsSize + diskSize,
+      farmId: 1,
+      availableFor: await gridClient.twins.get_my_twin_id(),
+    } as FilterOptions);
```
zaelgohary, line 102: I think we should stop the test if no node w valid resources is found.

packages/grid_client/tests/modules/applications/casperlabs.test.ts
```diff
@@ -0,0 +1,267 @@
...
+      })
+      .catch(() => {
+        log("gateway is not reachable");
+      });
+    if (reachable) {
+      break;
+    }
+    const wait = await setTimeout(5000, "Waiting for gateway to be ready");
+    log(wait);
+  }
```
zaelgohary, line 229: Maybe we can set the timeout first and reduce the number retries?

packages/grid_client/tests/modules/applications/casperlabs.test.ts
```diff
@@ -0,0 +1,267 @@
...
+  deploymentName = generateString(15);
+  gridClient.clientOptions.projectName = `casperlabs/${deploymentName}`;
+  gridClient._connect();
+  return gridClient;
+});
+
+//Private IP Regex
+const ipRegex = /(^127\.)|(^10\.)|(^172\.1[6-9]\.)|(^172\.2[0-9]\.)|(^172\.3[0-1]\.)|(^192\.168\.)/;
+
+test("TC2683 - Applications: Deploy Casperlabs", async () => {
```
zaelgohary, line 24: This test does way more than just deploying the instance. I suggest we remove all other code to seperate tests i.e checking for the contract, gateway. Maybe we can add another test for SSHing the deploying. All tests should be done only if the first has passed.

zaelgohary (changes requested):
The requested changes should be applied in Funkwhale & Peertube tests too.

packages/grid_client/tests/modules/applications/casperlabs.test.ts
A-Harby: The gateway takes time to load and get the site ready, more than 10 minutes in some cases.

packages/grid_client/tests/modules/applications/casperlabs.test.ts
A-Harby: Which resources could be invalid here?

packages/grid_client/tests/modules/applications/casperlabs.test.ts
zaelgohary: root fs should be dynamic depending on the CPU & memory.

packages/grid_client/tests/modules/applications/casperlabs.test.ts
zaelgohary: All. Here are the packages for Casperlabs as per the docs.

image

packages/grid_client/tests/modules/applications/casperlabs.test.ts
zaelgohary: But we did set a timeout before checking for the gateway, right?

packages/grid_client/tests/modules/applications/casperlabs.test.ts
A-Harby: Yes, I throw an error; it should stop the test. `throw new Error("no nodes available to complete this test");`.

packages/grid_client/tests/modules/applications/casperlabs.test.ts
A-Harby: image

packages/grid_client/tests/modules/applications/casperlabs.test.ts
A-Harby: The only test made here is to deploy and check the site availability; other assertions are necessary to make sure everything is done correctly and it doesn't take any time in any way.

I will check the possibility of making an SSH test and add in another PR if so.

packages/grid_client/tests/modules/applications/casperlabs.test.ts
```diff
@@ -46,10 +44,10 @@ test("TC2683 - Applications: Deploy Casperlabs", async () => {
   //Test Data
   const name = "gw" + generateString(10).toLowerCase();
   const tlsPassthrough = false;
-  let cpu = generateInt(1, 4);
-  let memory = generateInt(1, 6);
-  let rootfsSize = generateInt(0, 2);
-  let diskSize = generateInt(10, 50);
+  const cpu = 2;
+  const memory = 4;
+  const rootfsSize = 0;
```
zaelgohary, line 51: This shouldn't be 0. Can we set it to 2 instead?

packages/grid_client/tests/modules/applications/casperlabs.test.ts
```diff
@@ -0,0 +1,246 @@
...
+      log(res.status);
+      log(res.statusText);
+      log(res.data);
+      expect(res.status).toBe(200);
+      expect(res.statusText).toBe("OK");
+      expect(res.data).toContain("Your Casper node is now running succesfully on the ThreeFold Grid.");
+    });
+  } else {
+    throw new Error("Gateway is unreachable after multiple retries");
+  }
```
zaelgohary, line 221: Can we combine the 2 conditions in 1?

packages/grid_client/tests/modules/applications/casperlabs.test.ts
zaelgohary: This might not work all the time. Rootfs is computed on the CPU & memory size. So, if the CPU & memory > 0, rootfs will be more than 0. The VM might fail as we didn't provide the correct capacity it needs while filtering nodes. (The requested capacity is less than the needed capacity)

A-Harby: https://github.com/threefoldtech/tfgrid-sdk-ts/pull/2784/commits/d3bdda7299fbc574bd67833d5b24164aba18f714

amiraabouhadid (approved):

packages/grid_client/tests/modules/applications/casperlabs.test.ts
```diff
@@ -0,0 +1,242 @@
...
+    log(res);
+    expect(res.created).toHaveLength(0);
+    expect(res.updated).toHaveLength(0);
+    expect(res.deleted).toBeDefined();
+  }
+});
+
+afterAll(async () => {
+  return await gridClient.disconnect();
+}, 130000);
```
zaelgohary, line 242: No need to use afterAll as it's only one test. Can we combine both in afterEach?

A-Harby: Done in [d87b135](https://github.com/threefoldtech/tfgrid-sdk-ts/pull/2784/commits/d87b135083ec70715e0816363317dc54c7e9a16e), and a timeout is necessary because deleting and disconnecting take some time.

zaelgohary (approved):

Issue #2764, opened 2024-05-23
🐞 [Bug]: Error message Enhancement  in your farm page


khaledyoussef24:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### which package/s did you face the problem with?

Dashboard

### What happened?

while creating a farm if you use the same farm name (even if the name is used in another tfchain account) you get an error
![Screenshot from 2024-05-23 11-58-20](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/ab3cf317-2ce0-4131-a791-bed61733087d)
we can use a clear message "choose a different name" or "this name is already in use"

### which network/s did you face the problem on?

Main

### version

2.4.2

### Relevant screenshots/screen records

.

### Relevant log output

```console
.
```

khaledyoussef24:
verified on devnet 
version : 


![Image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/68cba285-72f0-4ba7-a4e1-d3e0b85722a3)

user now can not have 2 farms with the same name.
https://app.testlodge.com/a/26076/projects/40893/suites/234374?expand_section=426668#case_4289466

PR #2789, opened 2024-05-27
Update farm name validation

amiraabouhadid:
### Changes

Added validation for farm name to check if farm name exists in all farms on tfchain

### Related Issues

#2764

### Documentation PR

### DIFF
--- a/packages/playground/src/dashboard/components/create_farm.vue
+++ b/packages/playground/src/dashboard/components/create_farm.vue
@@ -70,6 +70,7 @@ import { ref } from "vue";
 import { getAllFarmsNames } from "@/utils/get_farms";
 import { notifyDelaying } from "@/utils/notifications";
 
+import { gridProxyClient } from "../../clients";
 import { useGrid } from "../../stores";
 import { createCustomToast, ToastType } from "../../utils/custom_toast";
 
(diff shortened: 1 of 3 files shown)


maayarosama (changes requested):
You shouldn't list all Farms, you should filter by farm name from the grid proxy and it the Farm exists then you should display it as a field validation error and block the creating of the farm

MohamedElmdary:
shouldn't we remove the unused code?
https://github.com/threefoldtech/tfgrid-sdk-ts/blob/development/packages/playground/src/dashboard/components/create_farm.vue#L82-L85
https://github.com/threefoldtech/tfgrid-sdk-ts/blob/development/packages/playground/src/dashboard/farms_view.vue#L8-L9
https://github.com/threefoldtech/tfgrid-sdk-ts/blob/development/packages/playground/src/dashboard/farms_view.vue#L28-L33
https://github.com/threefoldtech/tfgrid-sdk-ts/blob/development/packages/playground/src/dashboard/components/user_farms.vue#L332-L344

MohamedElmdary:
@amiraabouhadid if component no longer getting userFarms as ref so should we remove it's code?

https://github.com/threefoldtech/tfgrid-sdk-ts/blob/development/packages/playground/src/dashboard/farms_view.vue#L8-L9

https://github.com/threefoldtech/tfgrid-sdk-ts/blob/development/packages/playground/src/dashboard/farms_view.vue#L28-L33

maayarosama (approved):

Mahmoud-Emad (approved):
Good job ya Amira :)

Issue #2448, opened 2024-03-18
Add computecapacity module unit tests


zaelgohary:
### Which package/s are you suggesting this feature for?

grid_client

### Is your feature request related to a problem? Please describe

Add tests for [computecapacity.ts](https://github.com/threefoldtech/tfgrid-sdk-ts/blob/development/packages/grid_client/src/zos/computecapacity.ts)

zaelgohary:
Work completed:

Added some tests for the module and encountered some issues regarding its validation. Opened an issue [here](https://github.com/threefoldtech/tfgrid-sdk-ts/issues/2821).

A-Harby:
Verified test have been added here: https://github.com/threefoldtech/tfgrid-sdk-ts/blob/development/packages/grid_client/tests/modules/compute_capacity.test.ts and they are working and passing here: https://github.com/threefoldtech/tfgrid-sdk-ts/runs/27433452415#r0s3.

![image](https://github.com/user-attachments/assets/6b1f52d7-9db4-48d0-9915-b0fa37557b4e)

PR #2820, opened 2024-05-30
Add compute capacity unit tests

zaelgohary:
### Description

Add compute capacity unit tests

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/2448

### Checklist

- [x] Tests included
- [x] Build pass
- [ ] Documentation

### DIFF
--- a/packages/grid_client/tests/modules/compute_capacity.test.ts
+++ b/packages/grid_client/tests/modules/compute_capacity.test.ts
@@ -46,6 +46,12 @@ describe("Compute Capacity module", () => {
     expect(result).toThrow();
   });
 
+  test("cpu & memory empty values.", () => {
+    const result = () => computeCapacity.challenge();
+
+    expect(result).toThrow();
+  });
+
   test("An error should be thrown if cpu & memory negative values.", () => {
     const negative_cpu = -1;
     const negative_mem = -1;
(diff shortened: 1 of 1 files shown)


xmonader (commented):
Really good progress

I shouldn't be able to construct a class in invalid state, so what I suggest is we maybe introduce a constructor that enforces our invariants. 

That being said, I should be expecting the following
1- valid state scenarios to construct e.g a for loop to create computecapacity with numbers (or use a library to generate these)
2- checks for invariants e.g in negative, zeros, decimals, non-ints, boundary values, overflow values, and special numbers like min/max for ram and cpu and ensuring you can't initialize less or higher than them

Mahmoud-Emad (approved):
The code looks good to me. However, when I tried to run all of the tests, including yours, nothing happened, and my laptop crashed. Please have someone else run the tests instead of me.

zaelgohary:
> The code looks good to me. However, when I tried to run all of the tests, including yours, nothing happened, and my laptop crashed. Please have someone else run the tests instead of me.

Does this happen every time you run these tests?

amiraabouhadid (changes requested):
fails 
image
image

zaelgohary:
> fails image image

Because the validation doesn't work in all workloads. The tests are implemented correctly though.

amiraabouhadid (approved):

Issue #2726, opened 2024-05-19
Ensure IP is not already registred on other farms before adding it


sameh-farouk:
### Which package/s are you suggesting this feature for?

Dashboard

### Is your feature request related to a problem? Please describe

https://github.com/threefoldtech/tfchain_graphql/issues/157

### Describe the solution you'd like

Add a validation on add ip form to ensure IP is not already registred on other farms before adding it

atefnazmy:
- Verified on Dev
- Version: ac03039
- The validation that checks for an existing IP has been added successfully.

![IP](https://github.com/user-attachments/assets/07a7f239-f81e-458a-b6b3-d7587233042e)

khaledyoussef24:
verified on devnet
version : c112478


![Image](https://github.com/user-attachments/assets/fffe737f-6b71-4d82-ae4c-59a4b41744ba)
![Image](https://github.com/user-attachments/assets/7c09b605-8678-41c3-8eb3-cf0cf824b096)





[testcase](https://app.testlodge.com/a/26076/projects/40893/suites/234374?expand_section=426668#case_4329585)

PR #2948, opened 2024-06-10
Add ip validation against all IPs 

amiraabouhadid:
### Changes

Added a function to check if Ip exists in all farms and throws an error if it does in single Ip addition or range Ips addition

### Related Issues

[#2726](https://github.com/threefoldtech/tfgrid-sdk-ts/issues/2726)


amiraabouhadid:
incorrect history

Issue #2637, opened 2024-05-01
🐞 [Bug]: Node public config validation isn't consistent with the tfchain validation 


khaledyoussef24:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### which package/s did you face the problem with?

Dashboard

### What happened?

i tried some cases not covered leading to unclear error
<img width="1680" alt="Screenshot 2024-05-01 at 2 56 02 PM" src="https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/57c89ac9-8275-4e4c-bc3f-838593bf4e0a">
<img width="373" alt="Screenshot 2024-05-01 at 2 56 14 PM" src="https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/035fb1af-c22e-4110-a16f-54eb66ab43e4">

### which network/s did you face the problem on?

Dev, QA

### version

2.4.0

### Relevant screenshots/screen records

-

### Relevant log output

```console
NA
```

khaledyoussef24:
can not produce the same issue again shall we close it ?

A-Harby:
Devnet.
I found some cases, like:

- The IPv4 or IPv6 shouldn't be in a different subnet than their gateway. 
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/eb14f2be-341c-43ec-aa67-53ba2643bc61)
- The IPv4 or IPv6 don't have subnet mask validation.
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/d9fe8ee6-9414-4f86-b39b-ba9564b8d6b9)
- The IPv4 or IPv6 don't have private IP validation.
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/2cc0be08-8868-4c24-9d91-05627bf46792)

And there are more cases where the validation is different between the dashboard and the TFchain.

Mahmoud-Emad:
## Issue update

### Investigation and Solution:

The gateway must reside within the same IP range (subnet) as the node's IP address to facilitate proper routing. This ensures that the node can effectively communicate and route its traffic through the designated gateway.

### Work done
#### Changes

1. **Gateway IP Range Validation:**
   - A new validation rule has been added to ensure that the gateway IP address is within the same subnet as the node's IP address. This validation is applied based on the IP version (IPv4 or IPv6).

2. **IPv4 Address Type Validation:**
[...]

PR #2963, opened 2024-06-12
Ensure Gateway IP Resides in the Same Subnet as Node's IP

Mahmoud-Emad:
### Description

The gateway must reside within the same IP range (subnet) as the node's IP address to facilitate proper routing. This ensures that the node can effectively communicate and route its traffic through the designated gateway.

### Changes

- A new validation rule has been added to ensure that the gateway IP address is within the same subnet as the node's IP address. This validation is applied based on the IP version (IPv4 or IPv6).
- A new validation rule has been added to verify whether the provided IPv4 address is public or private.
- A new validation rule has been added to verify whether the provided IPv6 address is public or private.

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/2637

### Screenshots

- image
- image
- image
- image
- image

### Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation

### DIFF
--- a/packages/playground/src/dashboard/components/public_config.vue
+++ b/packages/playground/src/dashboard/components/public_config.vue
@@ -105,6 +105,9 @@
                 </input-tooltip>
               </input-validator>
             </form-validator>
+            <v-alert type="error" v-if="errorMessage">
+              {{ errorMessage }}
+            </v-alert>
           </v-card-text>
           <v-card-actions class="justify-end my-1 mr-2">
             <!-- Remove and Generate Config Buttons -->
@@ -195,6 +198,7 @@ export default {
     const isConfigChanged = ref(false);
     const formRef = useFormRef();
     const grid = useGrid();
+    const errorMessage = ref<string | undefined>(undefined);
 
     const defualtNodeConfig = ref<PublicConfig>(publicConfigInitializer());
     const config = ref<PublicConfig>(publicConfigInitializer());
@@ -238,6 +242,7 @@ export default {
 
     async function AddConfig() {
       try {
+        errorMessage.value = undefined;
         isSaving.value = true;
         await grid.client.nodes.addNodePublicConfig({
           farmId: props.farmId,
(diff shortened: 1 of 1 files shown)


Mahmoud-Emad:
> I tried adding invalid data and the validation didn't work.
> 
> image

There is a validation on the chain that may validate the prefix length. The value '0' you sent is not valid. This scenario is covered.

image

Also, the error message has been updated to be user-readable, and I have logged the actual error from the chain in the console/logger for developer investigations.

image

zaelgohary:
> Also, the error message has been updated to be user-readable, and I have logged the actual error from the chain in the console/logger for developer investigations.
> 
> image

Do we really need to show the error to the user twice? I think the toast is enough. If the user wants to copy it, he can get it from the logs.

Mahmoud-Emad:
> > Also, the error message has been updated to be user-readable, and I have logged the actual error from the chain in the console/logger for developer investigations.
> > image
> 
> Do we really need to show the error to the user twice? I think the toast is enough. If the user wants to copy it, he can get it from the logs.

The toast takes about 5 seconds to disappear. This is why I displayed the alert, to allow the user to read the error message without feeling rushed.

packages/playground/src/dashboard/components/public_config.vue
```diff
@@ -26,7 +26,11 @@
               <!-- IPv4 -->
               <input-validator
                 :value="config.ipv4"
-                :rules="[validators.required('IPv4 is required.'), validators.isIPRange('IP is not valid.', 4)]"
+                :rules="[
+                  () => isPrivateIP('ipv4'),
+                  validators.required('IPv4 is required.'),
+                  validators.isIPRange('IP is not valid.', 4),
```
zaelgohary, line 32:
```suggestion
                  validators.isIPRange('IPv4 is not valid.', 4),
```

packages/playground/src/dashboard/components/public_config.vue
```diff
@@ -293,6 +309,46 @@ export default {
...
+        } else {
+          isRange = contains(config.value.ipv6, config.value.gw6);
+        }
+      } catch {
+        isRange = false;
+      }
+
+      if (!isRange) {
+        return {
+          message: "Gateway IP not in the provided IP range.",
```
zaelgohary, line 347:
```suggestion
          message: "Gateway IP is not in the provided IP range.",
```

zaelgohary (approved):

amiraabouhadid (approved):

Issue #2947, opened 2024-06-10
🐞 [Bug]: The domain reload button is repeating the list when clicked


A-Harby:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### which package/s did you face the problem with?

Dashboard

### What happened?

The domain list is repeated every time the reload button is pressed.

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/b34e4129-e352-4b44-afd2-05ea44742f3a)

### which network/s did you face the problem on?

Dev

### version

1b19f4f

### Relevant screenshots/screen records

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/1613a145-14a1-4129-ab72-c0119b95b162)

### Relevant log output

```console
.
```

A-Harby:
Verified, 08a275d.

The reload click now doesn't repeat the domains.

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/b2f49f1e-7820-4138-9876-79106e8f1db3)

PR #2972, opened 2024-06-12
Fix: reload gateway nodes button 

0oM4R:
- add reload node logic, and reset the gateway field, separate from `loadDomains` logic that actually load more domains 

### Related Issues

 - #2947

### Documentation PR

### Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation

### DIFF
--- a/packages/playground/src/components/node_selector/TfDomainName.vue
+++ b/packages/playground/src/components/node_selector/TfDomainName.vue
@@ -52,7 +52,7 @@
               <template #append-item v-if="pagination.page !== -1">
                 <VContainer>
                   <VBtn
-                    @click="reloadDomains()"
+                    @click="loadDomains()"
                     block
                     color="secondary"
                     variant="tonal"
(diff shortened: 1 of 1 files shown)


packages/playground/src/components/node_selector/TfDomainName.vue
```diff
@@ -52,7 +52,7 @@
               <template #append-item v-if="pagination.page !== -1">
                 <VContainer>
                   <VBtn
-                    @click="reloadDomains()"
+                    @click="loadDomains()"
```
Mahmoud-Emad, line 55: The `()` is not needed when calling a function without parameters in the template; it will be called normally.

Mahmoud-Emad (commented):
Good job ya Omar

Mahmoud-Emad:
While testing, I noticed that 2 requests are invoked when the reload icon is pressed. Both of them have the same response/result. Are they both needed?

[Screencast from 06-13-2024 08:46:01 AM.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/57001890/fae17381-9c84-46f3-b185-87078710879a)

0oM4R:
> While testing, I noticed that 2 requests are invoked when the reload icon is pressed. Both of them have the same response/result. Are they both needed?
> 
>  Screencast.from.06-13-2024.08.46.01.AM.webm

@Mahmoud-Emad 
they are not the same; first is to get the page count, with return count flag, and the other is to get the gateway names. 
Anyway,  I think this is not related to the issue If you have any idea about enhancing the flow, please open a separate issue with it

Mahmoud-Emad (approved):

Mahmoud-Emad:
https://github.com/threefoldtech/tfgrid-sdk-ts/issues/2982

PR #2986, opened 2024-06-13
Add IP validation

amiraabouhadid:
### Changes

add validation that checks if ip exists in any farm, not just selected farm

### Related Issues

#2726


amiraabouhadid:
incorrect history

PR #2987, opened 2024-06-13
Add ip validation

amiraabouhadid:
### Changes
checks if IP exists in all farms 

### Related Issues

#2726

### DIFF
--- a/packages/playground/src/dashboard/components/add_ip.vue
+++ b/packages/playground/src/dashboard/components/add_ip.vue
@@ -156,7 +156,7 @@ export default {
         };
       }
 
-      if (await IpExistsCheck(toPublicIP.value)) {
+      if (await IpExistsCheck(publicIP.value)) {
         return {
           message: "IP exists.",
         };
@@ -214,7 +214,7 @@ export default {
           message: "IP is not public.",
         };
       }
-      if (await IpExistsCheck(publicIP.value)) {
+      if (await IpExistsCheck(toPublicIP.value)) {
         return {
           message: "IP exists.",
         };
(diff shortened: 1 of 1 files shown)


packages/playground/src/dashboard/components/add_ip.vue
```diff
@@ -201,6 +212,12 @@ export default {
           message: "IP is not public.",
         };
       }
+      const ips = await gqlClient.publicIps({ ip: true }, { where: { ip_eq: publicIP.value } });
+      if (ips.length > 0) {
+        return {
+          message: "IP exists",
+        };
+      }
```
zaelgohary, line 220: We use the same function implementation twice. Can we extract it outside the block and call it in both places?

packages/playground/src/dashboard/components/add_ip.vue
```diff
@@ -146,12 +149,19 @@ export default {
...
+    async function ipcheck() {
       if (PrivateIp(publicIP.value.split("/")[0])) {
         return {
           message: "IP is not public",
         };
       }
+
+      if (await IpExistsCheck()) {
+        return {
+          message: "IP exists",
```
zaelgohary, line 161:
```suggestion
          message: "IP exists.",
```

packages/playground/src/dashboard/components/add_ip.vue
```diff
@@ -201,6 +214,11 @@ export default {
           message: "IP is not public.",
         };
       }
+      if (await IpExistsCheck()) {
+        return {
+          message: "IP exists",
```
zaelgohary, line 219:
```suggestion
          message: "IP exists.",
```

zaelgohary (commented):
Good job, Amira! just small comment.

maayarosama (approved):

MohamedElmdary (approved):

Issue #2909, opened 2024-06-09
🐞 [Bug]: Node pilot deployments does not have Domain name


khaledyoussef24:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### which package/s did you face the problem with?

Dashboard

### What happened?

made a node pilot deployment but no domain name was provided 
i could only reach the deployment through public ip and not secured
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/b9af81f7-432f-4441-a8f0-491bc55ec135)
[Screencast from 06-09-2024 06:18:33 AM.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/748a94f9-3e0e-4f3e-82fe-c36b908215cc)

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/4dff2df9-d3dd-4c8f-8cc2-970d74d17f05)

there is no domain name in deployment details like other solutions also
![Screenshot from 2024-06-09 06-20-05](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/71678206-9619-4c5c-9392-beaa6d642414)
 

### which network/s did you face the problem on?

Dev

### version

2.5

### Relevant screenshots/screen records

.

### Relevant log output

```console
.
```

zaelgohary:
Work completed:

Added domain to node pilot.

khaledyoussef24:
verified on devnet 
version : db2daf4


![Image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/9b0f7740-5a1d-4798-ab10-a179948ac8d5)
![Image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/6e6c591e-161e-4fde-bf4b-a9700bae11e6)
![Image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/4a1f0884-5a18-4bb4-911a-ff6c7ede4887)
![Image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/8aa93216-5396-4340-aea2-ab55fe04bb87)
![Image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/a1527805-b847-4d70-bd51-4a933ec7299e)

https://app.testlodge.com/a/26076/projects/40893/suites/234374?expand_section=422332#case_4150351

PR #3010, opened 2024-06-23
Add Node Pilot domain

zaelgohary:
### Description

- Add Node Pilot domain

image

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/2909

### Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation


packages/playground/src/weblets/tf_node_pilot.vue
```diff
@@ -141,13 +166,34 @@ async function deploy() {
...
+
+  if (!selectionDetails.value?.domain?.enableSelectedDomain) {
+    vm[0].customDomain = selectionDetails.value?.domain?.customDomain;
+    finalize(vm);
+    return;
+  }
+
+  try {
+    layout.value.setStatus("deploy", "Preparing to deploy gateway...");
+    console.log("vm[0]", vm[0]);
```
Mahmoud-Emad, line 181: Remove this line if not needed

packages/playground/src/weblets/tf_node_pilot.vue
```diff
@@ -141,13 +166,34 @@ async function deploy() {
...
+      port: 34416,
+      network: vm[0].interfaces[0].network,
+    });
+
+    finalize(vm);
+  } catch (e) {
+    layout.value.setStatus("deploy", "Rollbacking back due to fail to deploy gateway...");
+
+    await rollbackDeployment(grid!, name.value);
+    layout.value.setStatus("failed", normalizeError(e, "Failed to deploy a Casperlabs instance."));
```
Mahmoud-Emad, line 195:
```suggestion
    layout.value.setStatus("failed", normalizeError(e, "Failed to deploy a Node Pilot instance."));
```

packages/playground/src/weblets/tf_node_pilot.vue
```diff
@@ -102,20 +103,41 @@ const selectionDetails = ref<SelectionDetails>();
...
+    deploymentName: name.value,
+    projectName,
+    twinId: profileManager.profile!.twinId,
+  });
+
+  const domain = selectionDetails.value?.domain?.enabledCustomDomain
+    ? selectionDetails.value.domain.customDomain
+    : subdomain + "." + selectionDetails.value?.domain?.selectedDomain?.publicConfig.domain;
+
+  let vm: any;
```
Mahmoud-Emad, line 128: Could we specify the type of the VM instead of using `any`?

Mahmoud-Emad (changes requested):
Good job ya Zainab!

packages/playground/src/utils/types.ts
```diff
@@ -14,3 +14,39 @@ export interface IPublicConfig {
...
+  entrypoint: string;
+  metadata?: any;
+  description?: any;
+  rootfs_size: number;
+  corex: boolean;
+  gpu: any[];
+  deploymentName: string;
+  projectName: string;
+  wireguard: string;
+}
```
Mahmoud-Emad, line 52: It's not good if the machine type is not defined in the grid client. If it's not defined, it should be created, and if it is defined, it needs to be exported. However, writing a `GridClient` type in the `Dashboard` is not a good approach.

Mahmoud-Emad (changes requested):

packages/playground/src/utils/types.ts
zaelgohary: Done.

packages/grid_client/src/helpers/types.ts
```diff
@@ -89,3 +89,41 @@ export interface ZmachineData {
   /** The list of the GPUs */
   gpu: string[] | undefined;
 }
+
+interface VM {
+  version: number;
+  contractId: number;
```
Mahmoud-Emad, line 95: Upon reviewing the newly created file `types.ts`, I noticed that there is an interface for the `ZOS Machine` called `ZMachineData` that supports most of the `VM` types. However, it requires some additional types.

We can add the extra types as optional fields, like this:

```ts
interface ZMachineData {
  ... // Old fields
  deploymentName: string;
  projectName: string;
  wireguard: string;
}
```

Mahmoud-Emad (changes requested):

packages/grid_client/src/helpers/types.ts
zaelgohary: I extended VM from ZmachineData.

Mahmoud-Emad (approved):

0oM4R (approved):
create a domain name works fine, some other flows have bugs will open issues for them

image

[Screencast from 27 يون, 2024 EEST 01:19:10 م.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/62248851/4e953242-b206-48f1-ad8e-d0e9ad66d624)

Issue #2805, opened 2024-05-29
New design for showing the node's price and discount


AhmedHanafy725:
### Which package/s are you suggesting this feature for?

Dashboard

### Is your feature request related to a problem? Please describe

the current view lacks some info like the price before and after the discount,...

### Describe the solution you'd like

it should be shown in a tool-tip only for the dedicated nodes when the user hovers over the price or the reserve button and should include
- the discount
- The price before the discount.
- The price after the discount.
- Price/month
- Price/hour
- How many TFTs are required to unlock more discounts (better to show them in packages(bronze, silver, and gold))

ehab-hassan:
WIP
Investigate the best way to show all details

ehab-hassan:
WIP
still investigating the best way

AhmedHanafy725:
let's go with option 1

amiraabouhadid:
Work done: added table of packages

amiraabouhadid:
Debugging vuetify hide sort in v-data-table

khaledyoussef24:
verified on devnet
version : cddbd85



![Image](https://github.com/user-attachments/assets/e035934b-2120-4c3a-8bee-c5cff24fd53a)

[testcase](https://app.testlodge.com/a/26076/projects/40893/suites/234374?expand_section=426661#case_4264048)

PR #3028, opened 2024-06-25
Development 2.6 update node price discount popup

amiraabouhadid:
### Changes
Updated node price and discount popup to include packages and price before/after discount

### Related Issues

#2805 

### Documentation PR

### Checklist

image
image


zaelgohary (commented):
Build fails.

Mahmoud-Emad (approved):
Amazing ya Amiraaa

packages/playground/src/components/node_selector/TfNodeDetailsCard.vue
```diff
@@ -466,7 +503,35 @@ export default {
       }
     }
 
+    const discountPackagesItems = [
+      { name: "Default", discount: "-20%", tfts: 200 },
+      { name: "Bronze", discount: "-30%", tfts: 400 },
+      { name: "Silver", discount: "-40%", tfts: 600 },
+      { name: "Gold", discount: "-60%", tfts: 1000 },
```
AhmedHanafy725, line 510: The TFTs required to unlock the discount should be calculated based on the node resources.

AhmedHanafy725 (changes requested):

yarn.lock
AhmedHanafy725, line 1: why do we have changes here? did we install a new package?

AhmedHanafy725 (changes requested):

yarn.lock
amiraabouhadid: I did but then I removed them

AhmedHanafy725:
- I think we should show the price after the discount on the node card
- the price approximation should be ceil not floor
- I think we need something to say what is that table on the top 
- As a user why can't I see the required TFTs if I am not logged in
image

amiraabouhadid:
> * I think we should show the price after the discount on the node card
> * the price approximation should be ceil not floor
> * I think we need something to say what is that table on the top
> * As a user why can't I see the required TFTs if I am not logged in
>   image

good catch, fixed, thanks

MohamedElmdary (approved):

yarn.lock
AhmedHanafy725: ok then the changes in yarn.lock should be reverted

AhmedHanafy725:
sometimes the TFTs required is 0 and sometimes it has a value
image

Screenshot from 2024-07-10 17-19-49

samaradel (approved):

AhmedHanafy725 (approved):

Issue #2821, opened 2024-05-30
🐞 [Bug]: Add validation to compute capacity module


zaelgohary:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### which package/s did you face the problem with?

grid_client

### What happened?

No errors were thrown when testing with invalid min, max, empty values, and decimal values. 

### Steps To Reproduce

Run the following test: https://github.com/threefoldtech/tfgrid-sdk-ts/pull/2820/files

### which network/s did you face the problem on?

Dev

### version

Locally

### Relevant screenshots/screen records

-

### Relevant log output

```console
-
```

MohamedElmdary:
We can add a `MethodDecorator` which should validate object before trigger the targeted function

```ts
function TriggerValidate(): MethodDecorator {
  return <T>(
    target: object,
    propertyKey: string | symbol,
    descriptor: TypedPropertyDescriptor<T>,
  ): void | TypedPropertyDescriptor<T> => {
    function value(...args: any[]): any {
      const errors = validateSync(target);

      if (errors.length > 0) {
        console.log({ errors });
        throw new Error("Something went wrong");
      }

      return target[propertyKey]?.(...args);
    }

    return {
      ...descriptor,
      value: value as T,
    };
  };
}
```

which can be used like
```ts
class Zmount extends WorkloadData {
  @TriggerValidate()
  challenge(): string {
    return ''
  }
}
```

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/31689104/03319d0a-3f3a-41a1-b787-2818a31c6a1e)

zaelgohary:
Work completed:

Implemented a way to validate class prop/methods when set/call.

- feat: Add a class decorator with options to validate any needed prop/method
- feat: Use Validate class decorator in ComputeCapacity class to throw errors when needed
- chore: Remove skip from tests and run them using SKIP_PROPS_VALIDATION=true env var
- feat: Included SKIP_PROPS_VALIDATION to grid client workflow so tests won't fail

khaledyoussef24:
verified compute capacity is now added to the gridclient tests module and ran the test and it is working fine


![Image](https://github.com/user-attachments/assets/97495c3d-d764-44d3-aefa-61b76812bf64)
![Image](https://github.com/user-attachments/assets/2f8e33a0-39f9-463c-9c72-71c8e32fe211)
[...]

PR #3029, opened 2024-06-25
implement a way to validate class prop/methods when set/call

MohamedElmdary:
### Description
implement a way to validate class prop/methods when set/call

### Changes
- feat: Add a class decrator with options to validate any needed prop/method
- feat: Use Validate class decrator in ComputeCapacity class to throw error when needed
- chore: Remove skip from tests and run them using SKIP_PROPS_VALIDATION=true env var

### Related Issues
- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/2821

### Documentation PR

### Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation


packages/grid_client/src/helpers/validator.ts
```diff
@@ -31,4 +31,106 @@ function validateHexSeed(seed: string, length: number): boolean {
   return true;
 }
 
-export { validateObject, validateInput, validateHexSeed };
+interface ValidationOptions {
+  props?: boolean | string | string[];
+  methods?: boolean | string | string[];
+}
+
+function Validate(options?: ValidationOptions): ClassDecorator {
```
Mahmoud-Emad, line 39: Can you please provide docstrings for all of the created functions and describe the purpose of each function? Also, we typically use the `validateInput` as a decorator in the grid client. I know it's different, but at the very least we need to change the name of the newly created decorator. what do you think?

Mahmoud-Emad (changes requested):
Good job ya Rabe3 :)

.github/workflows/grid_client_tests.yml
```diff
@@ -23,6 +23,7 @@ jobs:
       RMB_PROXY: true
       STORE_SECRET: secret
       MNEMONIC: ${{ secrets.MNEMONIC }}
+      SKIP_PROPS_VALIDATION: true
```
AhmedHanafy725, line 26: please remove this and update the test expectation to throw when setting the value on the object

AhmedHanafy725 (changes requested):

Mahmoud-Emad (approved):

.github/workflows/grid_client_tests.yml
MohamedElmdary: Fixed
image

AhmedHanafy725 (approved):

Issue #3016, opened 2024-06-24
🐞 [Bug]:  Show/Hide icon in password field shouldn't be selectable


MohamedElmdary:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### which package/s did you face the problem with?

Dashboard

### What happened?

Click `tab` while focusing on any password input will select the hide/show icon which badly impact UX because as a user I would expect to focus next input to continue entering data

Solution:
We need to remove icon from tab index by setting `tabindex="-1"`

### which network/s did you face the problem on?

Dev

### version

latest

### Relevant screenshots/screen records

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/31689104/3dbab8e9-9606-495b-a6ac-19e0e544c13c)

### Relevant log output

```console
None
```

samaradel:
### Work in progress:
investigating..

samaradel:
### Work Completed:
- Remove password input wrapper
- Create a prepend inner icon for inputs with type password
- Fix show/hide icon selection when pressing tab
- Fix the order of dialog actions tabindex

samaradel:
### Work in Progress (WIP):
trying another solution without deleting password-wrapper component

samaradel:
### Work Completed:
- Fix show/hide icon selection when pressing tab
- Fix the order of dialog actions ```tabindex```

khaledyoussef24:
https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/d7ee288c-1345-4470-86b4-497f2d1817d7

verified on devnet version : d863879

PR #3045, opened 2024-06-26
Fix inputs with show/hide password icon

samaradel:
### Description

- Fix show/hide icon selection when pressing tab
- Fix the order of dialog actions ```tabindex```

### Changes

[Screencast from 26-06-24 17:10:30.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/11662805/ef9070c2-baf1-41c6-bd1a-8f20c3fe0eaf)

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/3016

### Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation


MohamedElmdary:
@samaradel

`PasswordInputWrapper` component was created to avoid multiple define of `showX` variables. however this is not only the case but also to avoid defined `v-slot:append-inner` over and over.
Also it's not only used in profile manager but also in multiple solution/apps for example k8s and it's currently broken.
image

I would suggest extend `usePasswordInput` hook to get a ref from password input itself then use that ref in onMounted to append `tabindex="-1"` attribute in icon which should fix the issue in the entire app without any breaks.


Also Currently if we click on input then tab it will popup to `the manual > Login > input again > the manual > So on... ` won't focus any button which definitely not fixing this issue at all

MohamedElmdary:
image

amiraabouhadid (commented):
please merge development_2.6 to resolve this issue 
image

MohamedElmdary:
@samaradel This solution is not generic to all password inputs so the issue still presented in a lot of inputs in application

Mahmoud-Emad (approved):

amiraabouhadid (approved):

MohamedElmdary (approved):

Mahmoud-Emad (approved):
LGTM

Issue #3078, opened 2024-07-03
`sdk-ts` : Support multiple stacks per network 


xmonader:
https://github.com/threefoldtech/home/issues/1550

0oM4R:
Issue update :
started to implement the logic but can't use the monitor package 
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/62248851/11b58ac6-38ef-422f-a723-6e023478ab13)

0oM4R:
also current mentor services logs the aliveness of the services but do not return any values,i think it need some refactor adding new methods

0oM4R:
WIP: creating the logic to pick the first available stack per service add this init class to the `alivenessCheck` in monitoring 
```ts
export class serviceStackPicker {
  private result: ServicesUrls = {};
  constructor(public options: StackPickerOptions) {}

  private async pingService(service: ILivenessChecker) {
    const status = await service.isAlive();
    if ("disconnect" in service) {
      await (service as IDisconnectHandler).disconnect();
    }
    return status;
  }
  async GetAvailableServices(): Promise<ServicesUrls> {
    if ("tfChain" in this.options)
      this.result.tfChain = await this.getAvailableServiceStack(
        this.options.tfChain.slice(1),
        new TFChainMonitor(this.options.tfChain[0]),
      );
    if ("gridProxy" in this.options)
      this.result.GirdProxy = await this.getAvailableServiceStack(
        this.options.GirdProxy.slice(1),
        new GridProxyMonitor(this.options.GirdProxy[0]),
      );


    return this.result;
  }
  async getAvailableServiceStack(urls: string[], service: ILivenessChecker) {
    let index = 0;
    do {
      const status = await this.pingService(service);
      if (status.alive) return service.serviceUrl();
      console.log(`${service.serviceName()}: failed to ping ${service.serviceUrl()}, due to ${status.error}`);
      service.setServiceURl(urls[index++]);
    } while (index < urls.length);
  }
}
```
and should be called anywhere with
```ts 
 const test = new serviceStackPicker({
    tfChain: ["faf", "wss://tfchain.dev.grid.tf/ws", "wsss://tfchain.dev.grid.tf/ws"],
    GirdProxy: ["hahah", "https://gridproxy.dev.grid.tf"],
  });
  console.log(await test.GetAvailableServices());
```

this snips is the initial phase of coding, feel free to suggest any refactors

0oM4R:
[...]

PR #3134, opened 2024-07-18
`Monitoring`:  add service URL manager

0oM4R:
### Description
- Add `ServiceURLManager` : a class managing service URLs with liveness checking `pick the first reachable endpoint for each service`
- Enhancements and Refactoring in  service monitoring

### Changes
- Refactor:
  - replace ws connection on `RMB` and `TFChain` ping with http using health endpoints,
  - replace  axios with fetch.
  - Renamed `sendGetRequest` to `sendRequest` 
  - fix GraphQl `isAlive` logic
  - replace serviceUrl and serviceName with getters
  - add update method that uses URL setter
  - Make all `IServiceBase` classes take an optional options on their contractor, add error handling for side effect
  - Grouped all RMB properties to RMB to `RMBProps`
  - ping url without change the service url, by passing optional url to `isAlive`

- Feat:
   - add error event handler on RMB direct client
   - add stats monitor class that monitors stats service
   - add activation monitor class  that monitor activation service
   - Add `serviceURLManager`
      - Support silent mode settings
      - Implement constructor to handle `URLManagerOptions`
      - Add private `handleErrorsOnSilentMode` method for error handling based on silent mode
      - Implement `getAvailableStack` method to find reachable service URL
      - Implement `getAvailableServicesStack` method to fetch and store available service URLs for all services
      
- Examples: update service monitor example and add `serviceURlManager` Example, also update example command to take the example file path
- Fix: main and type paths in package.json
### Related Issues

-  #3078

### Documentation PR

### Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation


packages/monitoring/src/serviceMonitor/rmb.ts
```diff
@@ -1,31 +1,45 @@
...
-    return resolveServiceStatus(this.rmbClient.ping(2));
+  public update(param: { url: string }): void {
+    this._url = param.url;
   }
-  public async disconnect() {
-    await this.rmbClient.disconnect();
+  async isAlive(): Promise<ServiceStatus> {
+    if (!this.url) throw new Error("Can't access before initialization");
+    const splittedURL = this.url.split("relay");
+    const url = splittedURL[1] ? "https://gridproxy" + splittedURL[1] : this.url;
```
AhmedHanafy725: please use replace instead of split

packages/monitoring/src/serviceMonitor/serviceURLManager.ts
```diff
@@ -0,0 +1,127 @@
...
+ * @param {URLManagerOptions<N>} [options] - The options for configuring the manager.
+ *
+ * @property {Service[]} [services] - The list of services to be managed.
+ */
+export class ServiceUrlManager<N extends boolean = false> {
+  public results: { [key: string]: ServiceUrl<N> } = {};
+  private retries = 3;
+  private silent: N = false as N;
+  public services: Service[];
+  public timeout = 20;
```
AhmedHanafy725: timeout should be less than that, maybe 5 seconds

packages/monitoring/src/serviceMonitor/serviceURLManager.ts
```diff
@@ -0,0 +1,127 @@
...
+   * @throws {Error} - Throws an error if no reachable service URL is found after checking all URLs.
+   *
+   */
+  async getAvailableStack(urls: string[], service: ILivenessChecker): Promise<ServiceUrl<N>> {
+    let error: Error | string = "";
+    for (let i = 0; i < urls.length; i++) {
+      await service.update({ url: urls[i] });
+      monitorEvents.log(`${service.name}: pinging ${service.url}`, "gray");
+      for (let retry = 0; retry < this.retries; retry++) {
+        const status = await this.pingService(service);
```
AhmedHanafy725: let's not wait for each url to respond, we can spawn all the requests simultaneously. and the wait for the promises to resolve (resolve with the first success)

AhmedHanafy725 (changes requested):

packages/monitoring/src/types/index.ts
```diff
@@ -1,40 +1,36 @@
...
-export interface IDisconnectHandler {
   /**
-   * Performs the disconnection from the service.
-   * @returns {Promise<void>} A promise that resolves when the disconnection is successful.
+   * Updates the service with the provided parameters.
+   * @param {P} param - The parameter object with specific keys, should be specified on class.
    */
-  disconnect: () => Promise<void>;
+  update(param: P): void;
 }
```
AhmedHanafy725: do we still need this update method?

0oM4R: I guess I should replace it with normal setter, and use it [here](https://github.com/threefoldtech/tfgrid-sdk-ts/pull/3134/files#diff-0741e94a896c535b6c4a88b69466d0ab9bbd819c6367c57aab6ab85528051e32R127-R132), with the reachable url

packages/monitoring/example/serviceMonitor.ts
```diff
@@ -1,11 +1,11 @@
-import { GraphQLMonitor, GridProxyMonitor, RMBMonitor, ServiceMonitor, TFChainMonitor } from "../src/";
+import { GraphQLMonitor, GridProxyMonitor, RMBMonitor, ServiceMonitor, TFChainMonitor } from "../src";
 async function HealthCheck() {
   try {
     const services = [
       new GridProxyMonitor("<FakeURL>"),
       new GraphQLMonitor("https://graphql.dev.grid.tf/graphql"),
       new TFChainMonitor("wss://tfchain.dev.grid.tf/ws"),
-      new RMBMonitor("wss://relay.dev.grid.tf", "wss://tfchain.dev.grid.tf/ws", "mnemonic", "sr25519"),
+      new RMBMonitor("wss://relay.dev.grid.tf"),
```
Mahmoud-Emad, line 8: Are we going to run it only on Devnet? i think the network needs to be configurable

packages/monitoring/example/serviceURLManager.ts
```diff
@@ -0,0 +1,38 @@
...
+  RMBMonitor,
+  ServiceUrlManager,
+  TFChainMonitor,
+  URLManagerOptions,
+} from "../src";
+
+const gridproxy = ["https://gridproxy.dev.grid.tf/"];
+const rmb = ["https://graphql.dev.grid.tf", "wss://relay.dev.grid.tf", "wss://relay.02.dev.grid.tf"];
+const graphql = ["https://graphql.dev.grid.tf/graphql"];
+const tfChain = ["wss://tfchain.dev.grid.tf/ws"];
```
Mahmoud-Emad, line 13: Same here, These URLs need to be loaded from a config file, they should be configurable for sure

packages/monitoring/example/serviceURLManager.ts
```diff
@@ -0,0 +1,38 @@
...
+
+async function checkStacksAvailability<N extends boolean>(services: URLManagerOptions<N>) {
+  try {
+    const pickedUrls = await new ServiceUrlManager(services).getAvailableServicesStack();
+    console.log(pickedUrls);
+    process.exit(0);
+  } catch (err) {
+    console.log(err);
+  }
```
Mahmoud-Emad, line 22: Instead of just printing out the error, what do you think if we throw it?

packages/monitoring/example/serviceURLManager.ts
```diff
@@ -0,0 +1,38 @@
...
+  silent: true,
+  services: [
+    { service: new GridProxyMonitor(), URLs: gridproxy },
+    { service: new GraphQLMonitor(), URLs: graphql },
+    {
+      service: new RMBMonitor(),
+      URLs: rmb,
+    },
+    { service: new TFChainMonitor(), URLs: tfChain },
+  ],
```
Mahmoud-Emad, line 36: Suggestion: Instead of hardcoding the URLs directly in the code, it would be more maintainable to store them in a configuration file. This approach is especially beneficial if you expect to manage multiple URLs, as it centralizes the configuration and makes the code cleaner and easier to update.

For example, you could create a `config.json` file like this:

```json
{
  "gridproxy": {
    "urls": [
      "https://gridproxy.dev.grid.tf/",
      "https://gridproxy.qa.grid.tf/",
      "https://gridproxy.test.grid.tf/"
    ]
  },
  "graphql": {
// Also could be by the service network name, e.g. dev, qa, test etc, but I think to be an array will be better
    "urls": []
  }
}
```

Then, you'll define an interface to load this config and be accessible and have an autocomplete

```ts
export interface Config {
  gridproxy: {
    urls: string[];
  };
  graphql: {
    urls: string[];
  };
}
```

And yes, you have to initialize a new class to load and parse the config file

```ts
export class ConfigLoader {

  public static load(): Config {
    ....
  }
}
```

This is a proposal feature to load the service URLs dynamically instead of hard-coding them

packages/monitoring/src/serviceMonitor/gridproxy.ts
```diff
@@ -1,19 +1,23 @@
...
-  serviceName() {
-    return this.name;
+  public get name() {
+    return this._name;
   }
-  serviceUrl() {
-    return this.url;
+  public get url() {
+    return this._url ?? "";
   }
```
Mahmoud-Emad: I noticed that you have multiple classes that share a significant amount of similar functionality. To improve code maintainability and reduce duplication, I suggest creating a `Base` class that encapsulates the common logic. Then, each of the existing classes can inherit from this `Base` class, allowing them to share the common behavior while still implementing any specific functionality they need.


```ts
class Base {
  get url(): string{}
  set url(name: string){}
  async isAlive(url: string): Promise<ServiceStatus> {}
  ....
}

class X extends Base {
  ping(url: string){
    ...
  }
}

class Y extends Base {
  pong(){
    ...
  }
}
```

packages/monitoring/src/serviceMonitor/rmb.ts
```diff
@@ -1,31 +1,41 @@
...
   }
-  public async isAlive(): Promise<ServiceStatus> {
-    if (!this.rmbClient?.con?.OPEN) await this.setUp();
-    return resolveServiceStatus(this.rmbClient.ping(2));
-  }
-  public async disconnect() {
-    await this.rmbClient.disconnect();
+  async isAlive(url = this.url): Promise<ServiceStatus> {
+    if (!url) throw new Error("Can't access before initialization");
+    const proxyUrl = url.replace("wss://relay", "https://gridproxy");
```
Mahmoud-Emad, line 12: Define the proxy URL instead of doing this replace

packages/monitoring/src/serviceMonitor/serviceURLManager.ts
```diff
@@ -0,0 +1,162 @@
...
+      const timeoutPromise = new Promise((resolve, reject) => {
+        setTimeout(() => {
+          reject(new Error("Timeout"));
+        }, _timeout * 1000);
+      });
+
+      const result = await Promise.race([statusPromise, timeoutPromise]);
+      if (result instanceof Error && result.message === "Timeout") {
+        throw result;
+      }
```
Mahmoud-Emad:
```suggestion
    const TIME_OUT = "Timeout";
    try {
      const statusPromise = service.isAlive(url);

      const timeoutPromise = new Promise((resolve, reject) => {
        setTimeout(() => {
          reject(new Error(TIME_OUT));
        }, _timeout * 1000);
      });

      const result = await Promise.race([statusPromise, timeoutPromise]);
      if (result instanceof Error && result.message === TIME_OUT) {
        throw result;
      }
```

Mahmoud-Emad (changes requested):
Good job, Omar

packages/monitoring/src/serviceMonitor/rmb.ts
0oM4R: explain please

packages/monitoring/example/serviceMonitor.ts
0oM4R: This is an example script to show how to use this pckage

packages/monitoring/example/serviceURLManager.ts
0oM4R: I just put those urls as example

packages/monitoring/example/serviceURLManager.ts
0oM4R: as we discuss, no need as  it is an example script

packages/monitoring/example/serviceURLManager.ts
0oM4R: as we discuss, no need as it is an example script

packages/monitoring/src/serviceMonitor/rmb.ts
0oM4R: we shouldn't do this, we are using the gridproxy for getting the rmb status, We shouldn't ask for the gridproxy url directly as this is class related logic, We may change the way we get the rmb status later so for now, we handle this on the go.

Mahmoud-Emad (approved):

AhmedHanafy725 (approved):

Issue #2015, opened 2024-01-17
APP: Gitea updated & Mycelium inside


xmonader:

Mahmoud-Emad:
Blocked on https://github.com/threefoldtech/tf-images/issues/206

zaelgohary:
@maxux Can you promote this too? https://hub.grid.tf/petep.3bot/gitea_mycelium.flist.md

zaelgohary:
@ehab-hassan This too needs a logo, please.

zaelgohary:
WIP:

Added the solution but didn't test it. Currently working on adding the env vars needed and testing.

zaelgohary:
WIP:

Added Gitea to the deployment list. Also, removed the env tab and added mail server. However, when I tried deploying and log the services, one of them was in a state error. The issue is blocked till the flist is fixed.

maxux:
Promoted: https://hub.grid.tf/tf-official-apps/gitea_mycelium.flist.md

zaelgohary:
Work completed:

Edited flist.

ehab-hassan:
@zaelgohary  logo white
![gitea](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/10484295/a96c7e88-f7e9-487c-94f6-17acbaee501d)

khaledyoussef24:
tested on devnet
version : db2daf4
Users now can deploy a gittea solution from the dashboard.
but still need a domain to visit after creating the vm and needing also to add an icon to the deployment list to visit the domain 
and i think it's type of solutions which will have a control panel.
but the machine worked fine and the status of the tasks is running


![Image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/1d7d3d4a-9415-4ec0-a62c-d6c7bc99c9b6)
![Image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/b82c4f1c-85d7-499f-b217-bbc8c99a62a4)
![Image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/9c0d4f0d-4091-4d05-954a-987923b0e4ad)
![Image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/569559ef-8d42-4364-8c67-a5dc946b1392)

Mik-TF:
Testing for docs on info_grid. Looks great so far! Nice work.

Some suggestions while testing:

[...]

PR #3163, opened 2024-07-24
Add domain to Gitea

zaelgohary:
### Changes

Add domain, edit docs

image

image

### Related Issues

-https://github.com/threefoldtech/tfgrid-sdk-ts/issues/2015

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/packages/playground/src/weblets/tf_gitea.vue
+++ b/packages/playground/src/weblets/tf_gitea.vue
@@ -1,7 +1,6 @@
 <template>
   <weblet-layout
     ref="layout"
-    @mount="layoutMount"
     :cpu="solution?.cpu"
     :memory="solution?.memory"
     :disk="disks.reduce((total, disk) => total + disk.size, rootFilesystemSize)"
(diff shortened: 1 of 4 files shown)


packages/playground/src/weblets/tf_gitea.vue
```diff
@@ -171,29 +187,47 @@ async function deploy() {
...
+      finalize(vm);
+      return;
+    }
+
+    try {
+      layout.value.setStatus("deploy", "Preparing to deploy gateway...");
+      await deployGatewayName(grid, selectionDetails.value.domain, {
+        subdomain,
+        ip: vm[0].interfaces[0].ip,
+        port: 3000,
```
Mahmoud-Emad, line 201: I think the port here needs to be added as a constant at the top of the file. I think it can be changed later in the Dockerfile, so setting it at the top of the file will be clearer. What do you think?

Mahmoud-Emad (dismissed):
Good job ya Zainab, just small comment

packages/playground/src/weblets/tf_gitea.vue
zaelgohary: The thing is, this is how it's implemented in all solutions. If it will be changed, we can open an issue to edit it in all solutions at once.

maayarosama (approved):

Mahmoud-Emad (approved):

PR #3164, opened 2024-07-24
added dashboard suite app

mik-tf:
### Description

- Dashboard Suite app for the grid stack suite project/tfgrid stack deployment ([see here](https://github.com/threefoldtech/home/issues/1563))

### Changes

- Added dashboard suite app in dashboard to deploy

### Related Issues

- based on the tf-images: https://github.com/threefoldtech/tf-images/pull/277

### info_grid Manual

- https://github.com/threefoldtech/info_grid/pull/600

### Screenshot

image


xmonader:
Where is the original ticket that has the background of that app/feature?

mik-tf:
It is from the tfgrid stack deployment project [here](https://github.com/threefoldtech/home/issues/1563)

I added the link in the issue to be clear.

I think it would be beneficial to have devs reviewing the PR, so it can be as good as possible. We can put it in draft if you think it's better.

@xmonader thanks!

packages/playground/src/views/dashboardsuite_view.vue
```diff
@@ -0,0 +1,25 @@
...
+
+<script lang="ts">
+import { ProjectName } from "../types";
+import TfCasperlabs from "../weblets/tf_dashboardsuite.vue";
+import TfDeploymentList from "../weblets/tf_deployment_list.vue";
+export default {
+  name: "DashboardSuiteView",
+  components: {
+    TfDeploymentList,
+    TfCasperlabs,
```
xmonader, line 19: Why?

mik-tf: Why TfCasperlabs? Because it worked with this but didn't work with TfDashboardSuite. I couldn't find how to make it work. If you have some potential fixes I could work with, it would be appreciated. Thanks.

mik-tf:
We won't use dashboard apps for the grid stack.

Issue #3038, opened 2024-06-26
Use ValidateMembers decorator in all Grid Client workloads


zaelgohary:
### Which package/s are you suggesting this feature for?

grid_client

### Is your feature request related to a problem? Please describe

Workloads' validation doesn't work properly.

### Describe the solution you'd like

Use the decorator implemented in this [PR](https://github.com/threefoldtech/tfgrid-sdk-ts/pull/3029)

zaelgohary:
Work completed:

Validated all workloads using ValidateMembers.

PR #3187, opened 2024-07-31
Add ValidateMembers to all workloads

zaelgohary:
### Description

- Validate all workloads using ValidateMembers

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/3038

### Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation


AhmedHanafy725 (changes requested):
it should be added to all zos types classes that use the class validator

zaelgohary:
> it should be added to all zos types classes that use the class validator

All parent workloads have been validated. Validation of child workloads is unnecessary because the decorator handles the validation of nested workloads if ValidateNested is used. 


I removed the public IP from the network used in zmachine test to check if it would fail and it did.

image

AhmedHanafy725 (approved):

amiraabouhadid (approved):

Issue #3025, opened 2024-06-25
(Qanet verification) 🐞 [Bug]: prevent deploy contracts on Grace period rented node


A-Harby:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### which package/s did you face the problem with?

Dashboard

### What happened?

After my rented node went to grace period, I tried to deploy a VM on it, and I was able to choose node and have the deploy button active with no error, only an alert saying that the node is on grace period.

And the alert is only shown on manual selection; it doesn't show it on automated selection.

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/95569e77-7da5-439c-b6ad-4e58895f782e)

### which network/s did you face the problem on?

QA

### version

2.5.0-rc3

### Relevant screenshots/screen records

It should show an error on node selection with a clear error message.

[screen-capture (36).webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/9038da82-3ed7-4595-82a5-556c8162418b)

https://github.com/threefoldtech/tfgrid-sdk-ts/issues/2572

### Relevant log output

```console
.
```

samaradel:
### Work in progress:
Tried to get node on grace period on QA, I didn't get the error that mentioned yet.. still debugging

samaradel:
### Work in progress:
The card with a grace period turned to red, needs to be undetectable too

samaradel:
### Work Completed:
Prevent deployment on nodes with contracts in the grace period

Mahmoud-Emad:
I would recommend excluding these nodes from the filters
also, if this is important info that the user should know, we can display a dialog with the nodes that are excluded. 

But, I think it's not a good behavior to tell the user that this node isn't available after filtering, he user needs to deploy with only a single click.

A-Harby:
Qanet 2.6.0-rc3.

The deployment is not possible for a grace period rented node on the automated Node Selection option.
![image](https://github.com/user-attachments/assets/c1b2b778-ba26-4690-ac53-22eb6faf7fb7)

But it's still the same on manual Node Selection.
![image](https://github.com/user-attachments/assets/e37b9dca-d687-4d8b-b909-b8b133547ead)
The deploy button is active and enabled and gives me this error after clicking it.
![image](https://github.com/user-attachments/assets/d9499217-b37b-40a0-af7f-29244d1dd9d0)

samaradel:
[...]

PR #3191, opened 2024-07-31
Deployment contracts in grace period

samaradel:
### Description

Prevent deployment on nodes with contracts in the grace period.

### Changes
[Screencast from 31-07-24 16:38:10.webm](https://github.com/user-attachments/assets/983793f1-9267-4533-8759-d614029e23c6)

### Related Issues

### Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation

### DIFF
--- a/packages/playground/src/components/node_selector/TfAutoNodeSelector.vue
+++ b/packages/playground/src/components/node_selector/TfAutoNodeSelector.vue
@@ -286,6 +286,14 @@ export default {
 
     const nodeInputValidateTask = useAsync<true, string, [NodeInfo | undefined]>(
       async node => {
+        if (node && node?.rentContractId !== 0) {
+          const { state } = await gridStore.grid.contracts.get({
+            id: node?.rentContractId,
+          });
+          if (state.gracePeriod) {
+            return false as true;
+          }
+        }
         const nodeCapacityValid = await checkNodeCapacityPool(gridStore, node, props.filters);
         const rentContractValid = props.filters.dedicated ? await validateRentContract(gridStore, node) : true;
         return nodeCapacityValid && rentContractValid;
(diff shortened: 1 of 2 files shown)


packages/playground/src/components/node_selector/TfAutoNodeSelector.vue
```diff
@@ -286,6 +286,14 @@ export default {
 
     const nodeInputValidateTask = useAsync<true, string, [NodeInfo | undefined]>(
       async node => {
+        if (node && node?.rentContractId !== 0) {
+          const { state } = await gridStore.grid.contracts.get({
+            id: node?.rentContractId,
+          });
+          if (state.gracePeriod) {
+            return false as true;
```
0oM4R: why we should do that? should we just modify the return type from true to boolean?

samaradel: No, as useAsync should return true.. you can try it.

0oM4R: image
compiled successfully, unable to test the functionality now,, will try again whenever the chain transfer got fixed

Mahmoud-Emad:
I would recommend excluding these nodes from the filters
also, if this is important info that the user should know, we can display a dialog with the nodes that are excluded. 

But, I think it's not a good behavior to tell the user that this node isn't available after filtering, he user needs to deploy with only a single click.

samaradel:
@Mahmoud-Emad I think we had an issue for displaying them before, @AhmedHanafy725 what do you think?

AhmedHanafy725:
IMO, we should show the node to him and notify him that the contract is in a grace period instead of hiding it and he will keep looking for why it's not on the list

maayarosama (approved):

0oM4R (approved):

PR #3200, opened 2024-08-01
Support multiple stacks per network in gird client

0oM4R:
### Description

Using monitoring package, we are getting the available url for services if it is not provided in clien options

### Changes

add `setServiceURLs` that is updating client options with available services urls for only the services that do not associated with url in client options.
this method is used in connect method and before `getDefaultUrl` to insure that we have the available stack url for each service `if it is not provided` 

### Related Issues

- #3078
### Documentation PR

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation


packages/grid_client/src/client.ts
```diff
@@ -100,6 +101,14 @@ class GridClient {
...
     }
   }
+  /**
+   * Sets the missing service URLs by getting the available stacks using `ServiceURLManager` and updating them in the client options.
+   *
+   * @returns {Promise<void>} A promise that resolves when the client options got updated.
+   */
+  async setServiceURLs(): Promise<void> {
+    await getAvailableURLs(this.clientOptions);
+  }
```
AhmedHanafy725: what is the added value of this method?

packages/grid_client/src/manageURLs.ts
```diff
@@ -0,0 +1,123 @@
...
+  const services = missingServicesURLS.reduce((acc, servicesName) => {
+    acc.push(prepareServices(servicesName, URLS[servicesName]));
+    return acc;
+  }, [] as Service[]);
+  const result = await new ServiceUrlManager({
+    services: services,
+    silent: true,
+  }).getAvailableServicesStack();
+  Object.assign(clientOptions, mapResult(result, missingServicesURLS));
+}
```
MohamedElmdary: I think mutating object byRef might lead to unexpected behaviour specially in vue apps as it turns objects into proxy which rerender ui whenever it gets mutated

So I think we should avoid update objects byref

AhmedHanafy725 (approved):

MohamedElmdary (approved):

Issue #3072, opened 2024-07-03
Refactor tfchain `_applyExtrinsic` error handling to include other dispatch errors 


0oM4R:
### Which package/s are you suggesting this feature for?

tfchain_client

### Is your feature request related to a problem? Please describe

dispatch error have other types than module e.g token. this is not handled, and it got thrown as generic tfchain error
https://github.com/threefoldtech/tfgrid-sdk-ts/blob/db2daf4712060aa6b098fdb0cbcce07fa3512b63/packages/tfchain_client/src/client.ts#L378-L389

### Describe the solution you'd like

add new method to pars the error and throw the proper error type,
 #3071

Mahmoud-Emad:
## Issue update

### Work done

- Created a new Error wrapper to throw errors based on their type.
- Determined the type of error and threw it accordingly.
- Updated the error interface to differentiate between Module, Token, and other error types.
- Remove the TFChain error from the types package.

Issue #3071, opened 2024-07-03
Refactor error types or introduce new class 


0oM4R:
### Which package/s are you suggesting this feature for?

types

### Is your feature request related to a problem? Please describe

For TFChain not all errors are handled, we can't sync the changes of the chain errors.
many tfchain errors got thrown as generic tfchain error  

### Describe the solution you'd like

we should have an error class that have module name and error name props, and can be passed the contractor

Mahmoud-Emad:
## Issue update

### Work done

- Created a new Error wrapper to throw errors based on their type.
- Determined the type of error and threw it accordingly.
- Updated the error interface to differentiate between Module, Token, and other error types.
- Remove the TFChain error from the types package.

Issue #2896, opened 2024-06-06
🐞 [Bug]: TFChain errors is outdated in types package 


0oM4R:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### which package/s did you face the problem with?

types

### What happened?

the client mismatch the chain errors

this is because of the tfchain errors module in types package is outdated with tfchain errors 
and the client is getting the error by its index from types package https://github.com/threefoldtech/tfgrid-sdk-ts/blob/c642cb74c789859577dd4455e0cd99594943d261/packages/tfchain_client/src/client.ts#L379

### which network/s did you face the problem on?

Dev

### version

b72ee05

### Relevant screenshots/screen records

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/62248851/cf772611-c940-4575-8de0-ff7ddacf4830)

### Relevant log output

```console
.
```

0oM4R:
Issue update: 
Currently, we are going to enhance decoding chain errors, to get the write section of the errror
while implementing this logic as mentioned on [polkadot.js docs](https://polkadot.js.org/docs/api/cookbook/blocks#how-do-i-determine-if-an-extrinsic-succeededfailed)
i couldn't access registry `this.api.registry.findMetaError`
but i was able to do the same logic on another script using our client

````ts
import { Keyring } from "@polkadot/api";
import { DispatchError } from "@polkadot/types/interfaces";

import { Client } from "./client";

async function main() {
  const client = new Client({
    url: "wss://tfchain.dev.grid.tf",
    mnemonicOrSecret: "violin chair festival kiss double belt hen differ stamp boy else bag",
  });
  await client.connect();
  await client.loadKeyPairOrSigner();
  const nonce = await client.api.rpc.system.accountNextIndex(client.address);
  const x = await client.api.tx.smartContractModule
    .cancelContract(119696)
    .signAndSend(client.keypair, { nonce }, ({ status, events }) => {
      if (status.isInBlock || status.isFinalized) {
        events
          .filter(({ event }) => client.api.events.system.ExtrinsicFailed.is(event))
          .forEach(
            ({
              event: {
                data: [error, info],
              },
            }) => {
              if ((error as DispatchError).isModule) {
                const decoded = client.api.registry.findMetaError((error as DispatchError).asModule);
                const { docs, method, section } = decoded;

                console.log(`${section}.${method}: ${docs.join(" ")}`);
              } else {
                console.log(error.toString());
[...]

PR #3207, opened 2024-08-04
Implement TFChainError in chain client 

Mahmoud-Emad:
### Changes

- Created a new Error wrapper to throw errors based on their type.
- Determined the type of error and threw it accordingly.
- Updated the error interface to differentiate between Module, Token, and other error types.
- Remove the TFChain error from the types package.

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/3071
- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/3072

### Checklist

- [x] Tests included
- [x] Build pass
- [ ] Documentation


packages/tfchain_client/src/errors.ts
```diff
@@ -0,0 +1,140 @@
...
+interface ITFChainError {
+  message: string;
+  keyError?: string;
+  section?: string;
+  method?: string;
+  args?: AnyTuple;
+  docs?: string[];
+}
+
+class TFChainError extends Error implements ITFChainError {
```
0oM4R: i think this should be in types package.

0oM4R:
I see we are going to move the tfchain error from the types package to the tfchain package, is there a reason for that, as we discussed before we need to only refactor the class, and for the other issue we could add a method to that class to resolve the error.

Mahmoud-Emad:
> I see we are going to move the tfchain error from the types package to the tfchain package, is there a reason for that, as we discussed before we need to only refactor the class, and for the other issue we could add a method to that class to resolve the error.



As discussed with @AhmedHanafy725 , we'll create an interface with specific fields to capture error details. We'll populate these fields based on the error type. Placing the changes in the types package makes sense yeah. Please review the changes.

packages/types/src/errors/tfchain/index.ts
```diff
@@ -1,6 +1,155 @@
...
+   *   - If it is an unknown dispatch error, throws a TFChainError for unknown dispatch error.
+   * - If the error is an instance of Error:
+   *   - If the message includes "Inability to pay some fees", throws an InsufficientBalanceError.
+   *   - Otherwise, throws a TFChainError for a generic error.
+   * - If the error is not a dispatch error or an instance of Error, throws a TFChainError for an unknown error.
+   * @returns {TFChainError} The specific TFChainError thrown based on the encountered error type.
+   */
+  throw(): TFChainError {
+    const args = this.extrinsic.method.args;
+    const method = this.extrinsic.method.method;
```
0oM4R: Please don't use the extrinsic data, it may invalid error data https://github.com/threefoldtech/tfgrid-sdk-ts/issues/2896, it happened before and we decided to get the exact error method https://github.com/threefoldtech/tfgrid-sdk-ts/pull/2995

we can't rely on the extrinsic

0oM4R (changes requested):

packages/types/src/errors/readme.md
```diff
@@ -52,7 +52,6 @@ Also, in some cases, we have an error that comes from one of the clients, but we
 
 #### Generic Errors
 
-0. TFChainError
```
0oM4R, line 55: can we return the TFChain error docs please ?

packages/types/src/errors/tfchain/index.ts
```diff
@@ -1,6 +1,163 @@
...
+          message: `Module Error: ${errorMessage}'`,
+          keyError: name,
+          section,
+          method: extrinsicMethod,
+          args: extrinsicArgs,
+          docs,
+        });
+      } else if (this.error.isToken) {
+        const tokenError = this.error.asToken.toHuman();
+        console.log({ tokenError });
```
0oM4R: please remove the log line

packages/types/src/errors/tfchain/index.ts
```diff
@@ -1,6 +1,163 @@
...
+   *   - If it is a module error, throws a TFChainError with module details.
+   *   - If it is a token error, throws a TFChainError with token details.
+   *   - If it is an arithmetic error, throws a TFChainError with arithmetic details.
+   *   - If it is a bad origin error, throws a TFChainError for bad origin.
+   *   - If it is an unknown dispatch error, throws a TFChainError for unknown dispatch error.
+   * - If the error is an instance of Error:
+   *   - If the message includes "Inability to pay some fees", throws an InsufficientBalanceError.
+   *   - Otherwise, throws a TFChainError for a generic error.
+   * - If the error is not a dispatch error or an instance of Error, throws a TFChainError for an unknown error.
+   * @returns {TFChainError} The specific TFChainError thrown based on the encountered error type.
```
0oM4R: do we actually return an error, i think we are always throw, ?

packages/types/src/errors/tfchain/index.ts
0oM4R, line 1: #### suggestion 

i notice that when throwing generic error we have something like this 
Screenshot from 2024-08-06 12-39-25

what i suggest :
1- set keyError default value to `Generic Error`
2- hide the undefined props,

we can do that by remove the class properties and if it is passed use `Object.DefineProperty`

```ts
class TFChainError extends Error implements ITFChainError {
  constructor(options: ITFChainError) {
    super(options.message);
    this.name = "TFChainError";
    options.keyError && Object.defineProperty(this, "keyError", options.keyError);
    options.section && Object.defineProperty(this, "section", options.section);
    options.args && Object.defineProperty(this, "args", options.args);
    options.method && Object.defineProperty(this, "method", options.method);
    options.docs && Object.defineProperty(this, "docs", options.docs);
  }
}
```
this will hide the undefined props:
image

but this have <IMO> one consequenc: we won't have autcomplete and will be diffecult to access the props if it is defined, so the solution is to define an interface with the same name of the class with the optional props 
```ts
interface TFChainError {
  keyError?: string;
  section?: string;
  method?: string;
  args?: AnyTuple | string[];
  docs?: string[];
}

```
now we will be able to access them normally 
Screenshot from 2024-08-06 13-17-46

0oM4R (commented):
thank you for the great job ya Omda

packages/types/src/errors/tfchain/index.ts
Mahmoud-Emad: In my honest :), it's not good to follow, why? because we have a class that implements an interface, you have to follow the `interface`, there are no blockers with the `undefined` value, make it simple ya Omar xD :P

0oM4R: Do you mean the 'ITFChainError'? 
We can remove the optional values from the interface and follow it, anyways my suggestion is for better look only.

packages/types/src/errors/readme.md
```diff
@@ -95,11 +94,26 @@ For Errors that are related to workloads and deployments operations
...
+
+The `TFChainError` class is designed to capture and throw detailed error information.
+This class includes fields for the error `message`, `key`, `section`, `method`, `arguments`, and `documentation`.
+The TFChainError class extends the native JavaScript Error class, providing a structured way to handle errors in the TFChain context.
+
+#### TFChainErrorWrapper
+
+The `TFChainErrorWrapper` class processes `DispatchError` and other errors, throwing `TFChainError` with relevant details. This wrapper class ensures that all types of errors encountered during `TFChain` operations are appropriately captured and detailed error messages are thrown
+
+#### How to use it
```
0oM4R: please add the case of using TFChainError 
like i this case 
```ts 
throw new TFChainError({
        message: `Error getting current block number for contract ${options.id} deletion: ${err}`,
      });

```

packages/types/src/errors/tfchain/index.ts
AhmedHanafy725, line 1: this file should be in tfchain client as it uses some types from polkadot which is not installed here and shouldn't be installed to as minimum as possible

AhmedHanafy725 (changes requested):

packages/types/src/errors/tfchain/index.ts
AhmedHanafy725: please move the docs with it too.

AhmedHanafy725 (approved):

0oM4R (approved):

Issue #3110, opened 2024-07-10
🐞 [Bug]: Discourse Mail Server is enabled despite the solution having no public IP


atefnazmy:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### which package/s did you face the problem with?

Dashboard

### What happened?

- While deploying Discourse application without a public IP, I observed that the Mail Server option is enabled and I could configure it and deploy the instance.

### Steps To Reproduce

- Navigate to the Dashboard
- Navigate to Deploy from the sidebar then click on Applications.
- Navigate to the Discourse deployment page by clicking on the Discourse card option
- Expand the network option and turn off the public IPV4 toggle
- Switch to the SMTP tab and add your smtp configuration and click the Deploy button

### which network/s did you face the problem on?

Dev

### version

cd929e2

### Relevant screenshots/screen records

https://github.com/threefoldtech/tfgrid-sdk-ts/assets/32517137/9e500630-6077-4684-81d9-b9b8aab41fce

 ### Related Issue:
- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/2522

### Relevant log output

```console
n/a
```

samaradel:
### Work in Progress (WIP):
found an issue coming from ```zos``` working on it

samaradel:
Work in Progress (WIP):
The value of the switch is transformed somehow, debugging..

samaradel:
### Work Completed:
Fix discourse TLS switch on transform data

atefnazmy:
- The Issue is still producible as the SMTP option isn't disabled despite the solution having no public IP.

https://github.com/user-attachments/assets/d6deeca3-0da9-46a0-b22e-17be7663164f

khaledyoussef24:
the issue is still producable 
<img width="1680" alt="Screenshot 2024-07-30 at 9 06 12 PM" src="https://github.com/user-attachments/assets/66d2b989-8f01-4229-93c6-fd8503e22b88">
<img width="1680" alt="Screenshot 2024-07-30 at 9 07 39 PM" src="https://github.com/user-attachments/assets/f42362dd-b3da-4a41-9bbe-701b03691acc">

khaledyoussef24:
verified now public ip and mail server are related
version : 184fa12
in Discourse ipv4 is enabled and can not be disables so no problem with mail server




[...]

PR #3211, opened 2024-08-04
Fix discourse SMTP

samaradel:
### Description

- Enable ```IPV4``` by default as [discourse docs](https://www.manual.grid.tf/documentation/dashboard/solutions/discourse.html#deployment)

### Changes

[Screencast from 04-08-24 14:34:33.webm](https://github.com/user-attachments/assets/71e0834b-fba4-450f-9ca8-817fc7951f01)

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/3110

### Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation

### DIFF
--- a/packages/playground/src/components/networks.vue
+++ b/packages/playground/src/components/networks.vue
@@ -25,6 +25,7 @@
                 color="primary"
                 inset
                 label="Public IPv4"
+                :disabled="isDiscourse"
                 :model-value="$props.ipv4"
                 @update:model-value="$emit('update:ipv4', $event ?? undefined)"
               />
@@ -130,6 +131,10 @@ export default {
       default: () => null,
     },
     disabled: { type: Boolean },
+    isDiscourse: {
+      type: Boolean,
+      default: () => null,
+    },
   },
   emits: {
     "update:ipv4": (value?: boolean) => value,
--- a/packages/playground/src/components/networks.vue
+++ b/packages/playground/src/components/networks.vue
@@ -25,7 +25,7 @@
                 color="primary"
                 inset
                 label="Public IPv4"
-                :disabled="isDiscourse"
+                :disabled="enableIpv4"
                 :model-value="$props.ipv4"
                 @update:model-value="$emit('update:ipv4', $event ?? undefined)"
               />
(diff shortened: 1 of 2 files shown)


zaelgohary (changes requested):
The mail server is used in Gitea & Taiga too.

Mahmoud-Emad (approved):

samaradel:
> The mail server is used in Gitea & Taiga too.

yes, but IPV4 is optional in Taiga and Gitea contrariwise Discourse is not an option

zaelgohary:
> > The mail server is used in Gitea & Taiga too.
> 
> yes, but IPV4 is optional in Taiga and Gitea contrariwise Discourse is not an option

Mail server settings are the same in Gitea, Taiga & Discorse. Enabling the SMTP server should enable IPv4 in any solution, otherwise, it won't work.

samaradel:
@zaelgohary it's a different issue and my solution here is only for discourse.

packages/playground/src/components/networks.vue
```diff
@@ -130,6 +131,10 @@ export default {
       default: () => null,
     },
     disabled: { type: Boolean },
+    isDiscourse: {
+      type: Boolean,
+      default: () => null,
+    },
```
MohamedElmdary: networks component shouldn't know which solution using it so we should expose `isDiscourse` variable instead we can expose something like `disableIpv4`, `disableIpv6`, ...etc so on

packages/playground/src/weblets/tf_discourse.vue
```diff
@@ -121,7 +127,7 @@ const profileManager = useProfileManager();
 const name = ref(generateName({ prefix: "dc" }));
 const email = ref(profileManager.profile?.email || "");
 const solution = ref() as Ref<SolutionFlavor>;
-const ipv4 = ref(false);
+const ipv4 = ref(true);
```
MohamedElmdary, line 130: does updating ipv4 to true by default has no effect on other solutions?

samaradel: no, it's only for discourse

maayarosama (approved):

zaelgohary (approved):

PR #3213, opened 2024-08-04
Fix multiple_vms script

zaelgohary:
### Changes

- Filtering nodes outside machines to ensure nodes exist before setting the nodeId
- Add types 
- Edit yarn lock

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/3090

### Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation

### DIFF
--- a/packages/grid_client/scripts/multiple_vms.ts
+++ b/packages/grid_client/scripts/multiple_vms.ts
@@ -1,26 +1,32 @@
-import { FilterOptions, MachinesModel } from "../src";
+import { FilterOptions, GridClient, MachinesModel } from "../src";
 import { config, getClient } from "./client_loader";
 import { log } from "./utils";
 
-async function deploy(client, vms) {
+async function deploy(client: GridClient, vms: MachinesModel) {
   const res = await client.machines.deploy(vms);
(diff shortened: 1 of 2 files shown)


packages/grid_client/scripts/multiple_vms.ts
```diff
@@ -1,28 +1,34 @@
...
   log("================= Canceling the deployment =================");
   log(res);
   log("================= Canceling the deployment =================");
 }
 
+async function getNodeId(client: GridClient, options: FilterOptions) {
+  const nodes = await client.capacity.filterNodes(options);
+
+  return nodes[0]?.nodeId;
+}
```
Mahmoud-Emad: I think you have to ping the node first here before you return it's ID

Mahmoud-Emad (changes requested):

packages/grid_client/scripts/multiple_vms.ts
zaelgohary: Done.

Mahmoud-Emad (approved):

amiraabouhadid (approved):

Issue #3232, opened 2024-08-06
🐞 [Bug]: Manage domains dialog validation error 


Mahmoud-Emad:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### which package/s did you face the problem with?

Dashboard

### What happened?

When opened the Manage Domains dialog to attach a domain name to my machine I got `Subdomain should consist of lowercase letters only.` without touching anything.

### Steps To Reproduce

1. Go to the [Staging Dashboard](https://staging.dashboard.dev.grid.tf/)
2. Go to [Deploy Micro Virtual Machine](https://staging.dashboard.dev.grid.tf/deploy/virtual-machines/micro-virtual-machine/) page
3. Deploy a new machine
4. open the Manage Domains dialog

### which network/s did you face the problem on?

Dev

### version

967b554

### Twin ID/s

143

### Relevant screenshots/screen records

[Screencast from 08-06-2024 11:34:07 AM.webm](https://github.com/user-attachments/assets/56f5a74a-667a-47bd-b960-c6420d4f9818)

### Relevant log output

```console
N/A
```

A-Harby:
Verified, Devnet e00cb4c.

The validation are now corrected and working fine.
![image](https://github.com/user-attachments/assets/6db44133-667d-43e6-ba10-0238a65bf09e)
![image](https://github.com/user-attachments/assets/ac14bb62-4988-4fdd-9de9-e0d95dde6d27)
![image](https://github.com/user-attachments/assets/a2e62181-3f93-4e64-accd-c63c6738ef18)
![image](https://github.com/user-attachments/assets/f8a3dfb3-8e07-4a83-b913-5722bc962654)
![image](https://github.com/user-attachments/assets/ae2a9e1e-1fc1-4899-bbb3-49b3beefa22b)
![image](https://github.com/user-attachments/assets/1e0c6a12-d0d0-40a4-9ee4-c7f9f5d9768f)

[TC2124 - Add gateway to a vm](https://app.testlodge.com/a/26076/projects/40893/suites/234374?expand_section=422333#case_4150333)

Issue #3148, opened 2024-07-23
Support selecting the network while adding domain to vm


AhmedHanafy725:
### Which package/s are you suggesting this feature for?

Dashboard

### Is your feature request related to a problem? Please describe

currently, the user can't select the network to deploy the gateway workload.

### Describe the solution you'd like

allow the user to select the network (yggdrasil, mycelium, public ipv4, and public ipv6) based on the supported networks on the vm.

Mahmoud-Emad:
## Issue Update

### Work done

- Check if the IP is V6 and format it to match '[IP]:PORT' pattern
- Pass the IP if it's V4
- Created a new v-select that lists all supported VM networks
- Bind the selected network on mounting the page
- Refactor the vue component

Mahmoud-Emad:
As @AhmedHanafy725 reported, there was an issue when selecting another network other than the WireGuard IP

```log
WorkloadDeployError: Failed to deploy gateway-name-proxy with name vm26vmf1d6qf on node 11 due to: failed to setup name proxy: failed to get user network: couldn't load network with id (pL1JFSUwD5VFC): open /var/run/cache/networkd/networks/pL1JFSUwD5VFC: no such file or directory.
```
So, to avoid this error, we must send the network option field only if the WireGuard IP selected

A-Harby:
I deployed a FullVM with all the networks and git cloned a sample site, then hosted that site on port 8000.
And deployed a gateway using all the interfaces. 
![image](https://github.com/user-attachments/assets/0e422864-5965-4816-bd31-43fc1d3da760)
The server was shown on Wireguard and IPv4 only and failed on all of the other interfaces (Planetary, Mycelium, and IPv6).
![image](https://github.com/user-attachments/assets/7f739c63-fb65-42f1-a045-593d82ff94af)
![image](https://github.com/user-attachments/assets/77981234-650b-4f92-bb03-b45e0b440a23)

Mahmoud-Emad:
Are you sure your server bind IPv6?

A-Harby:
@Mahmoud-Emad mentioned to add IPv6 bind to server to accept IPv6, like `python3 -m http.server --bind ::` and it worked on all the interfaces.

Verified, Devnet 28e17de.
![image](https://github.com/user-attachments/assets/2604a977-b2eb-4a05-a819-00ebc4c709e7)

[...]

PR #3275, opened 2024-08-08
Fix adding network IP other than WireGuard IP

Mahmoud-Emad:
### Description

As @AhmedHanafy725  reported, there was an issue when selecting another network other than the WireGuard IP

```log
WorkloadDeployError: Failed to deploy gateway-name-proxy with name vm26vmf1d6qf on node 11 due to: failed to setup name proxy: failed to get user network: couldn't load network with id (pL1JFSUwD5VFC): open /var/run/cache/networkd/networks/pL1JFSUwD5VFC: no such file or directory.
```
So, to avoid this error, we must send the network option field only if the WireGuard IP selected

### Changes

- Set the network field as an optional field
- Re-order the list of the networks
- Set the 'WireGuard IP' as the default IP
- Remove the subnet from the IP on change
- Lowercase the subdomain to avoid the validation issues

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/3148
- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/3232

### Documentation PR

### Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation

### DIFF
--- a/packages/playground/src/components/manage_gateway_dialog.vue
+++ b/packages/playground/src/components/manage_gateway_dialog.vue
@@ -313,6 +313,7 @@ export default {
         }
 
         await deployGatewayName(grid, selectionDetails.value!.domain, gwConfig);
+        suggestName();
         layout.value.setStatus("success", "Successfully deployed gateway.");
       } catch (error) {
         layout.value.setStatus("failed", normalizeError(error, "Something went wrong."));
--- a/packages/playground/src/components/manage_gateway_dialog.vue
+++ b/packages/playground/src/components/manage_gateway_dialog.vue
@@ -186,7 +186,12 @@ import { onMounted, type PropType, ref, watch } from "vue";
 import { useGrid } from "../stores";
 import { ProjectName } from "../types";
 import type { SelectionDetails } from "../types/nodeSelector";
-import { DeployGatewayConfig, deployGatewayName, type GridGateway, loadDeploymentGateways } from "../utils/gateway";
+import {
+  type DeployGatewayConfig,
+  deployGatewayName,
+  type GridGateway,
+  loadDeploymentGateways,
+} from "../utils/gateway";
 import { updateGrid } from "../utils/grid";
 import { normalizeError } from "../utils/helpers";
 import { generateName } from "../utils/strings";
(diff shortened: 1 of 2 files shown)


packages/playground/src/components/manage_gateway_dialog.vue
```diff
@@ -240,7 +254,7 @@ export default {
         (props.vm.projectName.toLowerCase().includes(ProjectName.Fullvm.toLowerCase()) ? "fvm" : "vm") +
         grid.config.twinId;
       prefix.value = oldPrefix.value + props.vm.name;
-      subdomain.value = generateName({ prefix: prefix.value }, 4);
+      subdomain.value = generateName({ prefix: prefix.value }, 4).toLowerCase();
```
AhmedHanafy725: the vm's name might have `_`, so will it be valid after the lowercase?

Mahmoud-Emad: The issue arises because the VM name can contain underscores, but domain names cannot. Previously, we didn’t check for underscores when generating a domain name based on the VM name.

I added a function that suggests a subdomain name by checking for underscores in the VM name and replacing them with an empty string.

Good catch!

packages/playground/src/components/manage_gateway_dialog.vue
```diff
@@ -235,12 +249,7 @@ export default {
 
     onMounted(async () => {
       updateGrid(grid, { projectName: "" });
-
-      oldPrefix.value =
-        (props.vm.projectName.toLowerCase().includes(ProjectName.Fullvm.toLowerCase()) ? "fvm" : "vm") +
-        grid.config.twinId;
-      prefix.value = oldPrefix.value + props.vm.name;
-      subdomain.value = generateName({ prefix: prefix.value }, 4);
+      suggestName();
```
0oM4R, line 252: I'm wondering if we can avoid this warning 

[Screencast from 2024-08-12 18-33-45.webm](https://github.com/user-attachments/assets/71a75c00-8a7c-4102-95c4-2b54cb38ef57)


after deploying the domain, and i want to deploy a new domain, the suggested name is still as the previous one, maybe we can use computed ref to suggest new name on opening tab1

0oM4R (approved):
Screenshot from 2024-08-13 12-16-36
tested, 
also test each ip separately

AhmedHanafy725 (commented):
can we change the `supported ips` to `supported interfaces`? and it should be before the network name (we may hide the network name entirely)

AhmedHanafy725 (approved):

Issue #3283, opened 2024-08-09
All Apps on dashboard should have mycelium by default and present mycelium in deployment table


mik-tf:
### Which package/s are you suggesting this feature for?

Dashboard

### Is your feature request related to a problem? Please describe

- We want to put Mycelium at the forefront of the TF ecosystem
   - We are currently revising the Manual in this regard
- We should make sure Mycelium is enabled by default and shown to the users in the Dashboard
- Might be a duplicate or outdated. Writing the issue to be sure.
### Describe the solution you'd like

- Every dashboard apps should have mycelium enabled by default in the Network menu
- Mycelium address should always be shown in the deployment table
- Should we remove planetary address in the deployment table? IMO yes so the focus is on Mycelium.
- Here below we see mycelium but the planetary network is shown "first".
![image](https://github.com/user-attachments/assets/e110ec07-6799-40e8-b5ce-b7a8b3d5e93a)

mik-tf:
- data point
   - the mainnet dashboard node pilot provides only ipv4 and ipv6, not mycelium

![image](https://github.com/user-attachments/assets/d7a1648e-23b0-4280-aa20-8ee8c91e7dd3)

Mahmoud-Emad:
Should be implemented in https://github.com/threefoldtech/tfgrid-sdk-ts/pull/3287

A-Harby:
Verified, Devnet 99b5130.

The Mycelium option is added to all of the solutions: VMs, Orchestrators, and Applications; and is enabled by default in all of them.
 
![image](https://github.com/user-attachments/assets/dc72f5b0-609f-496a-ace0-2bf378d8bdbb)
![image](https://github.com/user-attachments/assets/94362395-fe1e-4002-9e4c-e4f0975c2816)
![image](https://github.com/user-attachments/assets/228fd978-fed4-4b9c-b2c7-8f38f29c25f2)
![image](https://github.com/user-attachments/assets/83019057-9943-482c-ac55-c6dc8e29279b)

Issue #3282, opened 2024-08-09
🐞 [Bug]: Small Mycelium network typo


mik-tf:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### which package/s did you face the problem with?

Dashboard

### What happened?

- Mycelium should be written with upper case: Mycelium, not mycelium

### Steps To Reproduce

1. Check any app on the dashboard
2. go to network and check mycelium

### which network/s did you face the problem on?

Main

### version

latest

### Relevant screenshots/screen records

![image](https://github.com/user-attachments/assets/05f8c66d-5caa-4587-b573-bc45fa36bf47)

### Relevant log output

```console
- No output
```

0oM4R:
should be fixed in #3287  
https://github.com/threefoldtech/tfgrid-sdk-ts/pull/3287/files#diff-be42196b7e6e7db3698453b1b1cf1e9d7a1cb3ef8881a8cbb8f7cb06add18ec9R71

A-Harby:
Verified, Devnet a66e759.

Mycelium is now capitalized.
![image](https://github.com/user-attachments/assets/e444e230-c2fa-4846-b718-985d2e6562b3)

Issue #3185, opened 2024-07-31
🐞 [Bug]: The Most Applications don't have a Wireguard Network option


atefnazmy:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### which package/s did you face the problem with?

Dashboard

### What happened?

-  The most applications don't have a Wireguard Network option.

### which network/s did you face the problem on?

Dev

### version

ac03039

### Relevant screenshots/screen records

![no_wireguard](https://github.com/user-attachments/assets/fd36e383-077a-4a1d-9603-1341066fec50)

![with_wireguard](https://github.com/user-attachments/assets/90b37974-d349-4b0d-b669-2aac43b6fb3b)

### Relevant log output

```console
n/a
```

Mahmoud-Emad:
## Issue update 

### Changes

- Enabled the Public IPv4 and displayed it as a 'read-only' field
- Send the value of the 'Wireguard' option to the deployment
- Updated the return of the 'createWorker' function to be ready to accept all networks
- Updated the worker interface to accept all networks

- Infrared the type of the name in the validator to avoid typescript errors
- Send the value of the 'Wireguard' to the deployment
- Enable Mycelium option by default
- Added the Wireguard option to the K8S deployment

- Added a watcher into all solutions that require a domain to be deployed to watch on the custom domain value then updated the Wireguard option
- Added a new attr 'read-only' in the 'Networks' component to set the 'Wireguard' option as 'read-only'
- Update the opacity of the 'Wireguard Switch' in the 'Networks' component
- Updated all applications with the new requirement and tested the full workflow

#### Proposal

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/3294

Mahmoud-Emad:
- Established a new constant by computing the options for IPv4 and custom domains to decide whether the Wireguard option should be read-only or not.
- Monitored the 'isWireguardReadonly' constant to determine the value of the Wireguard feature based on it.

A-Harby:
Verified, Devnet 99b5130.

The wiregurad option is added to all of the solutions: VMs, Orchestrators and Applcations.
 
![image](https://github.com/user-attachments/assets/dc72f5b0-609f-496a-ace0-2bf378d8bdbb)
![image](https://github.com/user-attachments/assets/94362395-fe1e-4002-9e4c-e4f0975c2816)
[...]

PR #3287, opened 2024-08-11
Support adding Wireguard option to all applications

Mahmoud-Emad:
### Description

WireGuard Option Across All Applications

- Added the ability to configure WireGuard access for machines.
- Integrated the WireGuard option into all applications within the Dashboard, enhancing security and connectivity options.

### Changes

- Enable Wireguard access to the machine.
- Include the option in all Dashboard applications.

- Enabled the Public IPv4 and displayed it as a 'read-only' field
- Send the value of the 'Wireguard' option to the deployment
- Updated the return of the 'createWorker' function to be ready to accept all networks
- Updated the worker interface to accept all networks

- Infrared the type of the name in the validator to avoid typescript errors
- Send the value of the 'Wireguard' to the deployment
- Enable Mycelium option by default
- Added the Wireguard option to the K8S deployment

- Added a watcher into all solutions that require a domain to be deployed to watch on the custom domain value then updated the Wireguard option
- Added a new attr 'read-only' in the 'Networks' component to set the 'Wireguard' option as 'read-only'
- Update the opacity of the 'Wireguard Switch' in the 'Networks' component
- Updated all applications with the new requirement and tested the full workflow

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/3185
- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/3283
- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/3282

### Screenshots

image

### Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation


zaelgohary (changes requested):
Still missing from the following apps:

- Freeflow
- Caprover
- K8S
- Owncloud

Mahmoud-Emad:
Let's use `Squash Merge` when merging this PR

packages/playground/src/weblets/tf_umbrel.vue
0oM4R, line 1: i think we don't need to disable the wg here we don't have domains

0oM4R:
I have a concern about switching off the wg access on enabling custom domain; I mean, I checked the network section and found that the wg is enabled, then scroll down and found gateways section and enabled the custom domain... then i will figure out later that my machine doesn't have wg access what do you think. 

Also, I will open an issue for adding tooltips for each disabled option to clarify why it is readonly

0oM4R:
> I have a concern about switching off the wg access on enabling custom domain; I mean, I checked the network section and found that the wg is enabled, then scroll down and found gateways section and enabled the custom domain... then i will figure out later that my machine doesn't have wg access what do you think.
> 
> Also, I will open an issue for adding tooltips for each disabled option to clarify why it is readonly

I understand that the gw deployment will fail if we disable the wireguard but my point is to notify the user somehow that we disabled the wireguard

0oM4R:
I deployed a wordpress inctance with custom domain and switched the wg option off, but i got wg configuration on my deployment!
I think this is because the wireguard is required as long as we have gw field,
the only case we should have wg is optional is when we have ipv4 and custom domain:
Screenshot from 2024-08-21 15-53-49

Mahmoud-Emad:
> I have a concern about switching off the wg access on enabling custom domain; I mean, I checked the network section and found that the wg is enabled, then scroll down and found gateways section and enabled the custom domain... then i will figure out later that my machine doesn't have wg access what do you think.
> 
> Also, I will open an issue for adding tooltips for each disabled option to clarify why it is readonly

please do

0oM4R:
> I have a concern about switching off the wg access on enabling custom domain; I mean, I checked the network section and found that the wg is enabled, then scroll down and found gateways section and enabled the custom domain... then i will figure out later that my machine doesn't have wg access what do you think.

both parts ?

0oM4R:
> > I have a concern about switching off the wg access on enabling custom domain; I mean, I checked the network section and found that the wg is enabled, then scroll down and found gateways section and enabled the custom domain... then i will figure out later that my machine doesn't have wg access what do you think.
> > Also, I will open an issue for adding tooltips for each disabled option to clarify why it is readonly
> 
> please do




 @Mahmoud-Emad  both parts ?

packages/playground/src/components/networks.vue
```diff
@@ -68,22 +68,24 @@
...
+                label="Mycelium"
                 :modelValue="$props.mycelium"
                 @update:modelValue="$emit('update:mycelium', $event ?? undefined)"
               />
             </input-tooltip>
             <input-tooltip
               v-if="wireguard !== null"
               inline
-              tooltip="Enabling WireGuard Access allows you to establish private, secure, and encrypted connections to your instance."
+              tooltip="Enabling WireGuard Access allows you to establish private, secure, and encrypted connections to your instance. Please note that Unless you use the custom domain with an IPV4, this field will be read-only."
```
amiraabouhadid: should be 
```suggestion
              tooltip="Enabling WireGuard Access allows you to establish private, secure, and encrypted instance connections. Please note that this field will be read-only unless you use a custom domain with IPV4."
```

packages/playground/src/weblets/full_vm.vue
```diff
@@ -25,7 +25,7 @@
           :rules="[
             validators.required('Name is required.'),
             validators.IsAlphanumericExpectUnderscore('Name should consist of letters ,numbers and underscores only.'),
-            name => validators.isAlpha('Name must start with alphabet char.')(name[0]),
+            (name: string) => validators.isAlpha('Name must start with alphabet char.')(name[0]),
```
amiraabouhadid:
```suggestion
            (name: string) => validators.isAlpha('Name must start with an alphabetical character.')(name[0]),
```

packages/playground/src/weblets/full_vm.vue
```diff
@@ -102,7 +102,7 @@
             :value="disks[index].name"
             :rules="[
               validators.required('Disk name is required.'),
-              name => validators.isAlpha('Name must start with alphabet char.')(name[0]),
+              (name: string) => validators.isAlpha('Name must start with alphabet char.')(name[0]),
```
amiraabouhadid:
```suggestion
            (name: string) => validators.isAlpha('Name must start with an alphabetical character.')(name[0]),
```

packages/playground/src/weblets/full_vm.vue
```diff
@@ -102,7 +102,7 @@
...
               validators.minLength('Disk minLength is 2 chars.', 2),
```
amiraabouhadid:
```suggestion
              validators.minLength('Disk name minimum length is 2 characters.', 2),
```

packages/playground/src/weblets/full_vm.vue
```diff
@@ -102,7 +102,7 @@
...
               validators.isAlphanumeric('Disk name only accepts alphanumeric chars.'),
```
amiraabouhadid:
```suggestion
              validators.isAlphanumeric('Disk name only accepts alphanumeric characters.'),
```

packages/playground/src/weblets/full_vm.vue
```diff
@@ -102,7 +102,7 @@
...
               validators.maxLength('Disk maxLength is 50 chars.', 50),
```
amiraabouhadid:
```suggestion
              validators.maxLength('Disk name maximum length is 50 characters.', 50),
```

packages/playground/src/weblets/micro_vm.vue
```diff
@@ -27,7 +27,7 @@
           :rules="[
             validators.required('Name is required.'),
             validators.IsAlphanumericExpectUnderscore('Name should consist of letters ,numbers and underscores only.'),
-            name => validators.isAlpha('Name must start with alphabet char.')(name[0]),
+            (name: string) => validators.isAlpha('Name must start with alphabet char.')(name[0]),
```
amiraabouhadid:
```suggestion
            (name: string) => validators.isAlpha('Name must start with alphabetical character.')(name[0]),
```

packages/playground/src/weblets/micro_vm.vue
```diff
@@ -91,7 +91,7 @@
             :value="envs[index].key"
             :rules="[
               validators.required('Key name is required.'),
-              key => validators.isAlpha('Key must start with alphabet char.')(key[0]),
+              (key: string) => validators.isAlpha('Key must start with alphabet char.')(key[0]),
```
amiraabouhadid:
```suggestion
              (key: string) => validators.isAlpha('Key must start with alphabetical character.')(key[0]),
```



please fix all other similar validation messages accordingly

packages/playground/src/weblets/micro_vm.vue
```diff
@@ -91,7 +91,7 @@
...
               validators.pattern('Invalid key format.', { pattern: /^[^0-9_\s][a-zA-Z0-9_]+$/ }),
               validators.maxLength('Key max length is 128 chars.', 128),
```
amiraabouhadid:
```suggestion
              validators.maxLength('Key maximum length is 128 characters.', 128),
```

amiraabouhadid (changes requested):

packages/playground/src/weblets/tf_casperlabs.vue
```diff
@@ -36,7 +36,14 @@
...
       />
 
-      <Networks v-model:ipv4="ipv4" v-model:planetary="planetary" v-model:mycelium="mycelium" v-model:ipv6="ipv6" />
+      <Networks
+        v-model:ipv4="ipv4"
+        v-model:planetary="planetary"
+        v-model:mycelium="mycelium"
+        v-model:ipv6="ipv6"
+        v-model:wireguard="wireguard"
+        :readOnlyWireGuard="!selectionDetails?.domain?.enabledCustomDomain"
```
0oM4R: we should add IPv4 to this condition

0oM4R (changes requested):

packages/playground/src/weblets/freeflow.vue
```diff
@@ -210,10 +223,22 @@ async function deploy() {
 function updateSSHkeyEnv(selectedKeys: string) {
   selectedSSHKeys.value = selectedKeys;
 }
+
+watch(
+  () => selectionDetails.value?.domain?.enabledCustomDomain,
```
0oM4R: can we move this logic to the network tab?
instead of passing readonly flag, i suggest to pass the customDomainEnabled and check for both of customdomain enabled and ipv4 values then handle the wireguard option

Mahmoud-Emad: I have an open issue for enhancing the entire component, please if you want to do this enhancement in this PR, check with @AhmedHanafy725

https://github.com/threefoldtech/tfgrid-sdk-ts/issues/3294

Mahmoud-Emad: It's also a valid comment tbh, we can do it here to avoid code redundancy

Mahmoud-Emad: - Established a new constant by computing the options for IPv4 and custom domains to decide whether the Wireguard option should be read-only or not.
- Monitored the 'isWireguardReadonly' constant to determine the value of the Wireguard feature based on it.

MohamedElmdary (approved):

packages/playground/src/weblets/tf_presearch.vue
0oM4R, line 1: i found the wg option is disabled while we don't have ipv4 here, but the idea is we are not using the gateway

amiraabouhadid (approved):

0oM4R (approved):

zaelgohary (approved):

Issue #3169, opened 2024-07-28
Configure the timeout in the dashboard


AhmedHanafy725:
### Which package/s are you suggesting this feature for?

Dashboard

### Is your feature request related to a problem? Please describe

In case the dashboard fails to load deployments as the network is slow, there is no way to change the timeout

### Describe the solution you'd like

The user should be able to configure the timeout he wants. This timeout should be stored in the user's metadata. and if it's not found fall back to the timeout configured in the config file.

0oM4R:
**Suggestion**: introduce a new `Setting` page that contains a timeout field, change password, dark/light mode.

amiraabouhadid:
@ehab-hassan can we please have a design for the settings page?

ehab-hassan:
@amiraabouhadid check this

https://www.figma.com/proto/7f7OANwfyTHQxtoP80D4rp/Untitled?node-id=1-16&t=eh8DWQe0dYbd8oMq-0&scaling=scale-down&content-scaling=fixed&page-id=0%3A1

ehab-hassan:
updated
![Frame 386](https://github.com/user-attachments/assets/1699dd79-73a1-45e6-859c-9f96f8af5d02)
https://www.figma.com/proto/7f7OANwfyTHQxtoP80D4rp/Untitled?node-id=9-228&t=eh8DWQe0dYbd8oMq-0&scaling=min-zoom&content-scaling=fixed&page-id=0%3A1

amiraabouhadid:
added update theme card and settings page

amiraabouhadid:
blocked on #3341

khaledyoussef24:
verfied on devnet
Version fccc6af

settings page is now available 
![Image](https://github.com/user-attachments/assets/3cc861c5-6b61-4324-9cf4-1d431b6b0fa4)
![Image](https://github.com/user-attachments/assets/9745d230-9b90-4572-9ea9-0c6fe3d1d9d7)


[testsuite](https://app.testlodge.com/a/26076/projects/40893/suites/234374?expand_section=454504#suite_section_454504)

xmonader:
The behaviour of the timeout should be verified

khaledyoussef24:
Tried with a throttled connection and increased time out and it works.

Do you recommend a specific test that could be more reliable?
@xmonader

xmonader:
> Tried with a throttled connection and increased time out and it works.
> 
[...]

PR #3297, opened 2024-08-12
Add settings page

amiraabouhadid:
### Changes
Added a settings page where user can update the following: 
- application theme
- user password
- session timeout 

### Related Issues

#3169 
#3341 

### Documentation PR

For UI changes, Please provide the documentation PR on [info_grid](https://github.com/threefoldtech/info_grid)

### Checklist
- [x] test cases:
  - [x] form validation is triggered only on change
   - Theme Selector  
     - [x] Theme is set to system mode on mount
     - [x] Selected theme is updated 
     - [x] Update button is disabled when the selected theme is the current theme
     - [x] Theme updates when there is a system theme update if the system option is selected
   - Password Updater
     - [x] The current password validation works by checking if the entered password is the same as the one entered in the session storage. It also checks for minimum length and that the password field is filled.
     - [x] new password validation works for a min length check, the field is filled check and that new password is not the same as the current password
     - [x] Confirm password validation works by watching the new password change and ensuring both passwords match. confirm password validation is only triggered when the input field is entered  
   - Timeout Updater
      - [x] query timeout is retrieved from window.env.TIMEOUT on mount and stored in local storage
      - [x] query timeout validation works by checking for min value of 3 seconds   
      - [x] query timeout update reflects in the entire application and in local storage even after refresh  
      - [x] deployment timeout is retrieved from grid client and stored in local storage on mount
      - [x] deployment timeout validation works by checking min value of 3 seconds
      - [x] deployment timeout reflects application wide even after refresh as updated value is stored in local storage
  
image


packages/playground/src/views/settings.vue
```diff
@@ -0,0 +1,245 @@
+<template>
+  <view-layout>
+    <v-card color="primary" class="d-flex justify-center items-center mt-3 pa-3 text-center">
+      <v-icon size="30" class="pr-3">mdi-cog</v-icon>
+      <v-card-title class="pa-0">Settings</v-card-title>
+    </v-card>
+    <v-card class="my-5"
+      ><v-card-title>Theme</v-card-title> <v-card-text>Pick an application theme!</v-card-text>
```
zaelgohary, line 8: Is this needed? Even if the user didn't choose a theme, the first value will always be chosen.

packages/playground/src/views/settings.vue
```diff
@@ -0,0 +1,245 @@
...
+      <form-validator v-model="isValidTheme" valid-on-init>
+        <input-validator :rules="[validateTheme]" #="{ props }" ref="themeInput" :value="selectedTheme">
+          <v-select class="pa-3" :items="themes" v-model="selectedTheme" v-bind="props" />
+        </input-validator>
+      </form-validator>
+      <v-card-actions class="justify-end" :disabled="!isValidTheme">
+        <v-btn :disabled="!isValidTheme" @click="UpdateTheme" class="justify-end ml-auto">Update</v-btn></v-card-actions
```
zaelgohary: The current flow has 3 different forms which might be a hassle for the user if he wants to make changes in all of them. What if we combine all forms in one?

packages/playground/src/views/settings.vue
```diff
@@ -0,0 +1,245 @@
...
+    </v-card>
+    <v-card class="my-5"
+      ><v-card-title>Timeout</v-card-title>
+
+      <v-tooltip text="Set desired session timeout in seconds">
+        <template v-slot:activator="{ props }">
+          <v-card-text v-bind="props">Adjust Timeout <v-icon icon="mdi-information-outline" /></v-card-text>
+        </template>
+      </v-tooltip>
+      <form-validator v-model="isValidTimeout" valid-on-init>
```
zaelgohary: Input shows an error although the form was not entered.

image

packages/playground/src/views/settings.vue
```diff
@@ -0,0 +1,245 @@
...
+      <v-icon size="30" class="pr-3">mdi-cog</v-icon>
+      <v-card-title class="pa-0">Settings</v-card-title>
+    </v-card>
+    <v-card class="my-5"
+      ><v-card-title>Theme</v-card-title> <v-card-text>Pick an application theme!</v-card-text>
+      <form-validator v-model="isValidTheme" valid-on-init>
+        <input-validator :rules="[validateTheme]" #="{ props }" ref="themeInput" :value="selectedTheme">
+          <v-select class="pa-3" :items="themes" v-model="selectedTheme" v-bind="props" />
+        </input-validator>
+      </form-validator>
```
zaelgohary: Same here.

image

zaelgohary (dismissed):

zaelgohary:
After changing the timeout and reloading the page, this happened.

image

zaelgohary:
Validation is being triggered once the form is submitted.

[change-pass.webm](https://github.com/user-attachments/assets/ec94e8c7-96ec-4e26-8c42-89746f1272ef)

zaelgohary:
The position of the timeout tooltip is far away from the icon. Can we change its location to be beside it?

image

packages/playground/src/views/settings.vue
```diff
@@ -0,0 +1,245 @@
...
+      createCustomToast("Password Updated!", ToastType.success);
+      // reset input fields
+      currentPassword.value = "";
+      newPassword.value = "";
+      confirmPassword.value = "";
+      isValidPassword.value = false;
+    }
+    function UpdateTimeout() {
+      try {
+        window.env.TIMEOUT = selectedTimeout.value * 1000;
```
Mahmoud-Emad: What is this Timeout used for? request? should we update all requests with this timeout?

packages/playground/src/views/settings.vue
```diff
@@ -0,0 +1,245 @@
...
+import { createCustomToast, ToastType } from "@/utils/custom_toast";
+
+import { getCredentials, updateCredentials } from "../utils/credentials";
+export default {
+  name: "Settings",
+  setup() {
+    const theme = useTheme();
+    const KEY = "APP_CURRENT_THEME";
+
+    const themes = ["Dark Mode", "Light Mode"];
```
Mahmoud-Emad: As discussed in the group, we have three options for setting the display mode:

1. **System Mode**: Follows the system's default settings.
2. **Dark Mode**: Enables a dark-themed interface.
3. **Light Mode**: Enables a light-themed interface.

Display Mode Options

Mahmoud-Emad (dismissed):

Mahmoud-Emad:
Please check [CodeQL Suggestion](https://github.com/threefoldtech/tfgrid-sdk-ts/security/code-scanning/8)

packages/playground/src/views/settings.vue
amiraabouhadid: I implemented the design proposed by ehab in the attached issue, what do you think @AhmedHanafy725 @ehab-hassan ?

packages/playground/src/views/settings.vue
amiraabouhadid: yes, because the user should not be able to update timeout unless the timeout entered is different from the current one. And the form is mounted with the current timeout so that user knows what the current value is. what do you think @AhmedHanafy725 ?

packages/playground/src/views/settings.vue
amiraabouhadid: please refer to explanation above https://github.com/threefoldtech/tfgrid-sdk-ts/pull/3297#discussion_r1715225192

amiraabouhadid:
> After changing the timeout and reloading the page, this happened.
> 
> image

makes sense because that is now the current timeout so user cannot update a value with the same value

packages/playground/src/views/settings.vue
AhmedHanafy725: on mount, the user shouldn't see the errors. we can start validating on change.

packages/playground/src/views/settings.vue
amiraabouhadid: no, the current theme is chosen.

packages/playground/src/views/settings.vue
ehab-hassan: > The current flow has 3 different forms which might be a hassle for the user if he wants to make changes in all of them. What if we combine all forms in one?

All three sections are separated I don't think the user ll change all altogether also in the future maybe we will have more items  and ll divide them into sections like GitHub or Gitea let's leave it as is

ehab-hassan:
@amiraabouhadid can u add some space below the button and border of the section it seems stuck to it
small

packages/playground/src/views/settings.vue
amiraabouhadid: made timeout persistent in local storage to be applicable session wide

amiraabouhadid:
> Please check [CodeQL Suggestion](https://github.com/threefoldtech/tfgrid-sdk-ts/security/code-scanning/8)

was ignored and reason is mentioned in comment

ehab-hassan (approved):
Approved as design-wise

Mahmoud-Emad:
IMO: The theme should change select, a watcher should be there, and no need for the save button

image

see on LinkedIn

image

see on X

image

Mahmoud-Emad:
I selected the `System` option and it detracted the new value after refreshing the page, also when I changed my system value to `Light` and refreshed the `Dashboard` it selected `Light` instead of `System`

Mahmoud-Emad:
I think you need to watch on the system value and change the theme based on it

[Screencast from 08-18-2024 10:59:59 AM.webm](https://github.com/user-attachments/assets/9b59a0e9-ca0c-41a9-930e-0bdd2156c6d7)

Mahmoud-Emad:
Please set the location of the tooltip beside the icon 

image


no need, the input label is enough 

image

Mahmoud-Emad:
Does the value of the timeout input `1 second = 1000 milliseconds`? 

if yes: 
- Does the user know this?
- I think we need to define a validation rule to allow timeout values starting from 3000. 
- 1 is not valid at all
- image

If not, I think you need to clarify this; it isn't clear to me.

amiraabouhadid:
> Does the value of the timeout input `1 second = 1000 milliseconds`?
> 
> if yes:
> 
> * Does the user know this?
> * I think we need to define a validation rule to allow timeout values starting from 3000.
> * 1 is not valid at all
> * image
> 
> If not, I think you need to clarify this; it isn't clear to me.

I'm assuming user knows basic time concepts such as 1 second = 1000 milliseconds

amiraabouhadid:
> IMO: The theme should change select, a watcher should be there, and no need for the save button
> 
> image
> 
> see on LinkedIn
> 
> image
> 
> see on X
> 
> image

@ehab-hassan  what do you think?

ehab-hassan:
> > IMO: The theme should change select, a watcher should be there, and no need for the save button
> > image
> > see on LinkedIn
> > image
> > see on X
> > image
> 
> @ehab-hassan what do you think?

Both Ok for me @Mahmoud-Emad if you check Gitea 
image

also, Github needs a save button to update there is no specific rule here so let us use what we did for now,
Thanks

Mahmoud-Emad:
Great job, It would also be helpful if you could update the value of the selector. I also noticed that when the page reloads, the default value is 'Light mode' instead of matching the system.

[Screencast from 08-20-2024 11:25:28 AM.webm](https://github.com/user-attachments/assets/1ee6c723-4343-4014-b17a-dea0ae5522ed)

Mahmoud-Emad (dismissed):
My last comment isn't required, we can go with the current implementation since it won't break anything and it's only a styling issue <if you want>

zaelgohary (commented):
Validation gets triggered on submit here.

image

zaelgohary:
If I choose system mode, the dropdown will show dark/light mode depending on which value my system mode is. I suggest removing the system mode option as the dark & light modes are enough.

[screen-capture.webm](https://github.com/user-attachments/assets/8c3aa7d8-9987-478d-8a44-2006714390a7)

amiraabouhadid:
> If I choose system mode, the dropdown will show dark/light mode depending on which value my system mode is. I suggest removing the system mode option as the dark & light modes are enough.
> 
> [screen-capture.webm](https://github.com/user-attachments/assets/8c3aa7d8-9987-478d-8a44-2006714390a7)
> 

The system mode was requested by @Mahmoud-Emad above, what do you think @ehab-hassan @AhmedHanafy725 ?

ehab-hassan:
> > If I choose system mode, the dropdown will show dark/light mode depending on which value my system mode is. I suggest removing the system mode option as the dark & light modes are enough.
> > 
> >   
> >     
> >     
> > 
> >     screen-capture.webm
> 
> The system mode was requested by @Mahmoud-Emad above, what do you think @ehab-hassan @AhmedHanafy725 ?

When I tested it on the Windows system it automatically changed without the need for the user to go to the dashboard to choose light or dark mode so I think its good feature we can leave it @AhmedHanafy725 what you think?

AhmedHanafy725:
the default values need to be adjusted 10000 sec is too much for a query timeout and 0 in the deployment timeout looks wrong.
Screenshot from 2024-08-27 11-21-21

AhmedHanafy725:
changing the confirmation password first makes the matching validation fails

image

packages/playground/src/views/settings.vue
```diff
@@ -0,0 +1,337 @@
...
+              validateCurrentPassword,
+            ]"
+            #="{ props: validationProps }"
+          >
+            <VTextField
+              label="Current Password"
+              v-model="currentPassword"
+              v-bind="{ ...passwordInputProps, ...validationProps }"
+              autocomplete="off"
+              class="pa-3"
```
Mahmoud-Emad, line 37: Let's add the `hide-details="auto"` attr to all text fields to hide the space and only display it in case of errors/hints only

image

I also got this error when opening the console

image

packages/playground/src/App.vue
```diff
@@ -248,11 +262,31 @@ function navigateToHome() {
...
+    window.env.TIMEOUT = +localStorage.getItem(TIMEOUT_QUERY_KEY)! * 1000;
+  }
+
+  if (!localStorage.getItem(TIMEOUT_DEPLOYMENT_KEY)) {
+    localStorage.setItem(TIMEOUT_DEPLOYMENT_KEY, `${gridStore.client.clientOptions.deploymentTimeoutMinutes}`);
+  } else {
+    const client = gridStore.client as GridClient;
+    if (client) {
+      client.clientOptions.deploymentTimeoutMinutes = +localStorage.getItem(TIMEOUT_DEPLOYMENT_KEY)! / 60;
+      await client._connect;
```
Mahmoud-Emad: You shouldn't do this, you need to call the connect method

```suggestion
      await client.connect();
```

packages/playground/src/views/settings.vue
```diff
@@ -0,0 +1,337 @@
...
+    <v-card class="my-5"
+      ><v-card-title>Theme</v-card-title> <v-card-text>Pick an application theme!</v-card-text>
+
+      <v-select class="pa-3" :items="themes" v-model="selectedTheme" />
+
+      <v-card-actions class="justify-end mb-3 mx-3">
+        <v-btn :disabled="isCurrentTheme()" @click="UpdateTheme" class="justify-end ml-auto"
+          >Update</v-btn
+        ></v-card-actions
+      >
```
Mahmoud-Emad, line 16: [Screencast from 09-02-2024 11:30:44 AM.webm](https://github.com/user-attachments/assets/c5842dde-39d2-457e-b92a-b673de2074ba)

packages/playground/src/views/settings.vue
```diff
@@ -0,0 +1,337 @@
...
+    </v-card>
+    <v-card class="my-5"
+      ><v-card-title>Timeout</v-card-title>
+
+      <form-validator v-model="isValidTimeout">
+        <v-card-text>
+          Adjust Query Timeout
+          <v-tooltip location="end">
+            <template #activator="{ props: tooltipProps }">
+              <v-icon v-bind="tooltipProps"> mdi-information-outline </v-icon>
```
Mahmoud-Emad, line 106: The spaces are not the same 

image

packages/playground/src/views/settings.vue
```diff
@@ -0,0 +1,337 @@
...
+          <v-text-field
+            label="Enter Deployment timeout (sec)"
+            class="pa-5"
+            v-bind="props"
+            type="number"
+            v-model="selectedDeploymentTimeout"
+          ></v-text-field>
+        </input-validator>
+        <v-card-actions class="justify-end mb-3 mx-3">
+          <v-btn :disabled="!isValidTimeout || isCurrentTimeout()" @click="UpdateTimeout" class="justify-end ml-auto"
```
Mahmoud-Emad, line 163: Why i have to change the `Deployment Timeout` field to update my settings?

image

packages/playground/src/views/settings.vue
```diff
@@ -0,0 +1,337 @@
...
+import { useGrid } from "../stores";
+import { updateCredentials } from "../utils/credentials";
+
+export default {
+  name: "Settings",
+  setup() {
+    const theme = useTheme();
+    const THEME_KEY = "APP_CURRENT_THEME";
+    const TIMEOUT_QUERY_KEY = "APP_QUERY_TIMEOUT";
+    const TIMEOUT_DEPLOYMENT_KEY = "APP_DEPLOYMENT_TIMEOUT";
```
Mahmoud-Emad: I noticed that you have created the same constants with the same values in multiple files. What do you think about creating an enum to hold all keys and values, and then importing and using it?

packages/playground/src/views/settings.vue
```diff
@@ -0,0 +1,337 @@
...
+
+    const themes = ["System Mode", "Dark Mode", "Light Mode"];
+
+    const currentTheme = ref(localStorage.getItem(THEME_KEY));
```
Mahmoud-Emad: The `getItem` method will return null if no matched keys are found.
Please set the default theme to `System` if the `currentTheme` is `null`

packages/playground/src/views/settings.vue
```diff
@@ -0,0 +1,337 @@
...
+    const selectedTheme = ref(currentTheme.value == "light" ? "Light Mode" : "Dark Mode");
```
Mahmoud-Emad: Exactly, you are loading the current theme as one of just two options, while you support three options.

packages/playground/src/views/settings.vue
```diff
@@ -0,0 +1,337 @@
...
+    function validateNewPassword() {
+      if (newPassword.value === currentPassword.value) {
+        return { message: "New password cannot be the same as current password. Please enter a different password." };
+      }
+    }
+    function validateConfirmPassword(value: string) {
+      if (value !== newPassword.value) {
+        return { message: "Passwords should match." };
+      }
+    }
```
Mahmoud-Emad: We have already file for the validators

packages/playground/src/views/settings.vue
```diff
@@ -0,0 +1,337 @@
...
+          selectedQueryTimeout.value = currentQueryTimeout.value;
+        }
+
+        if (selectedDeploymentTimeout.value != currentDeploymentTimeout.value) {
+          localStorage.setItem(TIMEOUT_DEPLOYMENT_KEY, `${selectedDeploymentTimeout.value}`);
+          currentDeploymentTimeout.value = +localStorage.getItem(TIMEOUT_DEPLOYMENT_KEY)!;
+          selectedDeploymentTimeout.value = currentDeploymentTimeout.value;
+          if (client) {
+            client.clientOptions.deploymentTimeoutMinutes = +localStorage.getItem(TIMEOUT_DEPLOYMENT_KEY)! / 60;
+            await client._connect;
```
Mahmoud-Emad: Also here, we shouldn't call the `_connect` method, also, idk why you need to connect the client in the end of the block

Mahmoud-Emad (dismissed):
Good job ya Amira

Mahmoud-Emad:
Also, I noticed that the timeout only reflects when mounting the page, If I reloaded the window it doesn't.

Watch the deployment successfully deployed over the 3G network 

image

and the deployment timeout is only 3 sec

image

packages/playground/src/views/settings.vue
amiraabouhadid: default deployment timeout fixed

packages/playground/src/views/settings.vue
amiraabouhadid: it is set to System in App.vue

packages/playground/src/views/settings.vue
amiraabouhadid: to save new deployment timeout

packages/playground/src/views/settings.vue
amiraabouhadid: moved to validators.ts

amiraabouhadid:
> Also, I noticed that the timeout only reflects when mounting the page, If I reloaded the window it doesn't.
> 
> Watch the deployment successfully deployed over the 3G network
> 
> image
> 
> and the deployment timeout is only 3 sec
> 
> image

image
please check again

maayarosama (changes requested):
Confirm password validation is triggered after updating password 
[Screencast from 2024-09-03 14-31-16.webm](https://github.com/user-attachments/assets/073ed041-e5f4-48e3-abd1-8fee20dc4e3c)

maayarosama (dismissed):
I get incorrect password validation error onmount 
[Screencast from 2024-09-03 16-28-23.webm](https://github.com/user-attachments/assets/6c20fe0e-8672-40ec-b9d8-c9a329e7bb97)

Mahmoud-Emad:
I wonder if we should disable the `Confirm password` field in case the `New password` field has errors

image

PS: Let's skip it for now, reported in https://github.com/threefoldtech/tfgrid-sdk-ts/issues/3366

Mahmoud-Emad:
The confirm password field was invoked without touching it.

[Screencast from 09-04-2024 11:10:33 AM.webm](https://github.com/user-attachments/assets/b1fce8c6-86c7-42fe-8a14-61ecca8a0ce4)

Mahmoud-Emad:
The timeout is still not reflecting when reloading the page, I set the deployment time out to `3sec` and changed my network to 3G after refreshing the page I deployed a micro VM, and the deployment was successfully deployed. should see an error instead

image

maayarosama (approved):

packages/playground/src/App.vue
```diff
@@ -253,6 +271,42 @@ onMounted(async () => {
   openProfile.value = true;
 });
 
+watch(hasClient, () => setTimeouts());
```
zaelgohary, line 329: Is there a reason why we used hasClient in stead of hasGrid?

zaelgohary:
Please adjust max validation for the timeout.

image


I noticed that validation is triggered only after submit.

image

packages/playground/src/App.vue
```diff
@@ -253,6 +271,42 @@ onMounted(async () => {
...
+      if (clientOptions) {
+        const deploymentTimeoutMinutes = clientOptions.deploymentTimeoutMinutes;
+        if (deploymentTimeoutMinutes) {
+          client.clientOptions.deploymentTimeoutMinutes = +localStorageDeploymentTimeout / 60;
+          await client.connect();
+        }
+      }
+    }
+  }
+}
```
zaelgohary:
```suggestion
async function setTimeouts() {
  if (!localStorage.getItem(LocalStorageSettingsKey.TIMEOUT_QUERY_KEY)) {
    localStorage.setItem(LocalStorageSettingsKey.TIMEOUT_QUERY_KEY, `${window.env.TIMEOUT / 1000}`);
  } else {
    window.env.TIMEOUT = +localStorage.getItem(LocalStorageSettingsKey.TIMEOUT_QUERY_KEY)! * 1000;
  }

  const client = gridStore.client as GridClient;

  const localStorageDeploymentTimeout = localStorage.getItem(LocalStorageSettingsKey.TIMEOUT_DEPLOYMENT_KEY);
  if (client && !localStorageDeploymentTimeout) {
    const deploymentTimeoutMinutes = client.clientOptions.deploymentTimeoutMinutes;
    if (deploymentTimeoutMinutes) {
      localStorage.setItem(LocalStorageSettingsKey.TIMEOUT_DEPLOYMENT_KEY, `${+deploymentTimeoutMinutes * 60}`);
    } else {
      client.clientOptions.deploymentTimeoutMinutes = +localStorageDeploymentTimeout! / 60;
      await client.connect();
    }
  }
}
```

packages/playground/src/App.vue
amiraabouhadid: deploymentTimeout minutes exists under gridstore.client, not gridstore.grid. Please refer to IGrid interface.

amiraabouhadid:
@zaelgohary validation is triggered before submit. Please check 
https://www.loom.com/share/cae84ce444884c3988924054777fd0cb?sid=94948f36-440e-4eb4-9381-aff5b3f621b5

packages/playground/src/App.vue
amiraabouhadid: this doesn't work as clientoptions is not seen if not checked for before setting the timeout in the client. To get what I mean: try setting the deployment timeout to 3, deploy then refresh then deploy again, you won't get a timeout error after the refresh. Thanks though.

zaelgohary: This should do.
```
async function setTimeouts() {
  if (!localStorage.getItem(LocalStorageSettingsKey.TIMEOUT_QUERY_KEY)) {
    localStorage.setItem(LocalStorageSettingsKey.TIMEOUT_QUERY_KEY, `${window.env.TIMEOUT / 1000}`);
  } else {
    window.env.TIMEOUT = +localStorage.getItem(LocalStorageSettingsKey.TIMEOUT_QUERY_KEY)! * 1000;
  }

  const client = gridStore.client as GridClient;
  const clientOptions = client.clientOptions;
  const localStorageDeploymentTimeout = localStorage.getItem(LocalStorageSettingsKey.TIMEOUT_DEPLOYMENT_KEY);
  const deploymentTimeoutMinutes = clientOptions.deploymentTimeoutMinutes;

  if (client && clientOptions && deploymentTimeoutMinutes) {
    if (!localStorageDeploymentTimeout) {
      localStorage.setItem(LocalStorageSettingsKey.TIMEOUT_DEPLOYMENT_KEY, `${+deploymentTimeoutMinutes * 60}`);
    } else {
      client.clientOptions.deploymentTimeoutMinutes = +localStorageDeploymentTimeout! / 60;
      await client.connect();
    }
  }
}
```

packages/playground/src/App.vue
```diff
@@ -308,6 +326,42 @@ onMounted(async () => {
...
```
zaelgohary:
```suggestion
async function setTimeouts() {
  if (!localStorage.getItem(LocalStorageSettingsKey.TIMEOUT_QUERY_KEY)) {
    localStorage.setItem(LocalStorageSettingsKey.TIMEOUT_QUERY_KEY, `${window.env.TIMEOUT / 1000}`);
  } else {
    window.env.TIMEOUT = +localStorage.getItem(LocalStorageSettingsKey.TIMEOUT_QUERY_KEY)! * 1000;
  }

  const client = gridStore.client as GridClient;
  const clientOptions = client.clientOptions;
  const localStorageDeploymentTimeout = localStorage.getItem(LocalStorageSettingsKey.TIMEOUT_DEPLOYMENT_KEY);
  const deploymentTimeoutMinutes = clientOptions.deploymentTimeoutMinutes;

  if (client && clientOptions && deploymentTimeoutMinutes) {
    if (!localStorageDeploymentTimeout) {
      localStorage.setItem(LocalStorageSettingsKey.TIMEOUT_DEPLOYMENT_KEY, `${+deploymentTimeoutMinutes * 60}`);
    } else {
      client.clientOptions.deploymentTimeoutMinutes = +localStorageDeploymentTimeout! / 60;
      await client.connect();
    }
  }
}
```

zaelgohary (approved):
Good job ya Amira!

Issue #3179, opened 2024-07-30
Add animation for what's new in version


AhmedHanafy725:
### Which package/s are you suggesting this feature for?

Dashboard

### Is your feature request related to a problem? Please describe

For example, we have a lot of solutions but nothing to say that there is a new solution or any other features added we would like to show it to the user.

### Describe the solution you'd like

Adding animation to guide the user to the newly added main features on the release would be nice

0oM4R:
**Suggestion**: hide after click on the new solution 
hit for search : CSS : sudo class dirty

samaradel:
### Work in Progress (WIP):
chips are added and working on the logic of displaying it

samaradel:
### Work in Progress (WIP):
investigating how to save solutions appearance in kvstore

samaradel:
### Work Completed:
Add chip with New for new applications and will expire after 1 month.

P.S. PR is drafted till we get the animated image from @ehab-hassan

ehab-hassan:
I added like this @samaradel I need ur help to apply at just the new App not all

![new_APP](https://github.com/user-attachments/assets/472edf6a-c6c9-42f9-a8eb-02db04bed23b)

samaradel:
@ehab-hassan updated only for the new app/version, thanks

![Image](https://github.com/user-attachments/assets/fac91cb0-195a-4812-bf72-23a41b9e99bc)

A-Harby:
Waiting for https://github.com/threefoldtech/tfgrid-sdk-ts/issues/3448 to verify.

0oM4R:
https://github.com/threefoldtech/tfgrid-sdk-ts/pull/3493#issuecomment-2388500470
looks like there is an issue with responsive will move it to accepted

samaradel:
> [#3493 (comment)](https://github.com/threefoldtech/tfgrid-sdk-ts/pull/3493#issuecomment-2388500470) looks like there is an issue with responsive will move it to accepted

@ehab-hassan please check

ehab-hassan:
fixed @ #3511

A-Harby:
Verified, https://github.com/threefoldtech/tfgrid-sdk-ts/issues/3448#issuecomment-2413166331.

PR #3299, opened 2024-08-12
Add new key word for new apps

samaradel:
### Description

- Add label with ```New``` for new applications.
- Just update ```hasUpdate``` prop in the applications list to true and ```releaseDate``` to see changes.
- Add badge for new updates in sidebar.

### Changes

image
image

### Test cases

image
- return ```true``` if the ```releaseDate``` is within 30 days.
- return ```false``` if the ```releaseDate``` was from more than 30 days.
- return ```true``` if ```releaseDate``` is today.

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/3179

### Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation


packages/playground/src/components/applications/ApplicationCards.vue
```diff
@@ -43,21 +58,93 @@ export default defineComponent({
...
     const baseURL = import.meta.env.BASE_URL;
+    const filteredCards = ref(props.cards);
+
+    function isCreatedToday(timestamp: number) {
+      const today = new Date();
+      const startOfDay = new Date(today.getFullYear(), today.getMonth(), today.getDate());
+      const endOfDay = new Date(today.getFullYear(), today.getMonth(), today.getDate(), 23, 59, 59, 999);
+
+      return timestamp >= startOfDay.getTime() && timestamp <= endOfDay.getTime();
+    }
```
Mahmoud-Emad: Can't you check if the `createdAt.day === today.day`?

packages/playground/src/views/solutions_view.vue
```diff
@@ -30,33 +30,43 @@ export default {
           "Nostr is a simple, open protocol that enables global, decentralized, and censorship-resistant social media.",
         icon: "nostr.png",
         route: DashboardRoutes.Applications.Nostr,
+        isNew: false,
+        createdAt: 0, // Update creation time in milliseconds ex: 1723468996310 is 'Mon Aug 12 2024 16:23:16'
```
Mahmoud-Emad: If you choose to save the createdAt field inside the card object, I guess you have to hard-code the date in place
```suggestion
        isNew: false,
        createdAt: 1723468996310,
```

Mahmoud-Emad (changes requested):
Good job ya Samar

zaelgohary (commented):
I tried to run the Dashboard multiple times but it lags on this page. I'm not sure why.

image


It happens only on this branch. I tried staging & different branches and they worked fine.

packages/playground/src/App.vue
```diff
@@ -248,7 +250,20 @@ function navigateToHome() {
...
+      const next30DaysInMs = 30 * 24 * 60 * 60 * 1000;
+      const timestamp = 1724769030739; // Update with the time of creation
+      const next30DaysFromNow = timestamp + next30DaysInMs;
+
+      if (item.hasUpdate == true) {
+        item.hasUpdate = Date.now() < next30DaysFromNow;
+      }
+    });
+  });
+});
```
zaelgohary: I think the issue I'm facing happens because of the change in onMounted. It worked once I've reverted the change.

packages/playground/src/views/solutions_view.vue
```diff
@@ -30,40 +30,52 @@ export default {
           "Nostr is a simple, open protocol that enables global, decentralized, and censorship-resistant social media.",
         icon: "nostr.png",
         route: DashboardRoutes.Applications.Nostr,
+        isNew: false,
+        createdAt: 1725283777169, // Update creation time in milliseconds ex: 1723468996310 is 'Mon Aug 12 2024 16:23:16'
```
zaelgohary: Instead of hard coding the timestamps, what if we created a function to create timestamps and call it here?

packages/playground/src/utils/types.ts
```diff
@@ -21,4 +21,6 @@ export interface ApplicationCard {
   icon: string;
   route: string;
   tags?: string[];
+  isNew: boolean;
+  createdAt: number;
```
zaelgohary: Please make this optional. This way, you won't need to add it to the old solutions.

zaelgohary: The createdAt I mean.

packages/playground/src/App.vue
```diff
@@ -253,6 +255,19 @@ onMounted(async () => {
   openProfile.value = true;
 });
 
+onMounted(() => {
+  routes.map(route => {
+    route.items.map(item => {
+      const next30DaysInMs = 30 * 24 * 60 * 60 * 1000;
+      const timestamp = 1724769030739; // Update with the time of creation
```
zaelgohary: This shouldn't be hardcoded.

packages/playground/src/App.vue
```diff
@@ -253,6 +255,19 @@ onMounted(async () => {
...
+      const next30DaysFromNow = timestamp + next30DaysInMs;
+      if (item.hasUpdate == true) {
+        item.hasUpdate = Date.now() < next30DaysFromNow;
+      }
+    });
+  });
+});
```
zaelgohary: Also, can you provide docstrings for the logic in the onMounted here?

packages/playground/src/utils/types.ts
samaradel: the old solutions could have releases too

packages/playground/src/views/solutions_view.vue
samaradel: I think it can't be dynamic as it will be called each time the component is mounted so it will get the date of the current day each time

MohamedElmdary:
@samaradel can't we have a simple way and a stronger typescript for dates for example
we can pass `{  year: 2024, month: "Sep", day: 4 }` instead of 1725454606447 which makes no sense just looking at it without understanding the deep details about project?

MohamedElmdary:
trying to understaind what is the usage of `hasUpdate` if the project has a date and that date within 30 days that enough to show `new` chip why do we need `hasUpdate` ?

samaradel:
@MohamedElmdary good point 👉

packages/playground/src/App.vue
```diff
@@ -253,6 +255,27 @@ onMounted(async () => {
...
+  const next30DaysInMs = 30 * 24 * 60 * 60 * 1000;
+  routes.map(route => {
+    route.items.map(item => {
+      const next30DaysFromNow = item.timestamp ? item.timestamp + next30DaysInMs : 0;
+      if (item.hasUpdate == true) {
+        item.hasUpdate = Date.now() < next30DaysFromNow;
+      }
+    });
+  });
+});
```
MohamedElmdary: this logic should be handled in applicationCard itself

packages/playground/src/components/applications/ApplicationCards.vue
```diff
@@ -1,10 +1,23 @@
...
-    <v-col v-for="card in cards" :key="card.title" cols="12" sm="12" md="12" lg="6" :xl="cards.length > 3 ? 4 : 6">
+    <v-col
+      v-for="card in filteredCards"
+      :key="card.title"
+      cols="12"
+      sm="12"
+      md="12"
+      lg="6"
+      :xl="filteredCards.length > 3 ? 4 : 6"
+    >
```
MohamedElmdary: please avoid style updates in code spots that has no code changes as it's very confusing

packages/playground/src/utils/types.ts
```diff
@@ -21,4 +21,6 @@ export interface ApplicationCard {
   icon: string;
   route: string;
   tags?: string[];
+  hasUpdate: boolean;
```
MohamedElmdary: can't get why we need `hasUpdate` also why it's required ... if it's not a prop in object so it's false

packages/playground/src/App.vue
samaradel: This part of adding badges in the sidebar, how could it be handled from the application part?

MohamedElmdary: ah my bad it's almost the same as application can we make a util which handling this logic if it's duplicated?

packages/playground/src/components/applications/ApplicationCards.vue
samaradel: it's handling the card sizes across all screens

amiraabouhadid:
Please fix build error

packages/playground/src/views/vms_view.vue
```diff
@@ -26,6 +26,7 @@ export default {
           "Deploy a full virtual machine on Threefold Grid. Full VM allows you to have a complete image with a custom kernel optimized for your own usecase.",
         icon: "vm.png",
         route: DashboardRoutes.VirtualMachines.FullVirtualMachine,
+        // releaseDate: { year: 2024, month: 9, day: 4 },
```
Mahmoud-Emad: Is that happen by mistake? the commented code

samaradel: No, it's the format to remind who will edit how it should be like

0oM4R: I think should be added in docs or, in doc string

samaradel: I think @Mahmoud-Emad is right, I should remove ```releaseDate``` comment it's interface job so there is no use for the doc string too

0oM4R: > I think @Mahmoud-Emad is right, I should remove `releaseDate` comment it's interface job so there is no use for the doc string

Agree, so is it mentioned on  the interface docstring?  if not please add it

samaradel: done

packages/playground/src/utils/date.ts
```diff
@@ -1,5 +1,15 @@
 import moment from "moment";
 
+import type { Date } from "./types";
 export default function toHumanDate(timeInSeconds: number): string {
   return moment(timeInSeconds * 1000).format("DD MMM YYYY, hh:mm A");
 }
+
+export function isReleasedOverMon(date: Date): boolean {
+  const releaseDate = new Date(date.year, date.month - 1, date.day);
```
amiraabouhadid: why do you re-assign the release date?
what if item.releaseDate is undefined as it is optional? Please wrap in try and catch

samaradel: I'm not reassigning it it's just a variable name. and it's inside if condition in the template

packages/playground/src/utils/types.ts
```diff
@@ -15,10 +15,37 @@ export interface IPublicConfig {
...
 
+/**
+ * Represents a date object with day, month, and year properties.
+ *
+ * @interface Date
+ * @property {number} day - The day of the month (1-31).
+ * @property {number} month - The month of the year (1-12).
+ * @property {number} year - The year.
+ */
+export interface Date {
```
xmonader: Why is this needed ? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date

packages/playground/src/utils/date.ts
```diff
@@ -1,5 +1,15 @@
...
+  const currentDate = new Date();
+  const next30Days = 24 * 60 * 60 * 1000;
```
xmonader: change the name to `ThirtyDaysInMilliseconds`

packages/playground/src/utils/date.ts
xmonader: I don't understand this logic. Specially subtracting 1 from the month of the passed date?


The function should be `differenceIsOverAMonth(Date1, Date2)` and when used you pass the current date and the release date

packages/playground/src/utils/date.ts
```diff
@@ -3,3 +3,27 @@ import moment from "moment";
...
+ * This function calculates the difference in days between the provided date and the current date.
+ * It then checks if this difference is less than or equal to 30 days.
+ *
+ * @example
+ * const date = new Date('2022-01-01');
+ * console.log(thirtyDaysInMilliseconds(date)); // Output: true or false depending on the current date
+ */
+
+export function thirtyDaysInMilliseconds(date: Date): boolean {
+  const releaseDate = new Date(date.getFullYear(), date.getMonth(), date.getDate());
```
xmonader: What is the point of this?

const releaseDate = date

packages/playground/src/utils/date.ts
```diff
@@ -3,3 +3,27 @@ import moment from "moment";
...
+ * @description
+ * This function calculates the difference in days between the provided date and the current date.
+ * It then checks if this difference is less than or equal to 30 days.
+ *
+ * @example
+ * const date = new Date('2022-01-01');
+ * console.log(thirtyDaysInMilliseconds(date)); // Output: true or false depending on the current date
+ */
+
+export function thirtyDaysInMilliseconds(date: Date): boolean {
```
xmonader: the name is wrong for this one, this one returns true if the difference between two dates (one of them is fixed as the current one and the other one is the release date)

so should be

isDifferenceBetweenDates30Days(date1, date2) and when you use it pass date1 as the current and date2 gets the value of the release date

packages/playground/src/utils/date.ts
```diff
@@ -12,14 +12,8 @@ export default function toHumanDate(timeInSeconds: number): string {
  * @returns {boolean} True if the release date is within the last 30 days, false otherwise.
  */
 
-export function isReleasedOverMon(releaseDate: Date, currentDate?: Date): boolean {
+export function isReleasedOverMon(releaseDate: Date, currentDate: Date): boolean {
   const thirtyDaysInMilliseconds = 24 * 60 * 60 * 1000;
```
xmonader: MillisecondsInADay

xmonader (approved):

Issue #3304, opened 2024-08-13
🐞 [Bug]: `TFChainError` is not displayed when the client stuck during the execution


Mahmoud-Emad:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### which package/s did you face the problem with?

Dashboard

### What happened?

Following the code changes suggested in [this PR discussion](https://github.com/threefoldtech/tfgrid-sdk-ts/pull/3207#discussion_r1705350580), we initially observed that the `Generic Type` handled the failure correctly but the `TFChain` failed. However, later on, @zaelgohary reported a change in the error pattern—neither KeyError nor TFChainError was being generated.

After further debugging with @0oM4R, we identified the root cause: the value of the property was being sent as a `string` or `undefined`, while the method expects an object in the format `{value: 'The actual value'}`. Additionally, we discovered that using `{enumerable: true}` when defining properties helps ensure that the attributes are displayed correctly when errors are thrown.

### Steps To Reproduce

1. Pull the development code
2. Create a `test.ts` file in the root of the TFChain package
3. Copy the attached script and run it 

### Script

```ts
import { Client } from "./src/client";
async function main() {
  const cl = new Client({
    url: "wss://tfchain.dev.grid.tf",
    mnemonicOrSecret: "<>",
  });
  await cl.connect();
  const contract = await (
    await cl.contracts.setDedicatedNodeExtraFee({
      nodeId: 73,
      extraFee: 10,
    })
  ).apply();
  console.log({ contract });
  await cl.disconnect();
}

main();
```

### which network/s did you face the problem on?

Dev

### version

 8c2f4b6

### Twin ID/s

143

### Relevant screenshots/screen records

![image](https://github.com/user-attachments/assets/8620b43b-7336-4e94-8e5a-d6cf42d71e81)

### Relevant log output

```console
NA
```

Mahmoud-Emad:
## Issue Update

### Work done

#### Changes
- Send the value of the prop that I want to define as an object that contains a 'value' attr
- Set the 'enumerable: true' to display the error args

PR #3305, opened 2024-08-13
Define the value argument in the TFChainError constructor

Mahmoud-Emad:
### Description

Following the code changes suggested in [this PR discussion](https://github.com/threefoldtech/tfgrid-sdk-ts/pull/3207#discussion_r1705350580), we initially observed that the `Generic Type` handled the failure correctly but the `TFChain` failed. However, later on, @zaelgohary reported a change in the error pattern—neither KeyError nor TFChainError was being generated.

After further debugging with @0oM4R, we identified the root cause: the value of the property was being sent as a `string` or `undefined`, while the method expects an object in the format `{value: 'The actual value'}`. Additionally, we discovered that using `{enumerable: true}` when defining properties helps ensure that the attributes are displayed correctly when errors are thrown.

### Changes

- Send the value of the prop that I want to define as an object that contains a 'value' attr
- Set the 'enumerable: true' to display the error args

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/3304

###

image

### Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation

### DIFF
--- a/packages/tfchain_client/src/errors.ts
+++ b/packages/tfchain_client/src/errors.ts
@@ -28,27 +28,18 @@ class TFChainError extends Error {
     super(options.message);
     this.name = "TFChainError";
 
-    if (options.keyError) {
-      Object.defineProperty(this, "keyError", { value: options.keyError, enumerable: true });
-    } else {
-      Object.defineProperty(this, "keyError", { value: "GenericError", enumerable: true });
+    if (!options.keyError) {
+      options.keyError = "GenericError";
     }
 
-    if (options.section) {
-      Object.defineProperty(this, "section", { value: options.section, enumerable: true });
-    }
-
-    if (options.args) {
-      Object.defineProperty(this, "args", { value: options.args, enumerable: true });
-    }
-
-    if (options.method) {
-      Object.defineProperty(this, "method", { value: options.method, enumerable: true });
-    }
-
-    if (options.docs) {
-      Object.defineProperty(this, "docs", { value: options.docs, enumerable: true });
-    }
+    Object.keys(options).forEach(key => {
+      if (options[key as keyof ITFChainError]) {
+        Object.defineProperty(this, key, {
+          value: options[key as keyof ITFChainError],
+          enumerable: true,
+        });
+      }
+    });
   }
 }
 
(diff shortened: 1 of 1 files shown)


packages/tfchain_client/src/errors.ts
```diff
@@ -27,12 +27,28 @@ class TFChainError extends Error {
...
+      Object.defineProperty(this, "args", { value: options.args, enumerable: true });
+    }
+
+    if (options.method) {
+      Object.defineProperty(this, "method", { value: options.method, enumerable: true });
+    }
+
+    if (options.docs) {
+      Object.defineProperty(this, "docs", { value: options.docs, enumerable: true });
+    }
```
zaelgohary: Instead of the repitition, can we do sth like:

```suggestion
    const { keyError = "GenericError", section, args, method, docs } = options;

    Object.defineProperties(this, {
      keyError: { value: keyError, enumerable: true },
      section: { value: section, enumerable: true },
      args: { value: args, enumerable: true },
      method: { value: method, enumerable: true },
      docs: { value: docs, enumerable: true },
    });
```

zaelgohary: Will discard the comment as we need to define properties only if its correspondent option is defined.

zaelgohary (approved):

packages/tfchain_client/src/errors.ts
AhmedHanafy725: I didn't test it but maybe something like that
```ts
    if (!options.keyError) {
      options.keyError = "GenericError"
    }
    for (const key of ["keyError", "section", "args", "method", "docs"]) {
      if (options[key]) {
        Object.defineProperty(this, key, { value: options[key], enumerable: true });
      }
```

Mahmoud-Emad: Pretty well. I also think that defining the key dynamically would enhance the code's scalability for future use.

```ts
class TFChainError extends Error {
  constructor(options: ITFChainError) {
    super(options.message);
    this.name = "TFChainError";

    if (!options.keyError) {
      options.keyError = "GenericError";
    }

    Object.keys(options).forEach(key => {
      if (options[key as keyof ITFChainError]) {
        Object.defineProperty(this, key, {
          value: options[key as keyof ITFChainError],
          enumerable: true,
        });
      }
    });
  }
}
```

AhmedHanafy725 (approved):

Issue #3202, opened 2024-08-02
(Mainnet verification) Error when voting on DAO🐞 [Bug]: 


sabrinasadik:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### which package/s did you face the problem with?

Dashboard

### What happened?

Got error 'vote failed' when trying to vote on a DAO proposal. The vote did went through tho. Multiple farmers have shared that the same thing happened to them. 

My nodes on the farm are active so that's not the problem

### Steps To Reproduce

1. dashboard.grid.tf (mainnet)
2. TFChain
3. DAO
4. Vote
5. Error: 'Vote Failed!' but did go through after checking tfchain

### which network/s did you face the problem on?

Main

### version

3.14

### Twin ID/s

785

### Node ID/s

6344, 4245, 6357

### Farm ID/s

452

### Relevant screenshots/screen records

![photo_2024-08-02 10 25 57](https://github.com/user-attachments/assets/82963683-40c5-4b09-bd48-9a0872f7810d)

### Relevant log output

```console
-
```

0oM4R:
Issue status: It is not clear what is the error here, will give it **2 hours** to investigate.

0oM4R:
Issue Status: the same happens to me if i vote on the same dao twice, 


solution: will handle this error and show new error msg, also if the eror not related to the duplicate vote, we will log the error,
will take about **1 hour**

0oM4R:
Issue status: 
work completed
- add log line in side the error handling to make it clear what exactly happens.
- handle the duplicate vote error, and clarify the toast message with it

A-Harby:
It's working fine on Devnet, 448a213.
![image](https://github.com/user-attachments/assets/fa22b082-a9b5-4245-a01f-ca7045ab4ed5)

PR #3317, opened 2024-08-19
clarify error in DAO page

0oM4R:
### Description

we are not sure exactly what happened  in the issue, but I assume that the user tried to vote twice; so we handled this case and added logs to make it easy to debug later 

### Changes

- clearify error in dao by logging the error.
- clearify the duplicate vote error by rephrasing the toast message
Screenshot from 2024-08-19 16-38-22

### Related Issues

- #3202

### Documentation PR

### Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation

### DIFF
--- a/packages/playground/src/dashboard/dao_view.vue
+++ b/packages/playground/src/dashboard/dao_view.vue
@@ -376,7 +376,7 @@ async function castVote() {
       let errMsg = `Vote Failed!`;
 
       if (err instanceof TFChainError && err.keyError == "DuplicateVote") {
-        errMsg = `Failed to vote; Already voted`;
+        errMsg = "Failed to vote. You have already voted.";
       }
       console.error(errMsg, err);
 
--- a/packages/playground/src/dashboard/dao_view.vue
+++ b/packages/playground/src/dashboard/dao_view.vue
@@ -288,7 +288,7 @@
 </template>
 <script lang="ts" setup>
 import type { GridClient } from "@threefold/grid_client";
-import type { DaoProposalDetails, DaoProposals } from "@threefold/tfchain_client";
+import { DaoProposalDetails, DaoProposals, TFChainError } from "@threefold/tfchain_client";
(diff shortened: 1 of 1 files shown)


amiraabouhadid:
@0oM4R fix build fail

samaradel (approved):

packages/playground/src/dashboard/dao_view.vue
```diff
@@ -373,7 +373,14 @@ async function castVote() {
       });
       openVDialog.value = false;
     } catch (err) {
-      createToast(`Vote Failed!`, {
+      let errMsg = `Vote Failed!`;
+
+      if (err instanceof TFChainError && err.keyError == "DuplicateVote") {
+        errMsg = `Failed to vote; Already voted`;
```
amiraabouhadid: should be "Failed to vote. You have already voted."

amiraabouhadid (changes requested):

packages/playground/src/dashboard/dao_view.vue
```diff
@@ -373,7 +373,14 @@ async function castVote() {
...
+      }
+      console.error(errMsg, err);
```
Mahmoud-Emad, line 381: Possible to be another error comes from the chain or another block, let's display an `error message/default error message` in the else block

Mahmoud-Emad (changes requested):
Good job ya 3omar

Mahmoud-Emad (approved):

packages/playground/src/dashboard/dao_view.vue
0oM4R: console.error will print on all cases whatever it is a `tfchainError` or not; the check `if (err instanceof TFChainError && err.keyError == "DuplicateVote")` is just for overwrite the errMsg with a specific one

amiraabouhadid (approved):

Issue #3147, opened 2024-07-23
🐞 [Bug]: email should be retrieved if the mnemonic are ok during login and its validation needs fixing


xmonader:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### which package/s did you face the problem with?

Dashboard

### What happened?

1- I expect the dashboard to fetch the email for me and fill the form if the mnemonic were okay
2- if i try to submit the form without the email filled in (and it's mandatory, i should see a validation error)

### Steps To Reproduce

I already entered my email before, now in a new browser i try to login. I paste my mnemonic and and encryption password and I don't remember the email.

### which network/s did you face the problem on?

Main

### version

3.14

### Relevant screenshots/screen records

![image](https://github.com/user-attachments/assets/98ba3f96-391c-44c4-8b50-ffda83ba5726)
no validation on email and yet the button is disabled 

### Relevant log output

```console
N/A
```

samaradel:
### Work Completed:
Load Email when mnemonic is entered if exists.

A-Harby:
Verified, Devnet cddbd85.

The email is now auto-written if the mnemonic has already entered the email value before.

![image](https://github.com/user-attachments/assets/334c11d4-ef5d-4a9e-907c-d6af8208f390)
![image](https://github.com/user-attachments/assets/b6067802-6986-442d-bd5e-7ac0120aa823)

[TC2113 - Connect Wallet](https://app.testlodge.com/a/26076/projects/40893/suites/234374?expand_section=422331#case_4150322)

A-Harby:
The input is stuck loading. 

![image](https://github.com/user-attachments/assets/91cd818b-9016-4fc9-9e08-a9717a5fb83b)


`Error: TwinNotExistError: Couldn't find a user for the provided mnemonic on dev network.`

samaradel:
### Work in Progress (WIP):
Investigating in checking if the mnemonic if it already has an account or not

samaradel:
### Work Completed:
Fix Email loading when the profile doesn't have ```twinId```

A-Harby:
Verified, Devnet 4829e02.

IF the twin is not registered on this network or doesn't have an email attached to it, it won't show an error anymore.
![image](https://github.com/user-attachments/assets/2082a34b-2106-4301-a05a-638b7980b097)

PR #3339, opened 2024-08-25
Check if the mnemonic has `twinId` before connect

samaradel:
### Description

- Check ```mnemonic``` if it has ```twinId``` or not, if no it will be redirected to create an account, otherwise it will load email from metadata normally.
### Changes

[Screencast from 25-08-24 23:52:11.webm](https://github.com/user-attachments/assets/789a6072-0c55-4668-81b8-f4de060f3e88)

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/3147#issuecomment-2291179960

### Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation

### DIFF
--- a/packages/playground/src/weblets/profile_manager.vue
+++ b/packages/playground/src/weblets/profile_manager.vue
@@ -193,11 +193,7 @@
                           color="secondary"
                           variant="outlined"
                           :disabled="
-                            (isValidForm ||
-                              !!mnemonic ||
-                              shouldActivateAccount ||
-                              keypairType === KeypairType.ed25519) &&
-                            !shouldCreateNewAccount
+                            isValidForm || !!mnemonic || shouldActivateAccount || keypairType === KeypairType.ed25519
                           "
                           :loading="creatingAccount"
                           @click="openAcceptTerms = termsLoading = true"
@@ -267,7 +263,7 @@
                   v-model="email"
                   v-bind="props"
                   :loading="loadEmail"
-                  :disabled="creatingAccount || activatingAccount || activating || loadEmail || shouldCreateNewAccount"
+                  :disabled="creatingAccount || activatingAccount || activating || loadEmail"
                   readonly
                   ref="emailRef"
                   @focus="handleFocus(emailRef)"
@@ -296,7 +292,7 @@
                           label="Password"
                           v-model="password"
                           v-bind="{ ...passwordInputProps, ...validationProps }"
-                          :disabled="creatingAccount || activatingAccount || activating || shouldCreateNewAccount"
+                          :disabled="creatingAccount || activatingAccount || activating"
                           autocomplete="off"
                         />
                       </div>
@@ -319,7 +315,7 @@
                       ...confirmPasswordInputProps,
                       ...validationProps,
                     }"
-                    :disabled="creatingAccount || activatingAccount || activating || shouldCreateNewAccount"
+                    :disabled="creatingAccount || activatingAccount || activating"
                     autocomplete="off"
                   />
                 </InputValidator>
@@ -341,7 +337,6 @@
                   !isValidForm ||
                   creatingAccount ||
                   activatingAccount ||
-                  shouldCreateNewAccount ||
                   (activeTab === 1 && isValidConnectConfirmationPassword)
                 "
               >
@@ -611,7 +606,6 @@ const isValidForm = ref(false);
 const mnemonicInput = useInputRef();
 
 const isNonActiveMnemonic = ref(false);
-const shouldCreateNewAccount = ref(false);
 const shouldActivateAccount = computed(() => {
   if (!mnemonic.value) return false;
   return isNonActiveMnemonic.value;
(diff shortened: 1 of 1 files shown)


amiraabouhadid (commented):
I get this instead. When I used a bip39 generated mnemonic, I got asked to agree to terms and conditions and then this error shows up. 
image

amiraabouhadid (commented):
also these console errors keep showing up 
image

samaradel:
@amiraabouhadid It was because you used more than 66 characters, there's no limit now

amiraabouhadid (commented):
why does it change the mnemonic? 
https://www.loom.com/share/cc2e10b335de479d8c6b2ea27e286243?sid=5c40ec53-f4eb-403a-b99d-3b22c485cf4d

samaradel:
@amiraabouhadid I assume you used a new one so it blocked the other fields till you create an account

amiraabouhadid:
> @amiraabouhadid I assume you used a new one so it blocked the other fields till you create an account

Yes but then it replaces the mnemonic i placed, I don't think it should do that

packages/playground/src/weblets/profile_manager.vue
```diff
@@ -193,7 +190,11 @@
                           color="secondary"
                           variant="outlined"
                           :disabled="
-                            isValidForm || !!mnemonic || shouldActivateAccount || keypairType === KeypairType.ed25519
+                            (isValidForm ||
+                              !!mnemonic ||
+                              shouldActivateAccount ||
+                              keypairType === KeypairType.ed25519) &&
+                            !shouldCreateAccount
```
0oM4R: if the user entered a valid mnemonic, and doesn't have twin, the expected behavior is to click on connect then and it will create an account with the provided mnemonic.


[Screencast from 2024-09-01 11-31-34.webm](https://github.com/user-attachments/assets/9cf55029-7c08-4855-a15a-134e19417472)


on dashboard 2.5.0


[Screencast from 2024-09-01 11-32-32.webm](https://github.com/user-attachments/assets/b0560e5e-6fe5-449b-8c6d-2c1945530161)

0oM4R (changes requested):

samaradel:
@xmonader can you confirm the expected behavior, please

packages/playground/src/weblets/profile_manager.vue
```diff
@@ -906,6 +904,20 @@ async function getEmail() {
   email.value = await readEmail(grid!);
   loadEmail.value = false;
 }
+async function hasTwinId() {
+  try {
+    await getGrid({ mnemonic: mnemonic.value, keypairType: keypairType.value });
```
Mahmoud-Emad: now i can create and activate a new account with no issues

[Screencast from 09-02-2024 10:36:27 AM.webm](https://github.com/user-attachments/assets/d357500d-b363-48dd-8f2b-3d750b4b540c)


but I noticed a small change maybe cause an unnecessary request, in the `hasTwinId` function you call the `getGrid` function and then call the `getEmail` function which calls the `getGrid` function internally, let's remove this call if not needed, please.


```ts
async function getEmail() {
  loadEmail.value = true;
  const grid = await getGrid({ mnemonic: mnemonic.value, keypairType: keypairType.value });
  email.value = await readEmail(grid!);
  loadEmail.value = false;
}
async function hasTwinId() {
  try {
    await getGrid({ mnemonic: mnemonic.value, keypairType: keypairType.value }); // Remove this call
    await getEmail();
  } catch (e) {
    if (e instanceof TwinNotExistError) {
      isNonActiveMnemonic.value = true;
    } else {
      return {
        message: normalizeError(e, "Something went wrong. please try again."),
      };
    }
  }
}
```

Mahmoud-Emad (changes requested):
Good job ya Samar

packages/playground/src/weblets/profile_manager.vue
```diff
@@ -906,6 +904,20 @@ async function getEmail() {
...
+  } catch (e) {
+    if (e instanceof TwinNotExistError) {
+      isNonActiveMnemonic.value = true;
+    } else {
+      return {
+        message: normalizeError(e, "Something went wrong. please try again."),
+      };
+    }
+  }
+}
```
0oM4R: I suggest to refactor the `readEmail` instead to have try and catch.
```ts 
async function getEmail() {
  loadEmail.value = true;
  try {
    const grid = await getGrid({ mnemonic: mnemonic.value, keypairType: keypairType.value });
    if (grid) email.value = await readEmail(grid);
  } catch (e) {
    if (!(e instanceof TwinNotExistError)) {
      return {
        message: normalizeError(e, "Something went wrong. please try again."),
      };
    }
  } finally {
    loadEmail.value = false;
  }
}
```
 The idea is we are already handling the hasTwinId logic in storeAndLogin function;
 If we could just add add try and catch to the get email function, it will be better, i think;
what do you think about this ?

Mahmoud-Emad (approved):
Go on

0oM4R (approved):
Please make sure to rename the PR before merge

Issue #3294, opened 2024-08-12
Enhance the Networks component


Mahmoud-Emad:
### Is your feature request related to a problem? Please describe

N/A

### Describe the solution you'd like

In the current implementation, the network configurations (e.g., IPv4, IPv6, Mycelium, Planetary, WireGuard) are managed individually within each parent component. This approach involves defining constants as references in the parent component and binding these values to the child `Networks` component. Here’s an example:

```vue
// Algorand.vue
<Networks
    v-model:mycelium="mycelium"
    v-model:planetary="planetary"
    v-model:ipv4="ipv4"
    v-model:ipv6="ipv6"
     v-model:wireguard="wireguard"
/>
```

```ts
const ipv4 = ref(false);
const ipv6 = ref(false);
const wireguard = ref(false);
const planetary = ref(false);
const mycelium = ref(true);

const vm = await deployVM(grid!, {
  machines: [
    {
      // ...
      publicIpv4: ipv4.value,
      publicIpv6: ipv6.value,
      mycelium: mycelium.value,
      planetary: planetary.value,
      //...
    },
  ],
});
```

So, imagine, if we want to enable only the `mycelium` to all of the solutions/applications, we need to go around all components change the value of all networks then enable only the `mycelium` network, so hard? No?

My proposal is to modify the `Networks` component and let it expose the values, e.g. defining a store to store the actual values of the networks, also setting the initial value in this store will reflect to all of the solutions

**Define network settings as reactive properties in this store.**
```ts
// networkStore.ts
import { reactive } from 'vue';

export const networkStore = reactive({
  ipv4: false,
  ipv6: false,
  wireguard: false,
  planetary: false,
  mycelium: true,
});

```
**Update the Networks Component:**

```vue
<!-- Networks.vue -->
<template>
  <!-- Bind the network settings to the appropriate UI elements -->
</template>

<script setup>
defineProps({
  networks: Object,
});
</script>
```

**Bind the Centralized Configuration to Components:**

```vue
<!-- Algorand.vue -->
<Networks :networks="networkStore" />
```

**Access and Use the Configuration in VM Deployment:**

```ts
const vm = await deployVM(grid!, {
  machines: [
    {
      // ...
      publicIpv4: networkStore.ipv4,
      publicIpv6: networkStore.ipv6,
      mycelium: networkStore.mycelium,
      planetary: networkStore.planetary,
      //...
    },
  ],
});
```

[...]

PR #3344, opened 2024-08-28
Enhance networks Component

MohamedElmdary:
### Description
Enhance networks Component

### Changes
- Make component more simpler
- Add an easier way to tell if it's readonly or not

```ts

// by default all networks `false` but mycelium
// pass object to update any default like below
const { ipv4, ipv6, wireguard, planetary, mycelium } = useNetworks({ ipv4: true });
```

```vue
<Networks v-model:mycelium="mycelium" v-model:planetary="planetary" :ipv4="ipv4" v-model:ipv6="ipv6" />
```
image

### Requirements
the requirements for the network component: 
1. Public IPv4: `An Internet Protocol version 4 address that is globally unique and accessible over the internet.`
2. Public IPv6: `Public IPv6 is the next-generation Internet Protocol that offers an expanded address space to connect a vast number of devices.`
3. Planetary Network: `The Planetary Network is a distributed network infrastructure that spans across multiple regions and countries, providing global connectivity.`
4. Mycelium: `Mycelium is an IPv6 overlay network. Each node that joins the overlay network will receive an overlay network IP.`
5. Add wireguard access: `Enabling WireGuard Access allows you to establish private, secure, and encrypted connections to your instance.
6. Should be displayed as switches in the same mentioned order
7. Myceluim should be enabled by default
8. Should have tooltip icon with the mentioned descriptions; in some cases will it will be conditional, will be mentioned later.
9. the user must enable at least one network option, if the user enabled only one option, and tried to disable it should we show an error alert with message  You must enable at least one of network options. Also the outer panel should have worining icon.
10. If the solution have gateway; 
- the wireguard option should be enabled and readonly.
- If the user enabled `IPv4` and `custom domain`, the wireguard option be optional not readonly .
12. if solution have enabled smtp: ipv4 should be enabled and readonly
13. caprover must have ipv4 enabled , readonly
14. node pilot: enabled ipv4,ipv6 readonly

some tooltips will bs changed based on conditions:
 - ipv4 : 
    - if smtp is enabled: `An Internet Protocol version 4 address that is globally unique and accessible over the internet. Please note that this field will be read-only as SMTP is enabled.`
 - wireguard:
    - if domain required, without custom domain and without ipv4: `Enabling WireGuard Access allows you to establish private, secure, and encrypted instance connections. Please note that this field will be read-only unless you use a custom domain with IPV4.`

### Related Issues
- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/3294

### Documentation PR

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation


packages/playground/src/components/networks.vue
```diff
@@ -121,9 +122,22 @@ export default {
...
     });
 
+    const readonlyWireguard = computed(() => props.hasCustomDomain && !props.ipv4);
+    watch(
+      readonlyWireguard,
+      readonly => {
+        const fn = attrs["onUpdate:wireguard"];
+        if (readonly && !props.wireguard && typeof fn === "function") {
+          fn(true);
+        }
```
Mahmoud-Emad: The Wireguard shouldn't be read-only in the solutions that don't require domain

image

also, the Wireguard option should be read-only even if the `custom domain` is disabled

image


PS: Please note that this field will be read-only unless you use a custom domain with IPV4.

Mahmoud-Emad (changes requested):
Good job ya MR-Rabeeee3

Mahmoud-Emad:
@all
Please let's merge https://github.com/threefoldtech/tfgrid-sdk-ts/pull/3287 first

packages/playground/src/weblets/micro_vm.vue
```diff
@@ -245,7 +244,6 @@ const mycelium = ref(true);
 const wireguard = ref(false);
```
Mahmoud-Emad: While we enhance the network component, I would recommend refactoring the weblets by removing all network consents

it could be simply 

```suggestion
function networks(){
    return {
        planetary, mycelium, publicIpv4, publicIpv6
    } as ref()
}

const networks = createNetworks()
const vm = await deployVM(grid!, {
      planetary: networks.planetary.value,
      mycelium: networks.mycelium.value,
      publicIpv4: networks.ipv4.value,
      publicIpv6: networks.ipv6.value,
})
```

PS: Me and @AhmedHanafy725 agreed [in](https://github.com/threefoldtech/tfgrid-sdk-ts/pull/3287) on adding all networks to all solutions and enabling the `Mycelium` 
 option by default

please check the Jenkins solution

image

packages/playground/src/components/networks.vue
```diff
@@ -193,26 +122,29 @@ export default {
...
     });
 
-    const isWireguardReadOnly = computed(() => !(props.domain?.enabledCustomDomain && props.ipv4 === true));
+    const wireguardTooltip = computed(() =>
+      props.hasCustomDomain
+        ? "Enabling WireGuard Access allows you to establish private, secure, and encrypted instance connections. Please note that this field will be read-only unless you use a custom domain with IPV4."
+        : "Enabling WireGuard Access allows you to establish private, secure, and encrypted instance connections.",
+    );
 
+    const readonlyWireguard = computed(() => props.hasCustomDomain && !props.ipv4);
```
Mahmoud-Emad: as we agreed, this should be read-only in all solutions required domain and hasn't domain 

- image

Mahmoud-Emad (changes requested):

packages/playground/src/components/networks.vue
```diff
@@ -1,154 +1,109 @@
...
-    "update:planetary": (value?: boolean) => value,
-    "update:mycelium": (value?: boolean) => value,
-    "update:wireguard": (value?: boolean) => value,
+    required: { type: Boolean, default: () => false },
+    ipv4: { type: Boolean, default: () => null },
+    ipv6: { type: Boolean, default: () => null },
+    planetary: { type: Boolean, default: () => null },
+    mycelium: { type: Boolean, default: () => null },
+    wireguard: { type: Boolean, default: () => null },
+    hasCustomDomain: { type: Boolean, default: () => false },
```
0oM4R, line 102: please note that the only case we have wg not `readOnly` is 
**having custom domain with ipv4 enalbled** 


[Screencast from 2024-09-01 14-57-56.webm](https://github.com/user-attachments/assets/4d4e0c5c-5f25-48e9-86e8-ddd5e17a086d)

0oM4R (changes requested):

packages/playground/src/components/networks.vue
MohamedElmdary: shouldn't we enable it if user didn't ask for custom domain at all?

0oM4R: no if we have custom domain without ipv4 we have to create fqdn record on gw node, but if we have ipv4 and custom domain the user should map the machine ip to the custom domain 
please check https://github.com/threefoldtech/tfgrid-sdk-ts/pull/3287#issuecomment-2301993915

Mahmoud-Emad (approved):
Teslam edak ya Rabe3

packages/playground/src/weblets/jenkins.vue
```diff
@@ -133,10 +133,7 @@ const flist: Flist = {
 };
 const dedicated = ref(false);
 const certified = ref(false);
-const ipv4 = ref(false);
-const ipv6 = ref(false);
-const mycelium = ref(true);
-const planetary = ref(true);
+const { ipv4, ipv6, mycelium, planetary } = useNetworks({ planetary: true });
```
zaelgohary: require-domain should be added to the Networks tag in template.

packages/playground/src/components/networks.vue
```diff
@@ -1,154 +1,109 @@
...
-    "update:planetary": (value?: boolean) => value,
-    "update:mycelium": (value?: boolean) => value,
-    "update:wireguard": (value?: boolean) => value,
+    required: { type: Boolean, default: () => false },
+    ipv4: { type: Boolean, default: () => null },
+    ipv6: { type: Boolean, default: () => null },
+    planetary: { type: Boolean, default: () => null },
+    mycelium: { type: Boolean, default: () => null },
+    wireguard: { type: Boolean, default: () => null },
+    requireDomain: { type: Boolean, default: () => false },
```
zaelgohary, line 101: Is this needed? selection details has `require-domain`.

0oM4R:
[Screencast from 2024-09-02 14-26-43.webm](https://github.com/user-attachments/assets/d98b98ca-e80c-44b5-9d45-cfd60fb3027b)


the only case we should have the wg not readonly is: when having ipv4 **with** customDomain enabled

packages/playground/src/weblets/jenkins.vue
MohamedElmdary: 1. require-domain isn't a network so it shouldn't be used in `useNetworks`
2. `require-domain` won't change for specific solution (solution will need domain or not)

zaelgohary: Indeed but it is only useful for the tooltip it shows here:

https://github.com/threefoldtech/tfgrid-sdk-ts/blob/30b685a62cfda801ab6cf2015196981cde47ad12/packages/playground/src/components/networks.vue#L143-L147

MohamedElmdary: it's also used as readonly for wireguard

packages/playground/src/weblets/freeflow.vue
```diff
@@ -76,6 +77,7 @@
           rootFilesystemSize,
           ssdDisks: disks.map(d => d.size),
         }"
+        :has-custom-domain="selectionDetails?.domain?.enabledCustomDomain"
```
0oM4R: Why do we pass this to selectionDetails, I couldn't find it on the component props can you explain please?

packages/playground/src/components/networks.vue
```diff
@@ -159,29 +115,21 @@ export default {
     ) {
       throw new Error("You must provide at least one network  option");
     }
-    const error = computed(
-      () => !(props.ipv4 || props.ipv6 || props.planetary || props.wireguard || props.mycelium) && props.required,
-    );
-    expose({
-      error,
+    const error = computed(() => {
```
0oM4R, line 118: on k8s  I was able to disable all networks and click on deploy without any errors 
image

packages/playground/src/components/networks.vue
```diff
@@ -1,154 +1,110 @@
...
+          <v-chip v-if="mycelium" variant="outlined" class="ml-2" text="Mycelium" />
+          <v-chip v-if="wireguard" variant="outlined" class="ml-2" text="Wireguard" />
+        </v-expansion-panel-title>
+
+        <v-expansion-panel-text>
+          <NetworkItem
+            tooltip-text="An Internet Protocol version 4 address that is globally unique and accessible over the internet."
+            label="Public IPv4"
+            :value="$props.ipv4"
+            :emit-function="$attrs['onUpdate:ipv4']"
```
0oM4R: on caprover we need the ipv4 to be read only

0oM4R (dismissed):

packages/playground/src/weblets/freeflow.vue
MohamedElmdary: it's added by mistake using vscode `replace all` but it have no effect

packages/playground/src/components/networks.vue
MohamedElmdary: this is because it's not required but I checked early versions and it wasn't require too should I turn it into required?

packages/playground/src/components/networks.vue
MohamedElmdary: should be fixed

0oM4R:
In k8s worker and leader i still can disable all network options
as we decided in the BAM we have to change this behavior, not sure it was to force wg or mycelium 
Screenshot from 2024-09-11 15-09-21

0oM4R:
also I was able to deploy an algorand instance while disable all network options,

0oM4R:
image
 is this normal we don't have wg at all in Jenkins

Issue #3340, opened 2024-08-26
[Qanet] Ensure compatibility with TFChain 2.9 release


sameh-farouk:
### Is your feature request related to a problem? Please describe

Ensure compatibility with TFChain 2.9 release / TFChain runtime v152

### Describe the solution you'd like

**For better context**
PR: https://github.com/threefoldtech/tfchain/pull/992

**Relevant SDK Adjustments Needed:**

**Fund Reservations**: The system now employs a more reliable reservation method, replacing the previous use of locks/freezes.

**Balance Handling**: Payments on hold are no longer considered part of the free balance. The total balance should now reflect both the reserved and free portions. Previously, the free balance was considered the total balance, with amounts reserved for future payments subtracted from it. The new approach aligns with the balances pallet terminology:

  - Total Balance: The sum of the reserved and free balances in the user's account.
  - Reserved: Funds still owned by the account holder but set aside (e.g., for future contract payments).
  - Free: The portion of the balance that is available for spending, minus the Existential Deposit, assuming no locks are present.
  - Locked: A freeze on a specific amount of the free balance, no longer used in TFChain smart contract payments.

**Contract Payment Tracking**: The ContractLock has been replaced with ContractPaymentState to better track and manage the states of contract payments, particularly during grace periods and when dealing with overdue payments.
https://github.com/threefoldtech/tfchain/blob/622e7d255b9c76203fba8a316b84b280b59fbba7/substrate-node/pallets/pallet-smart-contract/src/types.rs#L219-L224

https://github.com/threefoldtech/tfchain/blob/83ccaa45d86076f6730d3611ce0c826c69dce649/substrate-node/pallets/pallet-smart-contract/src/types.rs#L236-L243

**Dashboard Restore Contracts Feature**: After the update to the new runtime, the dashboard's "restore contracts" feature will no longer be usable. To correctly reinstate this functionality, the following steps are needed:

  - Overdraft Tracking: The system now tracks overdrafts per contract. The overdraft is updated every hour or when billing is triggered.
  - Calculating Overdue Amounts: To calculate the overdue amount, sum the overdraft in the contract payment state with the product of ((seconds elapsed since the last billing + an extra time allowance for the user to take action) * the contract cost per second). This will give the correct overdue amount that needs to be addressed.
  
**Update:**
**Possibly less relevant changes::**
**Event Emission Changes**: The events emitted from the billing process have been updated, and several changes have been made:
- New Events:
[...]

PR #3345, opened 2024-08-28
Chain compatibility 

0oM4R:
-  no need to calculate the free balance, now we have the pure free field without the reserved balance.
- rename locked field with reserved.
- in profile manager added the free to the reserved to show total balance

### Description

Describe the changes introduced by this PR and what does it affect

### Changes

#3340 

### Related Issues

List of related issues

### Documentation PR

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/packages/grid_client/src/clients/tf-grid/contracts.ts
+++ b/packages/grid_client/src/clients/tf-grid/contracts.ts
@@ -2,6 +2,7 @@ import {
   Contract,
   ContractLock,
   ContractLockOptions,
+  ContractPaymentState,
   Contracts,
   ExtrinsicResult,
   GetDedicatedNodePriceOptions,
@@ -12,6 +13,8 @@ import { Decimal } from "decimal.js";
 import { formatErrorMessage } from "../../helpers";
 import { ContractStates } from "../../modules";
 import { Graphql } from "../graphql/client";
+
+const ONE_HOUR = 1000 * 60 * 60;
 
 export type DiscountLevel = "None" | "Default" | "Bronze" | "Silver" | "Gold";
 
@@ -98,6 +101,11 @@ export interface GetConsumptionOptions {
   id: number;
 }
 
+export interface CalculateOverdueOptions {
+  graphqlURL: string;
+  id: number;
+  paymentState: ContractPaymentState;
+}
 export interface CancelMyContractOptions {
   graphqlURL: string;
 }
@@ -108,6 +116,18 @@ export interface LockContracts {
   nodeContracts: LockDetails;
   rentContracts: LockDetails;
   totalAmountLocked: number;
+}
+export interface ContractOverdueDetails extends ContractPaymentState {
+  overdueAmount: number;
+}
+
+export type OverdueDetails = { [key: number]: ContractOverdueDetails };
+
+export interface ContractsOverdue {
+  nameContracts: OverdueDetails;
+  nodeContracts: OverdueDetails;
+  rentContracts: OverdueDetails;
+  totalOverdueAmount: number;
 }
 
 class TFContracts extends Contracts {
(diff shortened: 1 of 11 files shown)


packages/grid_client/src/clients/tf-grid/contracts.ts
```diff
@@ -13,6 +14,8 @@ import { formatErrorMessage } from "../../helpers";
 import { ContractStates } from "../../modules";
 import { Graphql } from "../graphql/client";
 
+const ONE_HOUR = 1000 * 60 * 60;
```
sameh-farouk, line 17: In what unit exactly? Please change the variable name to a more descriptive one.
An ideal name for a variable that holds the number of milliseconds in one hour could be `MILLISECONDS_PER_HOUR`, which is descriptive and clearly convey the purpose of the variable.

packages/grid_client/src/clients/tf-grid/contracts.ts
```diff
@@ -308,6 +328,30 @@ class TFContracts extends Contracts {
...
+   *
+   * The resulting overdue amount represents the amount that needs to be addressed.
+   *
+   * @param {CalculateOverdueOptions} options - The options containing the contract ID.
+   * @returns {Promise<number>} - The calculated overdue amount.
+   */
+  async calculateContractOverDue(options: CalculateOverdueOptions) {
+    const { lastUpdatedSeconds, standardOverdraft, additionalOverdraft } = options.paymentState;
+    const contractCost = await this.getConsumption({ id: options.id, graphqlURL: options.graphqlURL });
+    const elapsedSeconds = new Decimal(Date.now()).minus(lastUpdatedSeconds).div(1000);
```
sameh-farouk, line 347: Given that ContractPaymentState.lastUpdatedSecond is already a timestamp in seconds, you need to convert the current timestamp from milliseconds to seconds before performing the subtraction, right?

The corrected code would be
```
const elapsedSeconds = new Decimal(Date.now()).div(1000).minus(lastUpdatedSeconds);
```
`Date.now()` returns the current timestamp in milliseconds, then `.div(1000)`  converts the current time from milliseconds to seconds by dividing by 1000. finally `.minus(lastUpdatedSeconds)` subtracts lastUpdatedSeconds (which is already in seconds) from the current time in seconds.

packages/grid_client/src/clients/tf-grid/contracts.ts
```diff
@@ -308,6 +328,30 @@ class TFContracts extends Contracts {
...
+    const overdue = standardOverdraft
```
sameh-farouk, line 348: Can you please explain how the overdue is calculated, add a comment, and possibly simplify it by calculating it progressively using descriptive intermediary variable names?

packages/grid_client/src/clients/tf-grid/contracts.ts
```diff
@@ -308,6 +328,30 @@ class TFContracts extends Contracts {
...
+      .plus(additionalOverdraft)
+      .times(elapsedSeconds.plus(ONE_HOUR).times((contractCost / 60) * 60));
```
sameh-farouk, line 350: Units are not consistent here.
From what I see `elapsedSeconds` is in seconds, while `ONE_HOUR` is in milliseconds.
Perhaps you intended to set ONE_HOUR to 3600 (seconds) instead of 3600000 (milliseconds)?

Also, I'm trying to understand why we divide by 60 and then multiply by 60.

packages/grid_client/src/clients/tf-grid/contracts.ts
```diff
@@ -308,6 +328,30 @@ class TFContracts extends Contracts {
...
+
+    return Number(overdue.div(10 ** 7));
```
sameh-farouk, line 352: Again, add a comment to explain the division here, and/or first assign the result to a descriptive variable name like `overdueTFT` for better reliability, then return it. you can refer to the raw amount (before division) as `overdueMicroTFT`

packages/grid_client/src/clients/tf-grid/contracts.ts
```diff
@@ -308,6 +328,30 @@ class TFContracts extends Contracts {
...
+   * 2. The contract cost per second.
+   *
+   * The resulting overdue amount represents the amount that needs to be addressed.
+   *
+   * @param {CalculateOverdueOptions} options - The options containing the contract ID.
+   * @returns {Promise<number>} - The calculated overdue amount.
+   */
+  async calculateContractOverDue(options: CalculateOverdueOptions) {
+    const { lastUpdatedSeconds, standardOverdraft, additionalOverdraft } = options.paymentState;
+    const contractCost = await this.getConsumption({ id: options.id, graphqlURL: options.graphqlURL });
```
sameh-farouk, line 346: This is just an estimate, as some contracts might have public IP addresses, and the cost will depend on the amount of bandwidth used, which is likely to vary.

The cost calculation should simulate the billing process on tfchain (dry run) to provide a more accurate estimate. This involves checking the pricing policy, NU, and other resource consumption to determine the actual cost of the contract. Then, account for the time between the calculation and the submission of the transaction.

We still need to estimate the cost of this extra time, but it shouldn't be a problem. Since the cost changes dynamically, we should also update the cost accordingly. It's unlikely that we will receive a new NU report, especially if we refresh the cost at short intervals, such as every 10~20 seconds.

0oM4R:
changes on #3387

Issue #2410, opened 2024-03-12
🐞 [Bug]: Threefold logo in playground is too small (unreadable) at small screen sizes


amiraabouhadid:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### which package/s did you face the problem with?

Dashboard, UI

### What happened?

I reduced device screen size, and the threefold logo became unreadable 

### Steps To Reproduce

1. run yarn dev in playground
2. login/ sign up 
3. reduce screen size to 900x 845 or smaller
4. watch the logo on the left hand side in the toolbar become unreadable mambo jumbo 

### which network/s did you face the problem on?

Dev

### version

3.14

### Relevant screenshots/screen records

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/56790126/61db4e0f-6267-4287-88da-3ca46a77b8cd)

### Relevant log output

```console
no logs, UI issue
```

ehab-hassan:
Work complete

fixed at medium size and small size 
![image](https://github.com/user-attachments/assets/c561603b-a7d5-40a1-86f0-b67b581f3258)
![image](https://github.com/user-attachments/assets/e59f0cf8-3369-4f67-9440-16c966d12c1a)

ehab-hassan:
Work Completed

khaledyoussef24:
verified on devnet 
Version c2d6060

<img width="1680" alt="Image" src="https://github.com/user-attachments/assets/06f86f2f-830b-497c-a3cd-a2c9850e99fe" />
<img width="1680" alt="Image" src="https://github.com/user-attachments/assets/97bc678b-a959-4628-8915-dbe9e3722c08" />
<img width="1680" alt="Image" src="https://github.com/user-attachments/assets/b0fe9100-e574-4d19-8c0f-db73f6ac7407" />
<img width="1680" alt="Image" src="https://github.com/user-attachments/assets/db9ade2f-9433-4ff8-ba25-5b6bfc72d214" />

[testcase](https://app.testlodge.com/a/26076/projects/40893/suites/234374?expand_section=422331#case_4431572)

Issue #2342, opened 2024-03-05
Responsive design 


0oM4R:
### Description

Can we have a design for small screens, We need to make the current dashboard responsive and it will be great if we have a design to guide us

### Related issues 

- #1261 
- #1987 
- #1985 
- #2410

ehab-hassan:
WIP
Making Figma show the best way of showing the navbar on mobile

ehab-hassan:
WIP
Figma done
https://www.figma.com/proto/KAQ6497YfczErqfh1Bz62l/Dashboard2?node-id=808-1816&t=MlsBUVOeVNGm3Woy-0&scaling=scale-down&content-scaling=fixed&page-id=0%3A1&starting-point-node-id=1%3A1248&show-proto-sidebar=1

ehab-hassan:
WIP
almost done 
what needs to be fixed is
- simulator
- landing home without login

ehab-hassan:
Work Completed

khaledyoussef24:
verified on devnet
Version c2d6060

verified in 
https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1261
https://github.com/threefoldtech/tfgrid-sdk-ts/issues/2410
https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1985
also the simulator page : 
<img width="1680" alt="Image" src="https://github.com/user-attachments/assets/49f18a66-8fee-480d-b55a-497c9ef21560" />
<img width="1680" alt="Image" src="https://github.com/user-attachments/assets/c4bb9962-879b-423a-9254-c27835244203" />
<img width="1680" alt="Image" src="https://github.com/user-attachments/assets/a743a4a2-5c9f-466a-8ded-99e241c938c6" />
[testcase](https://app.testlodge.com/a/26076/projects/40893/suites/234374?expand_section=422331#case_4431573)

Issue #1987, opened 2024-01-15
Navbar isn't responsive 


A-Harby:
### Description

https://dashboard.bknd1.ninja.tf/#/tf-chain/twin, 2.3.0-alpha6

The logo, TFT price, and Get TFT need to be responsive to the different small screen sizes.

### Steps to reproduce

[Screencast from 01-15-2024 02:44:59 PM.webm](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/4d976a3b-8804-4daf-ac4f-77f73456c22f)

0oM4R:
will move this to blocked on the UI design #2342

ehab-hassan:
@AhmedHanafy725 need someone  to help me with Vue to apply this 

https://www.figma.com/proto/KAQ6497YfczErqfh1Bz62l/Dashboard2?node-id=808-1816&t=MlsBUVOeVNGm3Woy-0&scaling=scale-down&content-scaling=fixed&page-id=0%3A1&starting-point-node-id=1%3A1248&show-proto-sidebar=1

ehab-hassan:
Work Completed
work done by @Mahmoud-Emad Thanks 
will add to my PR

ehab-hassan:
@Mahmoud-Emad nav is still opening dropdown at the small screen can u check, please

[05662543-f754-4ed9-9d1d-11fb89c155b2.webm](https://github.com/user-attachments/assets/c4d68415-0964-40fb-8619-75c22c4ee84f)

Mahmoud-Emad:
@ehab-hassan 
Should be fixed, please check

ehab-hassan:
Work Completed

khaledyoussef24:
verified on devnet
Version c2d6060

the navbar disappears while zooming in
the hamburger icon appears to open the navbar
<img width="1680" alt="Image" src="https://github.com/user-attachments/assets/bc1f5f73-cf2e-49ee-99de-8bfdef76323d" />
<img width="389" alt="Image" src="https://github.com/user-attachments/assets/f562524f-eee6-43c0-9ad0-8582b7cd7734" />
<img width="1680" alt="Image" src="https://github.com/user-attachments/assets/1e080820-c7be-4162-993a-27d6c1a4c379" />

[testcase](https://app.testlodge.com/a/26076/projects/40893/suites/234374?expand_section=422331#case_4431575)

Issue #1985, opened 2024-01-15
Your profile page isn't responsive


A-Harby:
### Description

Your profile page isn't responsive.

A clear description of the bug, including relevant identifiers like:

- **Twin ID**: 11223

### Steps to reproduce

1. Navigate to the Dashboard.
2. Connect your wallet.
3. Change page size.

### Logs/Alerts

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/979989ce-53e3-4862-b87b-fb506b2012ce)

A-Harby:
Same issue on Devnet, b43ea4c.
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/8a5277ff-847a-4ea6-973e-7dd3f2baf575)

khaledyoussef24:
verified on devnet
version : 7b43120




![Image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/b6956a3c-d0d7-4153-b33a-c428158a2234)


![Image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/7245f604-0332-4d31-91e2-bab0c7d2394a)



![Image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/99ea987d-d4c0-470a-8ff7-4101188a1bfb)
![Image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/45c192e0-8153-46f4-b1b2-e81e4c5a39d0)

A-Harby:
I tried it on Qanet 2.3.0-alpha12 and on devnet staging, and it is still not fixed on small screens.

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/c66e0064-74a4-44d4-b7af-3f4e994c61d1)

0oM4R:
@samaradel should we move this to blocked on #2342 ?

ehab-hassan:
Work completed 

should be fixed

khaledyoussef24:
verified on devnet
Version c2d6060

<img width="1680" alt="Image" src="https://github.com/user-attachments/assets/6da5836b-1046-4874-a551-ef09d682ea4a" />
<img width="1680" alt="Image" src="https://github.com/user-attachments/assets/11846260-5f79-4c5e-b8a9-6709184c659f" />

[...]

PR #3347, opened 2024-08-29
Fix responsive

ehab-hassan:
### Description

Fix the dashboard at all screen size 

### Changes
**Note : please when you review keep in mind to check the standard size width (1440, 1024, 768, 425, 375)px**

[nav.webm](https://github.com/user-attachments/assets/0d42636b-60bd-4826-b676-45472eb8e6d4)

fixed
TF Chain
image

Twin Details
image

Transfer TFT
image

Transfer TFT/Deposit
image

Transfer TFT/Withdraw
image

calculator
image

Home when login
image

Simulation
image

Home without login
image

### Related Issues

#2342
#1987
#1261
#1985
#2410

### Documentation PR

### Checklist

- [x] Tests included
- [x] Build pass
- [ ] Documentation

### DIFF
--- a/packages/playground/src/App.vue
+++ b/packages/playground/src/App.vue
@@ -264,6 +264,12 @@ const hasGrid = computed(() => !!gridStore.grid);
 const permanent = ref(window.innerWidth > 980);
 const openSidebar = ref(permanent.value);
 
+watch(permanent, value => {
+  if (value) {
+    toolbarExtended.value = false;
+  }
+});
+
 function setSidebarOnResize() {
   permanent.value =
     window.innerWidth >
--- a/packages/playground/src/App.vue
+++ b/packages/playground/src/App.vue
@@ -108,14 +108,14 @@
 
       <v-main :style="{ paddingTop: navbarConfig ? '140px' : '70px' }">
         <v-toolbar
-          :extended="permanent"
+          :extended="toolbarExtended"
           extension-height="auto"
           class="border position-fixed"
           :style="{ zIndex: 1100, top: 0, left: 0, right: 0 }"
           :elevation="permanent ? '1' : '15'"
         >
           <template #extension>
-            <v-card class="w-100 ma-0 pa-0" v-if="openProfileResponsiveDialog">
+            <v-card class="w-100 ma-0 pa-0" v-if="toolbarExtended">
               <div class="pa-1">
                 <div class="d-flex justify-center">
                   <div class="">
@@ -127,10 +127,7 @@
                 <v-divider class="mb-2" />
                 <div class="d-flex justify-center">
                   <div class="">
-                    <ProfileManager
-                      v-model="openProfile"
-                      @update:modelValue="(e: boolean) => openProfileResponsiveDialog = e"
-                    />
+                    <ProfileManager v-model="openProfile" @update:modelValue="(e: boolean) => toolbarExtended = e" />
                   </div>
                 </div>
               </div>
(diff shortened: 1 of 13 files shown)


samaradel:
create account button is not centered
image

titles and contents are not on the same vertical line 
image

I think the balance field should be at the top then the convert field no?
image

the footer section, being ```Take Action``` ```About Us``` etc. at the center made it look weird, could it be 2 by row?
image

also, the stats card has very little content to take a row, could it be 2 by row?
image

buttons are not coordinated on size ```375px```
image

the home page(not logged) content could be wider on small screens
image

packages/playground/src/App.vue
```diff
@@ -212,8 +252,8 @@ const $router = useRouter();
 const profileManager = useProfileManager();
 const gridStore = useGrid();
 const network = process.env.NETWORK || (window as any).env.NETWORK;
-
-const openProfile = ref(false);
+const toolbarExtended = ref(false);
+const openProfile = ref(true);
```
MohamedElmdary, line 262: @0oM4R openProfile was set to `false` in your fix right? so we still need it or it can be `true` now?

packages/playground/src/App.vue
```diff
@@ -107,22 +107,47 @@
...
+                  <div class="">
+                    <TfSwapPrice>
+                      <FundsCard v-if="hasActiveProfile" />
+                    </TfSwapPrice>
+                  </div>
+                </div>
+                <v-divider class="mb-2" />
+                <div class="d-flex justify-center">
+                  <div class="">
+                    <ProfileManager v-model="openProfile" @update:modelValue="(e: boolean) => toolbarExtended = e" />
```
MohamedElmdary: don't use both v-model and @update:modelValue

instead
```suggestion
                    <ProfileManager :model-value="openProfile" @update:modelValue="(e: boolean) => {
                    toolbarExtended = e;
                    openProfile = e;
                    }" />
```

packages/playground/src/dashboard/twin_view.vue
```diff
@@ -33,14 +33,14 @@
...
             <v-list class="custom-list" density="compact">
               <v-row class="row-style">
-                <v-col cols="1" sm="2" style="min-width: fit-content">
-                  <v-list-item> ID :</v-list-item>
+                <v-col cols="3" class="px-0">
+                  <v-list-item class="px-0"> ID :</v-list-item>
                 </v-col>
-                <v-col>
-                  <v-list-item>
+                <v-col cols="9" class="px-0">
```
amiraabouhadid, line 42: Good job, please make sure that items in the right column are aligned to the end of their respective fields for a consistent look

amiraabouhadid (changes requested):

samaradel (approved):

ehab-hassan:
@samaradel 
image
image
image


"I think the balance field should be at the top then the convert field no?"
we can use the same order on the desktop(big) screen

"the footer section...."
Let's leave it on a row as is because we have an odd number of sections it ll be 2 rows then 1 row will not be nice also it'll not fit at 375px or 320px

" the stats card has very little content to take a row, could it be 2 by row?"
it's good to me

it is wider now
image

packages/playground/src/dashboard/twin_view.vue
ehab-hassan: image

packages/playground/src/App.vue
0oM4R: it should be initialized with false yes

ehab-hassan: @MohamedElmdary @0oM4R should I change something here at my PR?

MohamedElmdary: @ehab-hassan @0oM4R let's merge it now as it is and later invest more on how to fix that issue in another branch

MohamedElmdary (approved):

amiraabouhadid (approved):

amiraabouhadid (approved):

Issue #3149, opened 2024-07-23
Support managing domain on kubernetes


AhmedHanafy725:
### Which package/s are you suggesting this feature for?

Dashboard

### Is your feature request related to a problem? Please describe

The user might have some solutions on the cluster that require an ingress with a domain.

### Describe the solution you'd like

It would be good if he could create domains pointing to the cluster's master.

Mahmoud-Emad:
## Issue update 

### Work done

#### Changes

- Optimized the `ManageGatewy` component to work on both types of machines normal `VM` and `K8S` machine
- Generate the name based on the master
- takes the master networks as the domain-supported networks
- Change the dialog title to be dynamic
- Working on refactoring the component
- Load the WireGuard network name dynamic based on the machine type

#### Screenshots/Video

 [Screencast from 08-18-2024 03:10:58 PM.webm](https://github.com/user-attachments/assets/89de2baf-c59c-4849-9241-1e612823da9a)

Mahmoud-Emad:
After some investgation we found that our K3S image binds on IPv4 and won't work on IPv6 e.g. Mycelium or Planetary networks, we need to search if there a way to bind on both IPs v4 or v6 without break the current behavior

Mahmoud-Emad:
Due to https://github.com/threefoldtech/tf-images/issues/291 we have 2 options now

1. - [x] Support only IPv4 and Wireguard networks
2. - [ ] Wait until the operations update the K3S image

Mahmoud-Emad:
During the implementation, we discovered and agreed on:

- -[x] Set the default port to 443
- -[x] Define a new selector that lists all node's master/workers
- -[x] Let the user select the target node and load its interfaces based on the selected node
- -[x] Remove any V6 from the networks, let the user only choose Public IPv4 or Wireguard 

We'll go with the first option

> Due to [threefoldtech/tf-images#291](https://github.com/threefoldtech/tf-images/issues/291) we have 2 options now
> 
> 1. * [x]  Support only IPv4 and Wireguard networks
> 2. * [ ]  Wait until the operations update the K3S image

Mahmoud-Emad:
During the implementation i faced this bug https://github.com/threefoldtech/tfgrid-sdk-ts/issues/3321

A-Harby:
Can we change one of the logo of the manage workers or domain? Both having the same logo is kind of confusing.
[...]

PR #3368, opened 2024-09-04
Fix loading the Kubernetes attached machine with TLS

Mahmoud-Emad:
### Description

An issue was identified where the attached machine name was not displaying correctly when the machine had the `TLS` option enabled. This occurred because, in such cases, the backend was set to the machine's `IP:PORT` rather than a `URL`. To resolve this, I implemented a checker that determines whether the backend is a `URL` or an `IP`. Based on this check, the appropriate machine name is retrieved.

### Changes

- Check if the backends are URL or IP and get the attached machine based on it.

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/3149

#### Comment
- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/3149#issuecomment-2328560795

### Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation

### DIFF
--- a/packages/playground/src/components/manage_gateway_dialog.vue
+++ b/packages/playground/src/components/manage_gateway_dialog.vue
@@ -296,13 +296,7 @@ export default {
     };
 
     const getDomainNode = (domain: GridGateway): string => {
-      const extractIP = (input: string) => {
-        try {
-          return new URL(input).hostname;
-        } catch {
-          return input.split(":")[0];
-        }
-      };
+      const extractIP = (input: string) => input.replace("https://", "").replace("http://", "").split(":")[0];
 
       const IP = extractIP(domain.backends[0]);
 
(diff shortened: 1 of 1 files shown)


packages/playground/src/components/manage_gateway_dialog.vue
```diff
@@ -296,8 +296,15 @@ export default {
...
-      const IP = getIP(domain.backends[0]);
+      const extractIP = (input: string) => {
+        try {
+          return new URL(input).hostname;
+        } catch {
+          return input.split(":")[0];
+        }
+      };
+
+      const IP = extractIP(domain.backends[0]);
```
MohamedElmdary: can we stick into 1 solution either prefix url with `http://` so URL work with ip or use split with both cases so we don't get confuse later

MohamedElmdary (approved):

samaradel (approved):

Issue #3364, opened 2024-09-03
🐞[QAnet] [Bug]: Grace period contracts are showing zero locked tfts with no billing rate because its on a dedicated machine while it was never on a dedicated node


A-Harby:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### which package/s did you face the problem with?

Dashboard

### What happened?

Grace period contracts are showing `This contract is in a grace period because it's on a dedicated machine also in a grace period. That's why this node is locked with zero TFTS and no billing rate.` while I never rented a node or deployed any of my contracts on a dedicated machine!

### which network/s did you face the problem on?

Dev

### version

4829e02

### Twin ID/s

8823

### Node ID/s

1, 4416

### Relevant screenshots/screen records

![image](https://github.com/user-attachments/assets/aa144849-eb99-466d-b730-d04a1e8b7ba9)
![image](https://github.com/user-attachments/assets/6f1b48a3-4e57-44e7-8097-634b363ea53f)

[node_contracts.json](https://github.com/user-attachments/files/16850799/node_contracts.json)
[name_contracts.json](https://github.com/user-attachments/files/16850801/name_contracts.json)

### Relevant log output

```console
This contract is in a grace period because it's on a dedicated machine also in a grace period. That's why this node is locked with zero TFTS and no billing rate.
```

amiraabouhadid:
related to tfchain locked amount calculation being broken #3340 .

A-Harby:
Verified, Qanet 2.6.0-rc3.

The message now is more clearer and correct.
![image](https://github.com/user-attachments/assets/2525e454-28c2-4db0-bce0-fbf6764bf060)
![image](https://github.com/user-attachments/assets/fb144e55-4ee9-4082-bde3-f70206ca1c7b)

[TC2618 - Grace period locked TFT amount](https://app.testlodge.com/a/26076/projects/40893/suites/234374?expand_section=430129#case_4214522)

PR #3387, opened 2024-09-07
chain release2.9 compatibility  Grace Period and overdue

0oM4R:
### Description

We have new accurate way to resume/unlock contracts on tfchain2.9, this pull request is for replace the old way as it is no longer supported.

#### General idea:
Now the amount required to resume the contract is described as **Overdue**.
The overdue amount basically is the sum of three parts:
  1. Total over draft: is the sum of additional overdraft and standard overdraft.
  2. Unbilled NU: is the unbilled amount of network usage.
  3. The estimated cost of the contract for the total period: this part is dependent on the contract type and if the contract is on a rented node or not.
> Note:
> - The total period is the time since the last billing added to the allowance period.
> - If the contract is a rent contract, will add both of ipv4 cost and the total overdue of all associated contracts. 

### Changes
**Please node**: you may notice that we have multiple functions that generally do the same thing but require different arguments. 
for example: we need some information that is already in the contract type, 
so instead of requesting the contract in side the function we are passing it.
but in some cases we only have the contract id? we provide a function that take the contract id and then request the contract then pass it to the inner function.
 
#### TFChain client
1. Fixed the [generateTypes.bash](https://github.com/threefoldtech/tfgrid-sdk-ts/pull/3387/files#diff-b4638b04c8f8163527467a64e7acb4b40072b9697f55b635c1672fbd2015c4f7):
  - the old script required to have ts-node installed, now we can run those commands using yarn.
2. In contracts, we add some functions that help in grid client new features.
#### Grid client: 
  - add deprecated annotation to the no longer supported functions that are associated with locked amount.
  - Calculator module:  added a function to calculate the name pricing.
  - Contract module: 
     - added some functions that calculate overdue, you can pass the contract, or the contract id.
     - add functions to unlock contracts, all contracts, and batch unlock using array of contracts or array of contractss ids
    - we are using gridproxy in unlock contracts
 
#### Dashboard

support new functions in grid client, refactor, and remove the deprecated gird client functions.
we have some changes in the flow:
 - If the user select node contracts, and some of them are on a rent contract, we will show the rent contract id with the selected contracts
     Screenshot from 2024-09-09 15-18-12

 - if use click on grace period chip of node contract with ipv4 on rented node the content of the dialog depends on the rent contract state:
    - if the rent contract  state is `Created` : will get the node contract overdue, and user will be able to unlock the contract.
        
    - If on `Grace period`: will show a message that you will have to check the rent contract and unlock it.
        Screenshot from 2024-09-10 15-06-56

   
fix loading spinner in loading locked amount details: 
 Screenshot from 2024-09-09 14-55-38

### Related Issues

#3340 
#3364 

### Tested Scenarios
- show contract/s cost : this will be nearly from pricing calculator if the contract got billed right before moving to the grace period as in this case overdue will be small for example, name contract cost should be about `0.025` tfts
- select node contracts on rented node that its contract on grace period,
- unlock all contracts associated with rent contract: in this case to verify that the node contract got billed, use `smarContractModule/contractPaymentState` call, the additional and standred overdraft should be 0
- Unlock single or multiple contracts
- in any case we should'n have a contract got billed twice: you can add log line to log the ids in batchUnlock contracts
Tips: 
- To send contract to a grace period, you need to follow some steps 
   1- empty you walet, this can be done by polkadot UI `Balances/TransferAll`
   2- get the contract Id, must be billable one like rent contract or node contract on shared node, name contract, and call `smarContractsModule/billcontractForBlock` using any other funded account.
   
- to send node contract with public ip on rented node to grace period and keep it's rent contract is on created state
   - fund you wallet, and call bill contract for block for the rent contract,
   - then send the node contract with ipv4 only to grace period as mentioned
 
### Documentation PR

### Checklist

- [ ] Tests included 
- [x] Build pass
- [ ] Documentation


packages/grid_client/src/clients/tf-grid/contracts.ts
```diff
@@ -301,13 +331,216 @@ class TFContracts extends Contracts {
...
+   * @returns  {Decimal} The amount in TFT.
+   */
+  private async convertToTFT(USD: Decimal) {
+    try {
+      const tftPrice = (await this.client.tftPrice.get()) ?? 0;
+      return USD.div(tftPrice);
+    } catch (error) {
+      throw new GridClientError(`Failed to convert to mTFT due: ${error}`);
+    }
+  }
```
AhmedHanafy725: please use the currency module instead

packages/grid_client/src/clients/tf-grid/contracts.ts
```diff
@@ -301,13 +331,216 @@ class TFContracts extends Contracts {
...
+
+  /**
+   * Calculate total IPV4 and total overdraft on all node contracts on rented node.
+   * @description In node contracts on rented nodes we need to have both the overdraft to be added to rent contract overdraft.
+   * please note that the unbilled NU amount is added to the total overdraft exactly as the in the other contract types.
+   * the IPV4 cost to add to the estimated cost of the rent contract.
+   */
+  private async getContractsCostOnRentedNode(nodeId: number, proxy: GridProxyClient) {
+    const contracts = await proxy.contracts.list({ nodeId, state: [ContractState.GracePeriod], numberOfPublicIps: 1 });
```
AhmedHanafy725: I think this only returns the first page(50 items), what if I have more than that?

packages/grid_client/src/clients/tf-grid/contracts.ts
```diff
@@ -301,13 +331,216 @@ class TFContracts extends Contracts {
...
+      hru: hruInGB,
+      certified,
+      ipv4u: !!ipv4,
+    });
+
+    return {
+      dedicatedPrice: res.dedicatedPrice,
+      sharedPrice: res.sharedPrice,
+    };
+  }
```
AhmedHanafy725: Why do you add a calculate helper to the contracts module?

packages/grid_client/src/clients/tf-grid/contracts.ts
```diff
@@ -335,28 +568,60 @@ class TFContracts extends Contracts {
...
+   * This function lists all grace period contracts, then call {@link batchUnlockContracts}.
+   * @param {String} gridProxyUrl
+   * @returns contract ids that have been requested to resume.
+   */
+  async unlockMyContracts(gridProxyUrl: string) {
+    const proxy = new GridProxyClient(gridProxyUrl);
+
+    const contracts = await proxy.contracts.list({
+      state: [ContractState.GracePeriod],
+      twinId: await this.client.twins.getMyTwinId(),
```
AhmedHanafy725: this will be only one page

packages/grid_client/src/clients/tf-grid/contracts.ts
0oM4R: this is a private function to reduce redundant code and make the conversion from bytes to GB inside it instead of doing it multiple times, should be moved to the calculator module ?

AhmedHanafy725: yes, it would be a better place

packages/grid_client/src/clients/tf-grid/contracts.ts
```diff
@@ -301,13 +326,216 @@ class TFContracts extends Contracts {
...
+    const totalOverDraft = overDraftResult.reduce(
+      (acc: Decimal, paymentState) => acc.add(paymentState.additionalOverdraft).add(paymentState.standardOverdraft),
+      new Decimal(0),
+    );
+    const totalNUCost = billingInfoResult.reduce((acc: number, billingInfo) => acc + billingInfo.amountUnbilled, 0);
+    const totalIPCost = (ipPrice * (contracts.data.length || 0)) / 1e7;
+    return {
+      //return ip cost per month
+      totalIpCost: totalIPCost * 24 * 30,
+      totalOverdraft: totalOverDraft.add(totalNUCost),
```
sameh-farouk: You can't add them directly because `totalNUCost` is in unit-USD and `totalOverDraft` is in unit-TFT.

packages/grid_client/src/clients/tf-grid/contracts.ts
```diff
@@ -301,13 +326,216 @@ class TFContracts extends Contracts {
...
+
+    const overDraftResult = await Promise.all(overDraftPromises);
+    const billingInfoResult = await Promise.all(BillingInformationPromises);
+
+    const totalOverDraft = overDraftResult.reduce(
+      (acc: Decimal, paymentState) => acc.add(paymentState.additionalOverdraft).add(paymentState.standardOverdraft),
+      new Decimal(0),
+    );
+    const totalNUCost = billingInfoResult.reduce((acc: number, billingInfo) => acc + billingInfo.amountUnbilled, 0);
+    const totalIPCost = (ipPrice * (contracts.data.length || 0)) / 1e7;
```
sameh-farouk: `ipPrice` was previously converted from unit-USD to USD. Why divide it by 1e7 again here?

packages/grid_client/src/clients/tf-grid/contracts.ts
```diff
@@ -301,13 +326,216 @@ class TFContracts extends Contracts {
...
+      totalOverdraft: totalOverDraft.add(totalNUCost),
+    };
+  }
+
+  /**
+   * This function is for calculating the estimated cost of the contract per month.
+   * @description
+   *  Name contract cost is fixed price for the unique name,
+   * Rent contract cost is the cost of the total node resources,
+   * Node contract have to cases:
```
sameh-farouk: There is a small typo here to -> two

packages/grid_client/src/clients/tf-grid/contracts.ts
sameh-farouk: By multiplying by `contracts.data.length`, are we assuming that a node contract can only have 1 IP address?

packages/grid_client/src/clients/tf-grid/contracts.ts
```diff
@@ -301,13 +326,216 @@ class TFContracts extends Contracts {
...
+    const calc = new calculator(this.client);
+
+    if (contract.type == ContractType.Name) return await calc.namePricing();
+
+    //TODO allow ipv4 to be number
+
+    /** Other contract types need the node information */
+    const nodeDetails = await proxy.nodes.byId(contract.details.nodeId);
+
+    const certified = nodeDetails.certificationType == CertificationType.Certified ? true : false;
```
sameh-farouk: Consider using the name `isCertified` instead.

packages/grid_client/src/clients/tf-grid/contracts.ts
```diff
@@ -301,13 +326,216 @@ class TFContracts extends Contracts {
...
+    } else return contracts.data;
+  }
+
+  /**
+   * Calculate total IPV4 and total overdraft on all node contracts on rented node.
+   * @description In node contracts on rented nodes we need to have both the overdraft to be added to rent contract overdraft.
+   * please note that the unbilled NU amount is added to the total overdraft exactly as the in the other contract types.
+   * the IPV4 cost to add to the estimated cost of the rent contract.
+   */
+  private async getContractsCostOnRentedNode(nodeId: number, proxy: GridProxyClient) {
```
sameh-farouk: This function calculates two separate costs:
1. The overdue amount for all node contracts associated with a rent contract (total overdraft + unbilled NU cost).
2. The IP cost for all node contracts per month.

In my opinion, it would be better to handle these calculations with separate functions.

sameh-farouk (commented):
I submitted a partial review today and will continue tomorrow.

packages/grid_client/src/clients/tf-grid/contracts.ts
0oM4R: the current behavior, yes, each contract should have 1 IP

packages/grid_client/src/clients/tf-grid/contracts.ts
```diff
@@ -301,13 +326,216 @@ class TFContracts extends Contracts {
...
+    const contracts = await this.listAllContracts(proxy, {
+      nodeId,
+      state: [ContractState.GracePeriod],
+      numberOfPublicIps: 1,
+    });
+
```
sameh-farouk, line 398: Can we handle the case where there are no returned contracts and return early?

packages/grid_client/src/clients/tf-grid/contracts.ts
```diff
@@ -301,13 +326,216 @@ class TFContracts extends Contracts {
...
+    /** Node contract on rented node
+     * If the node contract has IPV4 will return the price of the ipv4 per month
+     * If not there is no cost, will return zero
+     */
+
+    if (nodeDetails.rented) {
+      if (!contract.details.number_of_public_ips) return 0;
+
+      const ipPrice = (await this.client.pricingPolicies.get({ id: 1 })).ipu.value / 10 ** 7;
+      return ipPrice * 30 * 24;
```
sameh-farouk: Why not use the number of public IPs in the contract instead?
FYI, from the tfchain side, I can create a contract that reserves more than one public IP. Since it is a valid case, despite some client limitations or restrictions, it should be handled when calculating the contract costs as it may be created with different clients.

packages/grid_client/src/clients/tf-grid/contracts.ts
```diff
@@ -301,13 +326,216 @@ class TFContracts extends Contracts {
...
+    const calc = new calculator(this.client);
+
+    if (contract.type == ContractType.Name) return await calc.namePricing();
+
+    //TODO allow ipv4 to be number
+
+    /** Other contract types need the node information */
+    const nodeDetails = await proxy.nodes.byId(contract.details.nodeId);
+
+    const certified = nodeDetails.certificationType == CertificationType.Certified ? true : false;
```
sameh-farouk: Do you need the ternary operator here?
The condition `nodeDetails.certificationType == CertificationType.Certified` already evaluates to true or false.

Also, it is better to use === to avoid unexpected behavior caused by implicit type conversion.

packages/grid_client/src/clients/tf-grid/contracts.ts
```diff
@@ -301,13 +326,216 @@ class TFContracts extends Contracts {
...
+    // time since the last billing with allowance time of **one hour**
+    const totalPeriodTime = elapsedSeconds + SECONDS_ONE_HOUR;
+
+    //USD
+    const contractCostPerSecond = contractMonthlyCost.div(30 * 24 * SECONDS_ONE_HOUR);
+    const contractCostTFT = await this.convertToTFT(contractCostPerSecond);
+
+    // cost of the current billing period with the mentioned allowance time
+    const totalPeriodCost = contractCostTFT.times(totalPeriodTime);
+    const overdue = totalOverDraft.add(unbilledNU);
```
sameh-farouk: As previously explained in my other comment, You can't add them directly.

packages/grid_client/src/clients/tf-grid/contracts.ts
```diff
@@ -301,13 +326,216 @@ class TFContracts extends Contracts {
...
+    const contractInfo = options.contractInfo;
+
+    /** Get the Un-billed amount for the network usage */
+    const unbilledNU = (await this.client.contracts.getContractBillingInformationByID(contractInfo.contract_id))
+      .amountUnbilled;
+
+    const { standardOverdraft, additionalOverdraft, lastUpdatedSeconds } =
+      await this.client.contracts.getContractPaymentState(contractInfo.contract_id);
+
+    /**Calculate the elapsed seconds since last pilling*/
```
sameh-farouk: There is a small typo here "pilling" -> "billing"

packages/grid_client/src/clients/tf-grid/contracts.ts
```diff
@@ -301,13 +326,216 @@ class TFContracts extends Contracts {
...
+    //USD
+    const contractCostPerSecond = contractMonthlyCost.div(30 * 24 * SECONDS_ONE_HOUR);
+    const contractCostTFT = await this.convertToTFT(contractCostPerSecond);
+
+    // cost of the current billing period with the mentioned allowance time
+    const totalPeriodCost = contractCostTFT.times(totalPeriodTime);
+    const overdue = totalOverDraft.add(unbilledNU);
+
+    //convert to TFT
+    const OverdueTFT = overdue.div(10 ** 7);
```
sameh-farouk: `OverdueTFT` name should follow camelCase convention.

packages/grid_client/src/clients/tf-grid/contracts.ts
0oM4R: Actually, I put both of them in one place to reduce listing the contracts again, here already have the contracts and their count, i was thinking about move the ip calculations to separate function and call it here but this will change nothing imo, what do you think?

packages/grid_client/src/clients/tf-grid/contracts.ts
```diff
@@ -301,13 +326,216 @@ class TFContracts extends Contracts {
...
+
+    /** Node contract on rented node
+     * If the node contract has IPV4 will return the price of the ipv4 per month
+     * If not there is no cost, will return zero
+     */
+
+    if (nodeDetails.rented) {
+      if (!contract.details.number_of_public_ips) return 0;
+
+      const ipPrice = (await this.client.pricingPolicies.get({ id: 1 })).ipu.value / 10 ** 7;
```
sameh-farouk: I would avoid using magic numbers like 10 ** 7 and define it as constant `TFT_CONVERSION_FACTOR` (maybe in the client) then use it everywhere else for clarity and maintainability.

packages/grid_client/src/clients/tf-grid/contracts.ts
0oM4R: now I get the ip counts not the contracts

packages/grid_client/src/clients/tf-grid/contracts.ts
sameh-farouk: This is because this function does a lot of things.
What about splitting it into three separate functions?
The three new functions would be:
1. `getNodeContractsOnRentedNode`
2. `calculateNodeContractsOverdue`
3. `calculateIPCostPerMonth`
This way, you can retrieve the contracts from the first function and then pass them to the second and third functions.

packages/grid_client/src/clients/tf-grid/contracts.ts
sameh-farouk: Another point to consider regarding unbilled NU is that it may be eligible for a discount or subject to an added premium:
* A discount may apply based on the user's balance.
* A premium is added when a certified node is used.

This has already been handled for overdraft stored on the contract payment state, as well as the cost of the contract retrieved from the calc module.

packages/grid_client/src/clients/tf-grid/contracts.ts
0oM4R: add for the current file, will open an issue to make it global value for the grid client

sameh-farouk:
@0oM4R I attempted a quick test, but it didn't go well.

Screenshot_20240930_160919

Screenshot_20240930_160829

sameh-farouk:
Another point, it seems you haven't incorporated the changes discussed in the tg group yet.

> Unlock All/ Unlock -> Resume All / Resume
Locked Balance -> On hold
Show in UI: Available and On hold

0oM4R:
> Another point, it seems you haven't incorporated the changes discussed in the tg group yet.
> 
> > Unlock All/ Unlock -> Resume All / Resume
> > Locked Balance -> On hold
> > Show in UI: Available and On hold

there is a separate issue for it https://github.com/threefoldtech/tfgrid-sdk-ts/issues/3400

0oM4R:
Screenshot from 2024-09-30 17-59-09
Screenshot from 2024-09-30 17-59-42

works fine with me but we should handle the zero tft case in the ui


> @0oM4R I attempted a quick test, but it didn't go well.
> 
> Screenshot_20240930_160919
> 
> Screenshot_20240930_160829

sameh-farouk:
Sorry, My bad.
I didn't pull the latest changes. So I will give it another try tomorrow.

sameh-farouk:
> works fine with me but we should handle the zero tft case in the ui

Sorry, My bad.
I didn't pull the latest changes. So I will give it another try tomorrow.

sameh-farouk:
Calculations looks much better now.

sameh-farouk (dismissed):

0oM4R:
Test report: 
will update this comment with the tested scenarios in details
 
#### Rent contract without any node contracts on it
image
image

### the allowance window in the next scenarios is __60__ seconds

PR #3395, opened 2024-09-10
fix: balance calculations to be compatible with new chain changes  

0oM4R:
### Description
To get the total balance now we have to add the free and the reserved, locked balance is not used anymore

### Changes

- replaced locked with reserved
- show the sum of reserved and free in profile manager
- make `freeBalance` variable computed to make it always synced with the balanc

### Related Issues

- #3340 

### Tested Scenarios
- having zero tft, and check the profile manager

- use polkadot UI and call `system/account` to retrieve the actual balance and compare it with the shown in profile manager
- try to empty you account from balance at all and check for the shown values 
- also try to deploy when you have balance less than 2 TFT < should fail> 
- try to deploy when the **free** balance is larger than 2 TFT 

Screenshot from 2024-09-10 11-47-17

### Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation

### DIFF
--- a/packages/playground/src/components/contracts_list/contracts_table.vue
+++ b/packages/playground/src/components/contracts_list/contracts_table.vue
@@ -384,7 +384,7 @@ const selectedContracts = ref<NormalizedContract[]>([]);
 const selectedItem = ref();
 const profileManagerController = useProfileManagerController();
 const balance = profileManagerController.balance;
-const freeBalance = computed(() => (balance.value?.free ?? 0) - (balance.value?.locked ?? 0));
+const freeBalance = computed(() => balance.value?.free ?? 0);
 const unlockContractLoading = ref(false);
 const unlockDialog = ref(false);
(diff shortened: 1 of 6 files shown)


zaelgohary (approved):

amiraabouhadid (commented):
works when balance is exactly 2tft
image

0oM4R:
> works when balance is exactly 2tft image

this is the expected behavior right ?

amiraabouhadid (changes requested):
however, rounding the numbers (1.998 is rounded to 2) is confusing in the err msg . I suggest you use the same number of decimals in err msg as used in balance display
image

0oM4R:
image

amiraabouhadid (approved):
tested with less than 2TFT, works
Screenshot from 2024-09-22 12-23-55tested with exactly 2TFT, works
image

Issue #3343, opened 2024-08-28
🐞 [Bug]: Change mail config validator on taiga


A-Harby:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### which package/s did you face the problem with?

Dashboard

### What happened?

The same [issue](https://github.com/threefoldtech/tfgrid-sdk-ts/issues/2152) is happing to the [taiga](https://github.com/threefoldtech/tfgrid-sdk-ts/issues/2152#issuecomment-2314834580), and also check all of the other application stmp.

### which network/s did you face the problem on?

Dev

### version

448a213

### Relevant screenshots/screen records

https://github.com/threefoldtech/tfgrid-sdk-ts/issues/2152#issuecomment-2314834580

### Relevant log output

```console
.
```

amiraabouhadid:
Use validation used in discourse solution. Remove the condition applying validation to just discourse.  
Check after the validation change that the solution won't crash because other solution might inherently have different email requirements. 
Check if you can still send emails with SMTP

samaradel:
### Work in Progress (WIP):
utils created to validate SMTP (email, apiKey, username), faced an error .. investigating..

samaradel:
### Work Completed:
- Remove ```isDiscourse``` check and apply adding email or username for all apps with SMTP.
- Create a function to validate applications with SMTP.
- All solutions with SMTP are tested except ```Taiga``` which is blocked on https://github.com/threefoldtech/tf-images/issues/293

samaradel:
### Work Completed:
- Doc string added to explain the new ```IsAlphaExpectDashAndUnderscore``` validator function.
- Unit test added too.

khaledyoussef24:
verified on devnet
Version a78daae

now user names are allowed so user can use api keys or user name normally
![Image](https://github.com/user-attachments/assets/ef31e170-5439-4bda-84f4-5e56162ca3db)
![Image](https://github.com/user-attachments/assets/36435ba0-69b9-4d89-a1f0-e2ad6a7618b6)

passwords is allowed more than 50 characters
validation is also added (in this case i used space in the user name to test validation ):
![Image](https://github.com/user-attachments/assets/597aca31-ead5-43ac-8e06-e71f46480373)
here to check that passwords can be more than 50 characters :
![Image](https://github.com/user-attachments/assets/0af14ced-0aea-4cb0-a465-11dac318027f)
here to check the validation of passwords when inserting space :
[...]

PR #3398, opened 2024-09-10
Allow using `username` and `email` in SMTP configuration

samaradel:
### Description

- Remove ```isDiscourse``` check and apply adding email or username for all apps with SMTP.
- Add the minimum and maximum lengths for the username and email.
- validate the email input to accept the username too.
- Accept usernames containing dashes and underscores they must start with an alphanumeric character and validate them.

image

- limit the password length to 69 chars as per ```sendgrid``` documentation 
image

### Changes

image
### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/3343

### Tested Scenarios

- Check if the user used an email or username in the Email/username input and validate it.
- Check if the input accepts less than 2 characters.
- Check if the input accepts more than 50 characters.
- Check if the username or email can start with anything except an alphanumeric character.
- Check if we could leave the input empty 

image

### Documentation PR

- https://github.com/threefoldtech/info_grid/issues/645

### Checklist

- [x] Tests included
- [x] Build pass
- [x] Documentation


packages/playground/src/components/smtp_server.vue
```diff
@@ -46,7 +45,7 @@
           :rules="[
             validators.required('Password is required.'),
             validators.minLength('Password must be at least 6 characters.', 6),
-            validators.maxLength('Password cannot exceed 50 characters.', 50),
+            validators.maxLength('Password cannot exceed 50 characters.', 69),
```
zaelgohary:
```suggestion
            validators.maxLength('Password cannot exceed 69 characters.', 69),
```

zaelgohary: Also, where is this validation coming from?

packages/playground/src/weblets/tf_discourse.vue
```diff
@@ -94,7 +94,7 @@
       </template>
 
       <template #mail>
-        <SmtpServer v-model="smtp" :persistent="true" :tls="true" :is-discourse="true">
+        <SmtpServer v-model="smtp" :persistent="true" tls>
```
zaelgohary:
```suggestion
        <SmtpServer v-model="smtp" persistent tls>
```

packages/playground/src/utils/validators.ts
```diff
@@ -777,3 +777,9 @@ export async function isAvailableName(grid: GridClient, name: string) {
     return { message: "Name is already taken." };
   }
 }
+
+export function validateSmtp(v: string) {
+  const emailValidation = isEmail("Please provide a valid email address.")(v);
+  const username = isAlpha("Name must start with an alphabetical character.")(v[0]);
+  return emailValidation && username;
```
zaelgohary: What if the user entered a username, not an email? This validation would expect the user to always to enter an email.

samaradel: No, it will expect both of them you can try it

packages/playground/src/components/smtp_server.vue
```diff
@@ -46,7 +45,7 @@
           :rules="[
             validators.required('Password is required.'),
             validators.minLength('Password must be at least 6 characters.', 6),
-            validators.maxLength('Password cannot exceed 50 characters.', 50),
+            validators.maxLength('Password cannot exceed 69 characters.', 69),
```
0oM4R, line 55: can you explain why 69 chars?

packages/playground/src/utils/validators.ts
```diff
@@ -777,3 +777,11 @@ export async function isAvailableName(grid: GridClient, name: string) {
     return { message: "Name is already taken." };
   }
 }
+
+export function validateSmtp(v: string) {
+  const emailValidation = isEmail("Please provide a valid email address.")(v);
+  const username = /[!@#$%^&*()_+\-={}:<>?,./]/.test(v);
+  if (username && emailValidation) {
+    return { message: "Please provide a valid input" };
```
0oM4R:
```suggestion
    return { message: "Please provide a valid username or email" };
```

packages/playground/src/utils/validators.ts
```diff
@@ -777,3 +777,11 @@ export async function isAvailableName(grid: GridClient, name: string) {
     return { message: "Name is already taken." };
   }
 }
+
+export function validateSmtp(v: string) {
```
0oM4R: we may name it `IsValidSmtp` or something similar, also, `v` to be something like `input` to increase readability

0oM4R:
please mention the idea behind the validation rules, 69 characters, and the regex of the username

packages/playground/src/utils/validators.ts
```diff
@@ -777,3 +777,11 @@ export async function isAvailableName(grid: GridClient, name: string) {
...
+  const emailValidation = isEmail("Please provide a valid email address.")(v);
+  const username = /[!@#$%^&*()_+\-={}:<>?,./]/.test(v);
```
maayarosama: why are you using regex here?

packages/playground/src/components/smtp_server.vue
samaradel: Because we use ```sendgrid``` and its password is about 69 chars.

packages/playground/src/utils/validators.ts
samaradel: I'm fine about naming but for ```v``` it stands for ```value``` whatever it looks like, like the rest of the validators

packages/playground/src/utils/validators.ts
samaradel: to avoid using symbols when adding username.

0oM4R:
I can't test the solutions, #3449 and discourse not responding for more than one hour

samaradel:
@0oM4R you should test the field validation to see if it accepts email or username in the discourse without problems

packages/playground/src/utils/validators.ts
0oM4R: It's more readable and best practice is to have arguments names that are clear and descriptive. 
and please add docstring to the new added functions

packages/playground/src/utils/validators.ts
0oM4R: What are the requirements of the username?

samaradel: no requirement just string, just added regex to avoid symbols so we can validate emails too in the same input

packages/playground/src/utils/validators.ts
0oM4R, line 1: Please add unit tests

0oM4R (changes requested):

0oM4R (commented):
i think we need to include those changes in manual Please open an issue for it in https://github.com/threefoldtech/info_grid

packages/playground/src/components/smtp_server.vue
```diff
@@ -20,17 +20,16 @@
...
         :value="$props.modelValue.username"
         :rules="[
-          validators.required('Email is required.'),
-          validators.isEmail('Please provide a valid email address.'),
-          v => {
-            return isDiscourse ? undefined : validators.isEmail('Please provide a valid email address.')(v);
-          },
+          validators.required('Email or Username is required.'),
+          (v: string) => {
+            return validators.isValidSmtp(v);
```
0oM4R: as we accepts a username now i think we should add min char validation to be at least two chars

samaradel: @AhmedHanafy725 agree?

samaradel:
> i think we need to include those changes in manual Please open an issue for it in https://github.com/threefoldtech/info_grid

I see all mail server apps in the manual don't mention anything about validation, it is just about filling fields, I don't think it's necessary

0oM4R:
> > i think we need to include those changes in manual Please open an issue for it in https://github.com/threefoldtech/info_grid
> 
> I see all mail server apps in the manual don't mention anything about validation, it is just about filling fields, I don't think it's necessary

what i mean is to mention that we accepts username not only email, 
image

samaradel:
@0oM4R you mean the ```Admin Email``` point, ok I will create one .. thanks :)

packages/playground/src/components/smtp_server.vue
0oM4R: I discussed it with him earlier today, you can go a head :D

packages/playground/src/utils/validators.ts
```diff
@@ -777,3 +777,20 @@ export async function isAvailableName(grid: GridClient, name: string) {
...
+/**
+ * Validates an SMTP input string.
+ *
+ * Checks if the input string is a valid email address and does not contain special characters.
+ *
+ * @param {string} input - The input string to validate.
+ * @returns {{ message: string }} - An object with an error message if the input is invalid, or undefined if the input is valid.
+ */
+
+export function isValidSmtp(input: string) {
```
xmonader: 1- where is the return type of the function?
2-if the function is named `isX` is expect it to return a `boolean`.  True if X and false if not. To return null that may denote a failure e.g failure to allocate memory, or no data available in such context.

So we have multiple options

1- return a boolean (and discard the error messages, keep it constant to invalid smtp information) 
2- return a boolean with a context, e.g `boolean, null` in case of success and `boolean and error string` in case of failure `  so [boolean, string|null]
3- return a boolean (success) or an error message (failure) `boolean | string`
4- return a new type e.g Result that encapsulates the success/failure status, and message/error fields

```
type Result<T> = {
  success: true;
  value: T;
} | {
  success: false;
  error: string;
};
```

packages/playground/tests/utils/isValidSmtp.test.ts
```diff
@@ -0,0 +1,24 @@
+import { describe, expect, it } from "vitest";
+
+import { isValidSmtp } from "../../src/utils/validators";
+
+describe("isValidSmtp", () => {
```
xmonader: Probably need more cases e.g empty string, multiple invalid forms of email as well, usage of special characters

packages/playground/src/utils/validators.ts
```diff
@@ -777,3 +777,20 @@ export async function isAvailableName(grid: GridClient, name: string) {
     return { message: "Name is already taken." };
   }
 }
+
+/**
+ * Validates an SMTP input string.
+ *
+ * Checks if the input string is a valid email address and does not contain special characters.
```
xmonader: Is it only email? if so change the input parameter to `email`

packages/playground/src/utils/validators.ts
```diff
@@ -777,3 +777,20 @@ export async function isAvailableName(grid: GridClient, name: string) {
...
+ *
+ * @param {string} input - The input string to validate.
+ * @returns {{ message: string }} - An object with an error message if the input is invalid, or undefined if the input is valid.
+ */
+
+export function isValidSmtp(input: string) {
+  const emailValidation = isEmail("Please provide a valid email address.")(input);
+  const username = /[!@#$%^&*()\s_+\-={}:<>?,./]/.test(input);
+  if (username && emailValidation) {
```
xmonader: I don't understand this condition, does that mean if it's email and a username (doesn't have special characters) then it's true? 

shouldn't it be like that

```
const isValidEmail = isEmail("Please provide a valid email address.")(input);
const hasSpecialCharacters = /[!@#$%^&*()\s_+\-={}:<>?,./]/.test(input);

if (hasSpecialCharacters && isValidEmail) {
      return { message: "Please provide a valid username or email" };
}
```

Which I still don't understand.

packages/playground/src/utils/validators.ts
```diff
@@ -777,3 +777,20 @@ export async function isAvailableName(grid: GridClient, name: string) {
...
+ *
+ * Checks if the input string is a valid email address and does not contain special characters.
+ *
+ * @param {string} input - The input string to validate.
+ * @returns {{ message: string }} - An object with an error message if the input is invalid, or undefined if the input is valid.
+ */
+
+export function isValidSmtp(input: string) {
+  const emailValidation = isEmail("Please provide a valid email address.")(input);
+  const username = /[!@#$%^&*()\s_+\-={}:<>?,./]/.test(input);
```
xmonader: This could have been simplified instead of `matching against special characters` you could have just made it `match against valid characters` which are `a-z[A-Z0-9]+`

packages/playground/tests/utils/isValidSmtp.test.ts
```diff
@@ -0,0 +1,24 @@
...
+  it("returns nothing for valid username and email", () => {
```
xmonader: I truly don't understand can you point out the `username` and the `email` because there's only just one input name `input` which looks like an email.

packages/playground/tests/utils/isValidSmtp.test.ts
```diff
@@ -0,0 +1,24 @@
...
+    const input = "valid.username@example.com";
+    expect(isValidSmtp(input)).toBeUndefined();
+  });
+
+  it("returns an error message for special characters in username", () => {
+    const input = "invalid_email";
```
xmonader: `_` is not the only special character

packages/playground/tests/utils/isValidSmtp.test.ts
```diff
@@ -0,0 +1,24 @@
...
+    expect(isValidSmtp(input)).toEqual({
+      message: "Please provide a valid username or email",
+    });
+  });
+
+  it("returns an error message for white spaces in username", () => {
+    const input = "invalid username";
```
xmonader: there're more types of spaces beyond `" "` there's is `\t` and `\v`

packages/playground/src/utils/validators.ts
xmonader: > This could have been simplified instead of `matching against special characters` you could have just made it `match against valid characters` which are `a-z[A-Z0-9]+`

Please note this is the right way, given there's a very long of list of unhandled utf8 input, that's why you should state the allowed ones as long as you can iterate them, instead of chasing a very long list of symbols.

packages/playground/src/utils/validators.ts
samaradel: done

packages/playground/src/components/smtp_server.vue
samaradel: done

packages/playground/src/utils/validators.ts
samaradel: good point. the function is deleted now and the validation is created in the input validation rules.

packages/playground/tests/utils/isValidSmtp.test.ts
samaradel: as the function is deleted, the test case is deleted too and you could check it in the current rules.

packages/playground/src/utils/validators.ts
samaradel: It now accepts usernames and email in the same input and it can validate each.

packages/playground/src/utils/validators.ts
samaradel: I meant here it could accept both and validate if it's not a valid email or username with a special character.

packages/playground/src/utils/validators.ts
samaradel: thanks, the regex was implemented separately in the validator function and used in the input validation.

packages/playground/tests/utils/isValidSmtp.test.ts
samaradel: done in the current validation rules.

packages/playground/tests/utils/isValidSmtp.test.ts
samaradel: Of course not but I used one of them to test as I can't test all of them (too many cases).

packages/playground/tests/utils/isValidSmtp.test.ts
samaradel: noted in the current validation.

packages/playground/src/components/smtp_server.vue
```diff
@@ -20,17 +20,23 @@
       <input-validator
         :value="$props.modelValue.username"
         :rules="[
-          validators.required('Email is required.'),
-          validators.isEmail('Please provide a valid email address.'),
-          v => {
-            return isDiscourse ? undefined : validators.isEmail('Please provide a valid email address.')(v);
+          validators.required('Email or Username is required.'),
+          validators.minLength('Username must be at least 2 characters.', 2),
+          validators.maxLength('Username must be at least 50 characters.', 50),
```
0oM4R: we shouldn't use the `at least` on max length

0oM4R: also is this validation on user name only? what if the email exceeds the 50 char ? the length validation message will appear

0oM4R (changes requested):
the validation need a revisit  
Screenshot from 2024-10-08 11-13-30
Screenshot from 2024-10-08 11-13-39

packages/playground/src/components/smtp_server.vue
samaradel: done

0oM4R (changes requested):
> * Check if the username or email can start with anything except an alphabetical character.

email pass the validation starting with number, also please add this test case to unit test
image

0oM4R:
Screenshot from 2024-10-17 16-05-18
this field should got validated on submit i think 

[Screencast from 2024-10-17 16-10-24.webm](https://github.com/user-attachments/assets/64f544b4-925e-40ba-a0f4-8b6a5cdccd3b)

samaradel:
@0oM4R ok, open an issue for this please

0oM4R (approved):

amiraabouhadid (approved):

Issue #2454, opened 2024-03-18
Add workload unit tests


zaelgohary:
### Which package/s are you suggesting this feature for?

grid_client

### Describe the solution you'd like

Add [workload.ts](https://github.com/threefoldtech/tfgrid-sdk-ts/blob/development/packages/grid_client/src/zos/workload.ts) unit tests

maayarosama:
1. Work Completed:
Added six test cases for workload class

maayarosama:
### Work in Progress (WIP):
- instead of re-writing the test for each type, I implemented a helper function That includes a switch-case to initialize the appropriate workload type. And I'm trying to find a way to invoke the same tests with different types

maayarosama:
### Work Completed
- Tests are now running on all workloadtypes 
- I initialized an instance for the the workloadtypes except 3

### Work in Progress (WIP):
- I need to initialize a qsfs , network and ipv4 instance for all the tests to pass
- 15 tests are failing while 40 is passing 
![Screenshot from 2024-09-25 18-04-13](https://github.com/user-attachments/assets/9f6d2dc2-4a78-438b-8128-4e597ac2e72b)

maayarosama:
###Work in Progress (WIP):

The following tests are failing so the pr is drafted till I fix them

Workload Tests for network
✕ should correctly serialize and deserialize Workload (7 ms)
✕ should correctly compute the challenge string (2 ms)
✕ should handle invalid DeploymentResult (25 ms)
Workload Tests for qsfs
✕ should correctly serialize and deserialize Workload (5 ms)
✕ should correctly compute the challenge string (1 ms)
✕ should handle invalid DeploymentResult (14 ms)

After some investigation, the challenges fail still not sure why

maayarosama:
### Work completed
- fixed all network's tests
- fixed most of qsfs tests
    -  should correctly compute the challenge string (1 ms)
    - should handle invalid DeploymentResult (14 ms)
### Work In Progress 
-    The following test is still failling ✕ should correctly serialize and deserialize Workload (12 ms)

maayarosama:
### Work completed
- Fixed ✕ should correctly serialize and deserialize Workload in qsfs
- refactored test file

khaledyoussef24:
verified both locally and on the workflow
also verified statically 
the tests are now all passing 

[...]

PR #3403, opened 2024-09-11
Adding workload unit tests

maayarosama:
### Description

Added six test case for workload 
### Changes
Test cases are:
-  should create a valid Workload instance
- should handle valid Workload properties
- should correctly serialize and deserialize Workload
-  should compute the challenge string correctly 
- should handle invalid Workload type
- should handle invalid DeploymentResult 

### Related Issues

- #2454
### Tested Scenarios

Screenshot from 2024-09-11 16-17-20

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation


packages/grid_client/tests/modules/workload.test.ts
```diff
@@ -0,0 +1,194 @@
...
+  const computeCapacity = new ComputeCapacity();
+  const disks = new Mount();
+
+  const rootfs_size = 2;
+
+  switch (type) {
+    case WorkloadTypes.zmachine:
+      instance = new Zmachine();
+      computeCapacity.cpu = 1;
+      computeCapacity.memory = 256 * 1024 ** 2;
```
amiraabouhadid: might be a good idea to declare 1024 as a constant as it is used more than once

maayarosama: > might be a good idea to declare 1024 as a constant as it is used more than once

I think this's not needed as I only used twice and the 1024 indicates the number of bytes

amiraabouhadid (commented):
Why aren't there test cases for the zmount and qsfs workload type?

maayarosama:
> Why aren't there test cases for the zmount and qsfs workload type?

Please elaborate

amiraabouhadid:
> > Why aren't there test cases for the zmount and qsfs workload type?
> 
> Please elaborate

why do you only have tests for workload type zmachine when there are other types we can test as well? I think we should write more tests that test workload types other than zmachine for a more extensive test coverage.

ramezsaeed (changes requested):
Why this still Draft PR.
also recommend Amira' comment here:
https://github.com/threefoldtech/tfgrid-sdk-ts/pull/3403#issuecomment-2355302197

maayarosama:
> Why this still Draft PR. also recommend Amira' comment here: [#3403 (comment)](https://github.com/threefoldtech/tfgrid-sdk-ts/pull/3403#issuecomment-2355302197)

The following tests are failing so the pr is drafted till I fix them
 - Workload Tests for network
    ✕ should correctly serialize and deserialize Workload (7 ms)
    ✕ should correctly compute the challenge string (2 ms)
    ✕ should handle invalid DeploymentResult (25 ms)
 -  Workload Tests for qsfs
    ✕ should correctly serialize and deserialize Workload (5 ms)
    ✕ should correctly compute the challenge string (1 ms)
    ✕ should handle invalid DeploymentResult (14 ms)

ramezsaeed (approved):

amiraabouhadid (commented):
test pass 
image
please remove changes in minting_view as they are not related to this pr

packages/grid_client/src/zos/znet.ts
```diff
@@ -55,4 +55,4 @@ class Znet extends WorkloadData {
   }
 }
 
-export { Znet, Peer };
+export { Znet, Peer, Mycelium };
```
amiraabouhadid, line 58: why do you export mycelium here? Is this change necessary?

maayarosama: It's necessary as we use it here https://github.com/threefoldtech/tfgrid-sdk-ts/blob/7974177e30c01e02f9b6429fdf418515cd63bde9/packages/grid_client/tests/modules/workload.test.ts#L48

packages/grid_client/tests/modules/workload.test.ts
```diff
@@ -0,0 +1,251 @@
...
+
+  test("should correctly compute the challenge string", () => {
+    const expectedChallenge =
+      workload.version.toString() +
+      workload.name +
+      workload.type.toString() +
+      workload.metadata +
+      workload.description +
+      workload.data.challenge();
+    console.log("data", workload.data);
```
0oM4R: do we need this log ?

amiraabouhadid:
@maayarosama please remove changes in minting _view as they are irrelevant

amiraabouhadid (approved):

0oM4R (changes requested):
tests are failing with me
image

0oM4R (approved):
image

passing now thank you

Issue #3411, opened 2024-09-17
🐞 [Bug]: Cannot read properties of null (reading 'clientOptions')


zaelgohary:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### which package/s did you face the problem with?

Dashboard

### What happened?

![image](https://github.com/user-attachments/assets/6ab8d958-1a72-4c10-9a0f-57d54ac735b0)

### Steps To Reproduce

1. Login to Dashboard
2. logout

### which network/s did you face the problem on?

Dev

### version

staging

### Relevant screenshots/screen records

attached above

### Relevant log output

```console
[Uncaught Error]
- Message: Cannot read properties of null (reading 'clientOptions')
- Stack: TypeError: Cannot read properties of null (reading 'clientOptions')
   at setTimeouts (http://localhost:5173/src/App.vue?t=1726578818318:109:36)
   at http://localhost:5173/src/App.vue?t=1726578818318:101:28
   at callWithErrorHandling (http://localhost:5173/node_modules/.vite/deps/chunk-S4KY57GZ.js?v=c45f874f:1584:18)
   at callWithAsyncErrorHandling (http://localhost:5173/node_modules/.vite/deps/chunk-S4KY57GZ.js?v=c45f874f:1592:17)
   at job (http://localhost:5173/node_modules/.vite/deps/chunk-S4KY57GZ.js?v=c45f874f:3181:9)
   at callWithErrorHandling (http://localhost:5173/node_modules/.vite/deps/chunk-S4KY57GZ.js?v=c45f874f:1584:32)
   at flushJobs (http://localhost:5173/node_modules/.vite/deps/chunk-S4KY57GZ.js?v=c45f874f:1782:9)
- Constructor: TypeError
```

khaledyoussef24:
verified on devnet
Version a78daae

console does not get the errors any longer.

https://github.com/user-attachments/assets/e8ef41bd-44e4-4fa5-86c8-9b4d52c9946b

PR #3412, opened 2024-09-17
check for client before calling clientoptions

amiraabouhadid:
### Changes

- fixed if conditions to not read null 

### Related Issues

#3411 

### Tested Scenarios

- tested when logged in and after logging out
- tested that deployment timeout isn't affected

### DIFF
--- a/packages/playground/src/App.vue
+++ b/packages/playground/src/App.vue
@@ -339,20 +339,16 @@ async function setTimeouts() {
 
   const localStorageDeploymentTimeout = localStorage.getItem(LocalStorageSettingsKey.TIMEOUT_DEPLOYMENT_KEY);
 
-  if (client) {
-    const clientOptions = client.clientOptions;
-    if (clientOptions) {
-      const deploymentTimeoutMinutes = clientOptions.deploymentTimeoutMinutes;
-      if (!localStorageDeploymentTimeout && deploymentTimeoutMinutes) {
-        localStorage.setItem(LocalStorageSettingsKey.TIMEOUT_DEPLOYMENT_KEY, `${+deploymentTimeoutMinutes * 60}`);
-      } else {
-        client.clientOptions.deploymentTimeoutMinutes = +localStorageDeploymentTimeout! / 60;
-        await client.connect();
-      }
+  if (client && client.clientOptions) {
+    const deploymentTimeoutMinutes = client.clientOptions.deploymentTimeoutMinutes;
+    if (!localStorageDeploymentTimeout && deploymentTimeoutMinutes) {
+      localStorage.setItem(LocalStorageSettingsKey.TIMEOUT_DEPLOYMENT_KEY, `${+deploymentTimeoutMinutes * 60}`);
+    } else {
+      client.clientOptions.deploymentTimeoutMinutes = +localStorageDeploymentTimeout! / 60;
+      await client.connect();
     }
   }
 }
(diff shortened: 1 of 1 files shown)


packages/playground/src/App.vue
```diff
@@ -336,16 +336,19 @@ async function setTimeouts() {
...
+    if (clientOptions) {
+      const deploymentTimeoutMinutes = clientOptions.deploymentTimeoutMinutes;
+      if (!localStorageDeploymentTimeout && deploymentTimeoutMinutes) {
+        localStorage.setItem(LocalStorageSettingsKey.TIMEOUT_DEPLOYMENT_KEY, `${+deploymentTimeoutMinutes * 60}`);
+      } else {
+        client.clientOptions.deploymentTimeoutMinutes = +localStorageDeploymentTimeout! / 60;
+        await client.connect();
+      }
     }
   }
```
zaelgohary, line 350:
```suggestion
  if (client && client.clientOptions) {
    let deploymentTimeoutMinutes = client.clientOptions.deploymentTimeoutMinutes;
    if (!localStorageDeploymentTimeout && deploymentTimeoutMinutes) {
      localStorage.setItem(LocalStorageSettingsKey.TIMEOUT_DEPLOYMENT_KEY, `${+deploymentTimeoutMinutes * 60}`);
    } else {
      deploymentTimeoutMinutes = +localStorageDeploymentTimeout! / 60;
      await client.connect();
    }
  }
```

zaelgohary: I tested this by logging then navigating to any page then loggin out. Works fine.

packages/playground/src/App.vue
```diff
@@ -336,16 +336,19 @@ async function setTimeouts() {
...
-      await client.connect();
+
+  if (client) {
+    const clientOptions = client.clientOptions;
+    if (clientOptions) {
+      const deploymentTimeoutMinutes = clientOptions.deploymentTimeoutMinutes;
+      if (!localStorageDeploymentTimeout && deploymentTimeoutMinutes) {
+        localStorage.setItem(LocalStorageSettingsKey.TIMEOUT_DEPLOYMENT_KEY, `${+deploymentTimeoutMinutes * 60}`);
+      } else {
+        client.clientOptions.deploymentTimeoutMinutes = +localStorageDeploymentTimeout! / 60;
```
maayarosama:
```suggestion
        clientOptions.deploymentTimeoutMinutes = +localStorageDeploymentTimeout! / 60;
```
you already have a clientOptions variable that you can use directly

maayarosama (dismissed):

packages/playground/src/App.vue
amiraabouhadid: doesn't update the timeout on the client. Note how I was still able to deploy here even though timeout is 3 seconds. 
Screenshot from 2024-09-21 21-34-16

amiraabouhadid:
@zaelgohary your suggestion fails to update the timeout on the client. I was able to deploy even though deploymentTimeout is set to 3 seconds in localstorage. 
Screenshot from 2024-09-21 21-34-16
Found issue in your suggestion and fixed. deploymentTimeout needs to be set on the client not a variable taken from a client.

zaelgohary (approved):

maayarosama (approved):

samaradel (approved):

Issue #3425, opened 2024-09-19
🐞 [Bug]: Network contract exist after deployment deletion


zaelgohary:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### which package/s did you face the problem with?

Dashboard

### What happened?

Jenkins network contracts exist after I deleted the deployment

### Steps To Reproduce

1. Go to dashboard
2. Deploy Jenkins
3. Delete deployment
4. Check network contracts

### which network/s did you face the problem on?

Dev

### version

1f123d2

### Relevant screenshots/screen records

[jenkins_deletion.webm](https://github.com/user-attachments/assets/7e67bb6a-dc07-437a-a655-d9b62f33fe93)

### Relevant log output

```console
-
```

maayarosama:
Might be related to contract type

ramezsaeed:
Failed to load all contracts after delete the jenkins instance !!
Devnet
version: 8529bf1

[Screencast from 09-29-2024 11:49:35 AM.webm](https://github.com/user-attachments/assets/25f4ab58-80b9-4836-b75b-d701d3cacde0)


![Screenshot from 2024-09-29 11-55-25](https://github.com/user-attachments/assets/acaae002-610a-4b3c-ba5a-5ec9f5e3b2d5)
![Screenshot from 2024-09-29 11-53-55](https://github.com/user-attachments/assets/4503d735-f050-436e-a3b8-18940e4ab980)
![Screenshot from 2024-09-29 11-51-21](https://github.com/user-attachments/assets/bbc0fe68-2d50-45b4-86ad-43324fcd72ea)

zaelgohary:
> Failed to load all contracts after delete the jenkins instance !! Devnet version: [8529bf1]

This issue is regarding the lazy loading and should be fixed [here](https://github.com/threefoldtech/tfgrid-sdk-ts/issues/3322)

ramezsaeed:
Verified:
- Devnet
- b760b10

Issue #3378, opened 2024-09-05
🐞 [Bug]: The contract type on the contract page not always correct


AhmedHanafy725:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### which package/s did you face the problem with?

Dashboard

### What happened?

the node contracts that contain `network` or `gateway` have wrong type `vm`

### Steps To Reproduce

- deploy micro vm
- add gateway to this vm
- check the contracts created in the contracts page

### which network/s did you face the problem on?

Dev

### version

development

### Relevant screenshots/screen records

these contracts are shown as vms while they are gateways
![image](https://github.com/user-attachments/assets/c9505904-168f-44ab-9026-e2e599a1919d)

### Relevant log output

```console
-
```

0oM4R:
we are using grid proxy to list the contracts, and using the type in the "details/deployment_date" 
in dev net all contracts have the type vm, while from their names they are starting with nw-- should be network

![image](https://github.com/user-attachments/assets/a6e01760-42f0-4568-a1f6-09df9f5174f7)


while on main net: 
![image](https://github.com/user-attachments/assets/1b4881ba-1e41-487b-a4dd-4e7897f0410f)

0oM4R:
@Omarabdul3ziz  can you confirm that gridproxy works fine with that point please?

0oM4R:
Issue update: the network contracts got updated to be vm  in deployment list, will continue investigating what is the reason of that call

0oM4R:
Issue update : 
in this pr we have some changes that effect the migrations, the networks contracts got listed and updated later in the migration, need to understand the migration and listing flow then will start to resolve the issue
will keep the estimated time as 3 hours of investigation

khaledyoussef24:
verified on devnet
Version a78daae

![Image](https://github.com/user-attachments/assets/054ea6be-b9ab-4108-900d-e1a1407696e4)
![Image](https://github.com/user-attachments/assets/007be5bb-d95d-47aa-99c4-2775e995001a)
![Image](https://github.com/user-attachments/assets/637894db-c056-42b5-b6a0-028c0ea188d4)
![Image](https://github.com/user-attachments/assets/b06a88d7-b723-48b8-bfea-2a24132a4d2a)
![Image](https://github.com/user-attachments/assets/600f3339-0bff-4bf4-89a7-ed26f4fd7efb)





[...]

Issue #3423, opened 2024-09-19
🐞 [Bug]: Gitea solutions shows as "vm" in the solution type in contracts


khaledyoussef24:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### which package/s did you face the problem with?

Dashboard

### What happened?

after deploying gitea and checking contracts 
the solution type is showed as vm

### which network/s did you face the problem on?

Dev

### version

1f123d2

### Relevant screenshots/screen records

<img width="1338" alt="Screenshot 2024-09-19 at 10 27 48 AM" src="https://github.com/user-attachments/assets/06f7b641-c429-442a-9af4-d5cb2eefb873">
<img width="1680" alt="Screenshot 2024-09-19 at 10 31 12 AM" src="https://github.com/user-attachments/assets/e8f41eb9-23bc-4c9d-9537-ca16cae25b42">
<img width="1680" alt="Screenshot 2024-09-19 at 10 31 14 AM" src="https://github.com/user-attachments/assets/ad919544-2f69-43a9-bb50-f43439a0d39a">
<img width="1680" alt="Screenshot 2024-09-19 at 10 31 29 AM" src="https://github.com/user-attachments/assets/3ceb6797-7c54-44c2-a385-90433a4cacd3">

### Relevant log output

```console
na
```

khaledyoussef24:
as mentioned in [#3397](https://github.com/threefoldtech/tfgrid-sdk-ts/issues/3397)
4 contracts were created

maayarosama:
Investigation and Solution:
Still investigating from the grid client side, no actual progress

maayarosama:
It might be fixed when #3378 is fixed, I'll have to test it again

maayarosama:
### Work Completed
@0oM4R figured out that the issue was [here](https://github.com/threefoldtech/tfgrid-sdk-ts/blob/development/packages/playground/src/types/index.ts#L132)

khaledyoussef24:
verified on devnet
version : ** b760b10**

deployed a gittea and contracts type is now showing right![Image](https://github.com/user-attachments/assets/be06dca8-f1e2-4861-be9b-4535e789163e)
![Image](https://github.com/user-attachments/assets/d0d2f952-047d-4f78-983a-c27e669dff1b)
[testcase](https://app.testlodge.com/a/26076/projects/40893/suites/234374?expand_section=422332#case_4311200)

PR #3456, opened 2024-09-25
Fix contracts types update 

0oM4R:
fix project name to be dedicated to the passed module name

### Description

what was the issue ?
In deployment list, all contracts got overwritten to be of type `vm`
**The cause** : while migrating machines of type `FullVM` from old playground, we are setting `moduleName` to `FullVm`.
https://github.com/threefoldtech/tfgrid-sdk-ts/blob/ca89db4265da0d544f9ece32b57ea402b6aacb4c/packages/playground/src/utils/migration.ts#L11
but this module name is not exist on [`moduleNames` enum](https://github.com/threefoldtech/tfgrid-sdk-ts/blob/ca89db4265da0d544f9ece32b57ea402b6aacb4c/packages/grid_client/src/modules/base.ts#L28-L35), 
then while listing the contracts we map the passed module name to the moduleNames enum, so the result is undefined, 
https://github.com/threefoldtech/tfgrid-sdk-ts/blob/ca89db4265da0d544f9ece32b57ea402b6aacb4c/packages/grid_client/src/modules/base.ts#L262
and call list function with type `undefine`
https://github.com/threefoldtech/tfgrid-sdk-ts/blob/ca89db4265da0d544f9ece32b57ea402b6aacb4c/packages/grid_client/src/modules/base.ts#L264-L268
so in this case all contract on that project name got listed as the listing function checks for the passed type which is in this case undefined so we are listing all the contracts on this project name 
https://github.com/threefoldtech/tfgrid-sdk-ts/blob/ca89db4265da0d544f9ece32b57ea402b6aacb4c/packages/grid_client/src/clients/tf-grid/contracts.ts#L176-L179

so the expected to list the contracts with type `FullVm` then change its type to vm but now we are listing all contracts on the project name then overwrite its type to be vm

### Changes

add nullish coalescing operator to make sure the module name never got undefined if it failed to map

### Related Issues

- #3378
- #3425

### Tested Scenarios

- deployed vm, wait for it to got listed in the deployment list, check the network contract type
- use `grid.networks.list` to list my contracts

### Documentation PR

### Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation

### DIFF
--- a/packages/grid_client/src/modules/base.ts
+++ b/packages/grid_client/src/modules/base.ts
@@ -259,7 +259,7 @@ class BaseModule {
    * @returns {Promise<Required<GqlNodeContract>[]>} - A promise that resolves to an array of contracts associated with the current module.
    */
   private async getMyContracts(fetch = false): Promise<Required<GqlNodeContract>[]> {
-    const moduleName = modulesNames[this.moduleName];
+    const moduleName = modulesNames[this.moduleName] ?? this.moduleName;
     if (fetch || !this.contracts) {
       let contracts = await this.tfClient.contracts.listMyNodeContracts({
         graphqlURL: this.config.graphqlURL,


AhmedHanafy725 (approved):

zaelgohary (dismissed):
Works fine.

image

zaelgohary (commented):
I deployed a Gitea instance and it's still shown as VM.

image

0oM4R:
> I deployed a Gitea instance and it's still shown as VM.
> 
> image

sorry my mistake added the wrong issue, the issue will be fixed on #3423

zaelgohary (approved):
Tested deleting a deployment and its network deleted as well. Good job, @0oM4R!

Issue #3367, opened 2024-09-04
🐞 [Bug]: Use the same validation for secret seed everywhere


AhmedHanafy725:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### which package/s did you face the problem with?

Dashboard

### What happened?

the validation for hex seed is different between client and the input field

### Steps To Reproduce

- use hex seed in the mnemonic or hex seed field in the dashboard without `0x` at the beginning.

### which network/s did you face the problem on?

Dev

### version

development

### Relevant screenshots/screen records

![image](https://github.com/user-attachments/assets/3b6cb3fe-28d0-4782-a362-4860b928679e)

### Relevant log output

```console
-
```

samaradel:
Works for me 
[Screencast from 11-09-24 17:21:19.webm](https://github.com/user-attachments/assets/7ab3e7d4-b3bf-4267-98fd-63aa0abd71cd)
[Screencast from 12-09-24 15:16:23.webm](https://github.com/user-attachments/assets/255acc37-1233-4b2b-849d-66cc990a388e)

xmonader:
@AhmedHanafy725

AhmedHanafy725:
@samaradel please try with a new seed and try to activate the account.

samaradel:
### Work in Progress (WIP):
Added function to force the input to include ```0x``` but with final test it converts the inserted mnemonic into a hex seed.

samaradel:
### Work in Progress (WIP):
faced an error from ```tf-chain``` while activating hex seed without ```0x``` investigating...
![Image](https://github.com/user-attachments/assets/11aea216-2933-43c8-8c04-82aec4761ce7)

samaradel:
### Work Completed:
- Validate the hex seed that passed to activate and create an account function.
- Fix the hex seed label in the profile dialog.

samaradel:
### Work in Progress (WIP):
while refactoring ```checkinput``` function in ```tfchain``` the function stopped working.. investigating...

samaradel:
### Work Completed:
Double check of mnemonic and hex seed that passed to validation function.

khaledyoussef24:
verified on devnet
Version d757d29

user can now login with the seed 0x

![Image](https://github.com/user-attachments/assets/782fe65b-db31-42f8-aa0e-9115062dc1c1)
[...]

PR #3474, opened 2024-09-26
Validate hex seed in case activate account

samaradel:
### Description

- Validate the hex seed that passed to activate and create an account function.
- Fix the hex seed label in the profile dialog.
- Refactor code as it could exit early if the case matched.

### Changes

[Screencast from 26-09-24 16:39:33.webm](https://github.com/user-attachments/assets/97954695-4ad3-4466-837c-36e00a68103f)

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/3367
### Tested Scenarios

- Using new hex seed without ```0x``` in the beginning and activate the account.

### Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation

### DIFF
--- a/packages/playground/src/weblets/profile_manager.vue
+++ b/packages/playground/src/weblets/profile_manager.vue
@@ -130,7 +130,6 @@
                             validateMnemonic(v) ||
                             ((v.length === 64 || v.length === 66) && isAddress(v.length === 66 ? v : `0x${v}`))
                           ) {
-                            mnemonic = toHexSeed(mnemonic);
                             getEmail();
                             return;
                           }
@@ -423,7 +422,7 @@
 </template>
 <script lang="ts" setup>
 import { isAddress } from "@polkadot/util-crypto";
-import { KeypairType, toHexSeed } from "@threefold/grid_client";
+import { KeypairType } from "@threefold/grid_client";
 import { validateMnemonic } from "bip39";
 import Cryptr from "cryptr";
 import { marked } from "marked";
--- a/packages/playground/src/weblets/profile_manager.vue
+++ b/packages/playground/src/weblets/profile_manager.vue
@@ -427,7 +427,7 @@
 </template>
 <script lang="ts" setup>
 import { isAddress } from "@polkadot/util-crypto";
-import { KeypairType, toHexSeed } from "@threefold/grid_client";
+import { KeypairType } from "@threefold/grid_client";
 import { validateMnemonic } from "bip39";
 import Cryptr from "cryptr";
 import { marked } from "marked";
--- a/packages/playground/src/weblets/profile_manager.vue
+++ b/packages/playground/src/weblets/profile_manager.vue
@@ -130,6 +130,7 @@
                             validateMnemonic(v) ||
                             ((v.length === 64 || v.length === 66) && isAddress(v.length === 66 ? v : `0x${v}`))
                           ) {
+                            mnemonic = toHexSeed(mnemonic);
                             getEmail();
                             return;
                           }
@@ -422,7 +423,7 @@
 </template>
 <script lang="ts" setup>
 import { isAddress } from "@polkadot/util-crypto";
-import { KeypairType } from "@threefold/grid_client";
+import { KeypairType, toHexSeed } from "@threefold/grid_client";
 import { validateMnemonic } from "bip39";
 import Cryptr from "cryptr";
 import { marked } from "marked";
--- a/packages/playground/src/weblets/profile_manager.vue
+++ b/packages/playground/src/weblets/profile_manager.vue
@@ -427,7 +427,7 @@
 </template>
 <script lang="ts" setup>
 import { isAddress } from "@polkadot/util-crypto";
-import { KeypairType } from "@threefold/grid_client";
+import { KeypairType, toHexSeed } from "@threefold/grid_client";
 import { validateMnemonic } from "bip39";
 import Cryptr from "cryptr";
 import { marked } from "marked";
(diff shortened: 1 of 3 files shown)


packages/tfchain_client/src/client.ts
```diff
@@ -294,11 +300,11 @@ class Client extends QueryClient {
           // seed shouldn't have spaces
           throw new ValidationError("Invalid mnemonic! Must be bip39 compliant");
 
-        if (!this.mnemonicOrSecret.startsWith("0x"))
-          throw new ValidationError("Invalid secret seed. secret seed should starts with 0x");
-        const secret = this.mnemonicOrSecret.substring(2);
-        if (secret.length !== 64) throw new ValidationError("Invalid secret length. Secret length should be 64");
-        if (!isValidSeed(secret)) throw new ValidationError("Invalid secret seed");
+        if (this.mnemonicOrSecret.startsWith("0x")) {
```
xmonader: too many nested level here, I suggest you change the checkInputs to be something like

```
function checkInputs(): void {
  // Validate URL presence
  if (!this.url) {
    throw new ValidationError("URL is required.");
  }

  // Validate Keypair Type
  if (!SUPPORTED_KEYPAIR_TYPES.includes(this.keypairType)) {
    const validTypes = SUPPORTED_KEYPAIR_TYPES.join(", ");
    throw new ValidationError(
      `Invalid keypair type: "${this.keypairType}". Valid options are: ${validTypes}.`
    );
  }

  // Ensure either mnemonicOrSecret or extSigner is provided
  if (!this.mnemonicOrSecret && !this.extSigner) {
    throw new ValidationError("Either 'mnemonicOrSecret' or 'extSigner' must be provided.");
  }

  // Early exit if mnemonicOrSecret is not provided
  if (!this.mnemonicOrSecret) {
    return;
  }

  // Early exit if mnemonicOrSecret indicates an extension signer
  if (this.mnemonicOrSecret.startsWith("//")) {
    return;
  }

  // Process secret by removing '0x' prefix if present
  const secret = this.mnemonicOrSecret.startsWith("0x")
    ? this.mnemonicOrSecret.slice(2)
    : this.mnemonicOrSecret;

  // Validate mnemonic
  if (!validateMnemonic(this.mnemonicOrSecret)) {
    throw new ValidationError("Invalid mnemonic provided.");
  }

  // Validate secret length
  if (secret.length !== 64) {
    throw new ValidationError("Secret must be exactly 64 characters long.");
  }

  // Validate seed derived from the secret
  if (!isValidSeed(secret)) {
    throw new ValidationError("Invalid seed derived from the secret.");
  }
}

```

So you can make use of early exits

samaradel: done

0oM4R:
Can you please update the PR description with the new changes? and explanation would be great addition.

xmonader (approved):

packages/playground/src/weblets/profile_manager.vue
```diff
@@ -722,7 +728,12 @@ async function activateAccount() {
   activatingAccount.value = true;
   activating.value = true;
   try {
-    await activateAccountAndCreateTwin(mnemonic.value);
+    const mnemonicValue = validateMnemonic(mnemonic.value)
+      ? mnemonic.value
+      : mnemonic.value.length === 66
+      ? mnemonic.value
+      : `0x${mnemonic.value}`;
```
0oM4R: should we use if/else if ? i think it will be more readable

packages/playground/src/weblets/profile_manager.vue
```diff
@@ -794,9 +805,9 @@ async function storeAndLogin() {
     await activate(mnemonic.value, keypairType.value);
   } catch (e) {
     if (e instanceof TwinNotExistError) {
-      isNonActiveMnemonic.value = true;
       openAcceptTerms.value = true;
       termsLoading.value = true;
+      isNonActiveMnemonic.value = true;
```
0oM4R, line 810: is there a reason for this change ?

packages/tfchain_client/src/client.ts
```diff
@@ -276,29 +276,38 @@ class Client extends QueryClient {
   }
 
   checkInputs(): void {
-    if (!this.url) throw new ValidationError("url should be provided");
+    if (!this.url) {
+      throw new ValidationError("URL is required.");
```
0oM4R: please make the error msg more clear, what is the url required, i remember we saw this error before and we didn't get it right

0oM4R (commented):
verified creating account from hex with the 0x and without it, all works fine. 
great job ya Samar, just few suggestions

packages/playground/src/weblets/profile_manager.vue
samaradel: I think it's good, less code, and the results depend on the previous one. it's fine.
@AhmedHanafy725 what do you think ?

packages/playground/src/weblets/profile_manager.vue
samaradel: yes, it turns ```shouldActivateAccount``` to true which fire ```activateAccount``` function

packages/tfchain_client/src/client.ts
samaradel: It validates input that you should enter the URL, and it fires when input is empty.
it's like the required attribute that we add to any input.
still not clear ?

xmonader: > please make the error msg more clear, what is the url required, i remember we saw this error before and we didn't get it right

Legit, Let's make it `TFChain URL is required`

packages/playground/src/weblets/profile_manager.vue
AhmedHanafy725: it's fine but what confuses me is the variable name, it can be mnemonic or hex seed

packages/playground/src/weblets/profile_manager.vue
0oM4R: I mean, why we move it down? it was already on 3 lines above, its fine but i just want to know if there is a logical reason

samaradel: to fire createAccount function after accepting terms.

packages/tfchain_client/src/client.ts
samaradel: done

0oM4R (approved):
activated two accounts, one with 0x prefix and another without, both works fine

Issue #3322, opened 2024-08-20
🐞 [Bug]: `Lazy Loading` Can not load Contracts 


khaledyoussef24:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### which package/s did you face the problem with?

Dashboard

### What happened?

loading contracts showing unclear error

### which network/s did you face the problem on?

Dev

### version

a542a73

### Relevant screenshots/screen records

<img width="1680" alt="Screenshot 2024-08-20 at 5 41 00 PM" src="https://github.com/user-attachments/assets/b85f0c67-69f2-4fa9-ae4f-5c2f0e8b4b02">

### Relevant log output

```console
.
```

amiraabouhadid:
check url error in screenshot (4th)

0oM4R:
Issue update:
I found that this happens because the `TfContractsList` is calling the clients/index.ts before setting the stacks urls, this happens randomly will continue investigating

0oM4R:
What happens: 
many components and utils files are using the clients/index.ts
https://github.com/threefoldtech/tfgrid-sdk-ts/blob/c573f1a06b0b71fbc8162ace62c455e3a00ca082/packages/playground/src/clients/index.ts#L1-L9
this exports the three clients, that was fine until we introduced the `Url Monitoring` feature this features set the urls in the `window.env` after pinging them on mount the app.
in some case the file `index.ts` got loaded before initializing urls values so the created clients got undefined url.
the issue is not from lazy loading, the issue is we are using the clients in many places and we aren't sure about its value!
the lazy load feature is great and helps for improving the performance, and i suggest to use it in all large components but we need to handle the clients we are using without checking if they are working or not!

Suggestion:
store the clients on store as grid, This will cost many changes but i think it worth that

ramezsaeed:
Got the same erorrs

![Screenshot from 2024-09-29 11-53-55](https://github.com/user-attachments/assets/e2ee0cda-3fc5-4a6d-8974-9dadfb925bea)
![Screenshot from 2024-09-29 11-55-25](https://github.com/user-attachments/assets/b2280d57-3fdd-495a-bbdb-a15ae6382849)
![Screenshot from 2024-09-29 11-51-21](https://github.com/user-attachments/assets/3da4c530-1a7e-4282-8af4-c66554224665)


[...]

Issue #3342, opened 2024-08-27
🐞 [Bug]: Weird errors on staging, and dashboard.dev  logs


0oM4R:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### which package/s did you face the problem with?

Dashboard

### What happened?

while opening the staging or deployed dashboard on **new incognito window**, you will see the errors on logs 

[Screencast from 2024-08-27 13-44-03.webm](https://github.com/user-attachments/assets/1a7de35f-bfab-47dc-8caa-3f85e0ec7868)

### which network/s did you face the problem on?

Dev

### version

2.5.0

### Relevant screenshots/screen records

[logs file](https://github.com/user-attachments/files/16762060/localhost-1724756165601.log)

### Relevant log output

```console
provided in previous section
```

0oM4R:
this is not related to lazy loading

xmonader:
@samaradel investigate and let us know the root cause of this please

samaradel:
update:
When `gridStore.client` is null initially, it can't access the `grid.twinId`, so it throws the error.

0oM4R:
I disagree with Samar, the `grid.twinId` is not accessed on any public pages, all weblets require the profile, and then the gridstore should be initialized 
Anyway, I tried to deploy a new version of the current staging version, which can be verified in http://185.206.122.44/#/

zaelgohary:
Verified on the mentioned machine. https://github.com/threefoldtech/tfgrid-sdk-ts/issues/3342#issuecomment-3376510520

PR #3475, opened 2024-09-26
Fix contracts page not working with lazy loading

MohamedElmdary:
### Description
Fix contracts page not working with lazy loading

### Changes
- fix: app was loading clients before checking is completed so we defer all components to avoid this behaviour
- fix: profile manage popup wasn't correctly loaded so we defer it's poping till it's fully loaded
image

### Related Issues
- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/3322

### Tested Scenarios
- check if error is happening due to lazy components for example; try to use url which is not yet set

### Documentation PR

#### Preliminary Checks:

- What about edge cases?

#### UI Checks:

- Does every button work?
- Does every validation work?

#### Code Quality Checks:

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/packages/playground/src/components/logger.vue
+++ b/packages/playground/src/components/logger.vue
@@ -203,7 +203,7 @@ export default {
     let _interceptorQueue: LI[] = [];
 
     async function interceptMessage(instance: LI) {
-      if (connectDB.value.error) {
+      if (connectDB?.value?.error) {
         _interceptorQueue.push(instance);
         return;
       }
--- a/packages/playground/src/components/logger.vue
+++ b/packages/playground/src/components/logger.vue
@@ -212,7 +212,19 @@ export default {
       const { logger: _, date: __, ...log } = instance;
 
       if (import.meta.env.DEV) {
-        if (log.messages.map(String).join().includes("vite") && log.type === "debug") {
+        if (
+          log.messages
+            .map(v => {
+              try {
+                return String(v);
+              } catch {
+                return "{ [[null proto]] }";
+              }
+            })
+            .join()
+            .includes("vite") &&
+          log.type === "debug"
+        ) {
           return;
         }
       }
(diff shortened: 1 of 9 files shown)


0oM4R:
image
the tft price not works fine looks like the issue is still exists

packages/playground/src/components/logger.vue
```diff
@@ -203,7 +203,7 @@ export default {
     let _interceptorQueue: LI[] = [];
 
     async function interceptMessage(instance: LI) {
-      if (connectDB.value.error) {
+      if (connectDB?.value?.error) {
```
0oM4R, line 206: there are still some errors coming from the logger 
https://github.com/threefoldtech/tfgrid-sdk-ts/blob/431d55a9b56dff5a2ffc4011f0da6bb704a465b0/packages/playground/src/components/logger.vue#L215
image

0oM4R:
can you please explain the solution please?

packages/playground/src/components/logger.vue
MohamedElmdary: this issue caused due to nullable prototype of object

image

MohamedElmdary:
> can you please explain the solution please?

here we define promise based lock
image

which allows us to await it later untill monitor service is resolved
image

awaiting it will block resolving component which blocking loading any clients file untill we release that lock

maayarosama (approved):
Works fine with me 
[Screencast from 2024-09-30 11-37-39.webm](https://github.com/user-attachments/assets/8ef530a7-10b4-4c7d-ba4b-e8fd123774ba)

0oM4R:
tested: 
- reload on contracts page 
- reload on deployments page 
- open contracts url in incognito
all works fine with me,  only the incognito have errors in console but this issue is reported  https://github.com/threefoldtech/tfgrid-sdk-ts/issues/3342

0oM4R (approved):

Issue #3406, opened 2024-09-12
🐞 [Bug]: Billing isn't reflected on the dashboard contracts


A-Harby:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### which package/s did you face the problem with?

Dashboard

### What happened?

I noticed that some of my contracts get delayed showing the billing value after a full billing cycle if it didn't show it at all.

Its billed on the TFChain and its showing the on chain and graphql.

### which network/s did you face the problem on?

Dev

### version

a948bfa

### Twin ID/s

9090

### Node ID/s

250

### Farm ID/s

1

### Contract ID/s

151438

### Relevant screenshots/screen records

![image](https://github.com/user-attachments/assets/dcbd9849-68c1-4089-962b-ce6e5069a374)

[dashboard (2).log](https://github.com/user-attachments/files/16980664/dashboard.2.log)

### Relevant log output

```console
{
  "contract_id": 151438,
  "twin_id": 9090,
  "type": "rent",
  "state": "Created",
  "created_at": 1726144326,
  "details": {
    "nodeId": 250,
    "farm_id": 1
  },
  "solutionName": "-",
  "solutionType": "-",
  "expiration": "-",
  "consumption": 0
}


{
  standardReserve: 17,937,800
  additionalReserve: 0
  standardOverdraft: 0
  additionalOverdraft: 0
  lastUpdatedSeconds: 1,726,147,368
  cycles: 1
}


with amount 17,937,800
```

zaelgohary:
The issue is with the created_at interface. Once fixed, the billing will be fixed.

amiraabouhadid:
problem in first time created billing report because billing rate is calculated by subtracting current from previous report

zaelgohary:
The issue happens in the first billing cycle only.

0oM4R:
I couldn't find the created_at interface issue but found that the issue is with accessing the contract field in the gql response 
![image](https://github.com/user-attachments/assets/644868c7-6623-4a6b-b738-1f75e115de7a)

so we need to update the `GqlConsumption` interface

khaledyoussef24:
verified on devnet
Version ffbb750

Billing is now reflected right on the dashboard
deployment list : 
![Image](https://github.com/user-attachments/assets/81d5fa25-d04d-4f92-aa3d-fd2eb3fc1ecb)
[...]

PR #3490, opened 2024-10-01
Fix: billing rate on first billing cycle

0oM4R:
-update GqlConsumption interface
-initialize duration with 1 to avoid accessed before initialization error -avoid divide by zero by using the || operator

### Description

the query response have each contract type separated while we are trying to access them indie contracts property
the solution is to access the contracts types directly 

### Changes
-update GqlConsumption interface
-initialize duration with 1 to avoid accessed before initialization error -avoid divide by zero by using the || operator

### Related Issues

- #3406 

### Tested Scenarios

get contract on its first billing cycle, and other with more than one billing report 

Screenshot from 2024-10-01 15-23-43
Screenshot from 2024-10-01 15-24-30

Screenshot from 2024-10-01 16-15-26
Screenshot from 2024-10-01 16-15-32

### DIFF
--- a/packages/grid_client/src/clients/tf-grid/contracts.ts
+++ b/packages/grid_client/src/clients/tf-grid/contracts.ts
@@ -67,7 +67,9 @@ export interface GqlContracts {
 }
 
 export interface GqlConsumption {
-  contracts: GqlContracts;
+  nameContracts: GqlNameContract[];
+  nodeContracts: GqlNodeContract[];
+  rentContracts: GqlRentContract[];
   contractBillReports: GqlContractBillReports[];
 }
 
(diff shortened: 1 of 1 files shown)


samaradel (approved):

packages/grid_client/src/clients/tf-grid/contracts.ts
```diff
@@ -67,7 +67,9 @@ export interface GqlContracts {
 }
 
 export interface GqlConsumption {
-  contracts: GqlContracts;
+  nameContracts: GqlNameContract[];
+  nodeContracts: GqlNodeContract[];
+  rentContracts: GqlRentContract[];
```
AhmedHanafy725: you can extend this interface instead `GqlContracts`

AhmedHanafy725 (approved):

amiraabouhadid (approved):

Issue #3408, opened 2024-09-16
🐞 [Bug]: load logs is taking forever 


amiraabouhadid:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### which package/s did you face the problem with?

Dashboard

### What happened?

i deployed a micro vm, opened the logs, there were none. pressed load logs and it is still spinning. it has been more than 10 minutes since I've pressed on it 

### Steps To Reproduce

1. run playground
2. open up logs
3. press load logs
4. wait forever

### which network/s did you face the problem on?

Dev

### version

current

### Relevant screenshots/screen records

![Uploading image.png…]()

### Relevant log output

```console
no logs
```

amiraabouhadid:
Investigating why logs are not updated after mount

amiraabouhadid:
Found two unhandled cases, added handling. Will test next.

amiraabouhadid:
Load logs still keeps loading on application mount in some cases, debugging

amiraabouhadid:
fix load logs on application mount. However, logs still sometimes do not load even after deploying a VM so investigating why.

amiraabouhadid:
this console error is now handled and removed as well 
![image](https://github.com/user-attachments/assets/66318c78-d518-46f0-93de-10abd31fb923)

A-Harby:
Verified, Devnet 038404b.

The load logs are working fine now.

[screen-capture (9).webm](https://github.com/user-attachments/assets/1e7bec11-1c06-476c-85f4-a2116299f5b1)

[TC2621 - Debug Logs Tab](https://app.testlodge.com/a/26076/projects/40893/suites/234374?expand_section=426660#case_4216871)

PR #3506, opened 2024-10-07
remove console DB error and loadlogs constant loading 

amiraabouhadid:
### Changes

- added checker of connectDB first before writing or reading from DB

### Related Issues

#3408

### Tested Scenarios

- click load logs directly after application mount
- deploy after application mount and observe logs get updated correctly 
- note that DB write before read console error is gone


samaradel (approved):

0oM4R (approved):
image
tested no errors and i tried to load more logs and works fine 
thanks @amiraabouhadid , great job!

Issue #3471, opened 2024-09-25
Reload you farms table after create new farm


0oM4R:
### Which package/s are you suggesting this feature for?

Dashboard

### Is your feature request related to a problem? Please describe

as in contracts page, any operation on the contracts will reload the contract after a while, 30 seconds
but in create farm you need to reload the page manually 

### Describe the solution you'd like

add timer to reload the farmss after a while

0oM4R:
make the table reload after 30 sec as in contracts table

amiraabouhadid:
Added emit from create user to update userFarms table.

khaledyoussef24:
Version fccc6af

verified on devnet

![Image](https://github.com/user-attachments/assets/0b43c3d6-0ca3-4424-90ca-4b0b97d0ffc0)
![Image](https://github.com/user-attachments/assets/a24e46c5-e6fc-47eb-b15b-85a1e0e807b6)
[testcase](https://app.testlodge.com/a/26076/projects/40893/suites/234374?expand_section=426668#case_4150414)

khaledyoussef24:
it reloads after the first farm creation but after the second farm creation it does not reload
![Screenshot 2024-10-20 at 11 22 29 AM (2)](https://github.com/user-attachments/assets/f2264084-c09a-446e-b58d-c76a696150d9)
![Screenshot 2024-10-20 at 11 22 40 AM](https://github.com/user-attachments/assets/9e4124e4-863c-415c-bc08-2cfd8809e269)
![Screenshot 2024-10-20 at 11 22 56 AM (2)](https://github.com/user-attachments/assets/efe3bb50-cc75-44c7-a97a-2546d30aca30)

A-Harby:
Verified, Devnet c0d9939.

The farm table reloads every time a new farm is created, even after 3rd one now.
![image](https://github.com/user-attachments/assets/036171ee-06f7-4416-bfeb-be984a381558)
![image](https://github.com/user-attachments/assets/99ac417e-4779-4cf8-9b98-ea747160139e)

Update test case: [TC2205 - Create a Farm](https://app.testlodge.com/a/26076/projects/40893/suites/234374?expand_section=426668#case_4150414)

PR #3515, opened 2024-10-08
Add your farms table reload on farm create

amiraabouhadid:
### Changes

- add emit event farm-created from createfarm after farm is created
- add listener in parent to receive event pass it down to userfarms
- add prop in userfarms to update on user create

### Related Issues

#3471

### Tested Scenarios

- created a farm, watched user farms table get reloaded after 30 seconds

### DIFF
--- a/packages/playground/src/dashboard/components/create_farm.vue
+++ b/packages/playground/src/dashboard/components/create_farm.vue
@@ -71,18 +71,20 @@ export default {
       required: true,
     },
   },
-  setup(props) {
+
+  setup(props, context) {
     const showDialogue = ref(false);
     const isCreating = ref(false);
     const gridStore = useGrid();
     const valid = ref(false);
+
     async function createFarm() {
       try {
         isCreating.value = true;
         await gridStore.grid.farms.create({ name: props.name });
         createCustomToast("Farm created successfully.", ToastType.success);
-        notifyDelaying();
         showDialogue.value = false;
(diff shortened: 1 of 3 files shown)


0oM4R (commented):
it load actually but it not visualized, i can't see the loading bar;
and when tried to test it on slow network, it never got reloaded

amiraabouhadid:
@0oM4R  works, just wait 30 seconds after creating farm on regular network
https://www.loom.com/share/b1084b6950e749d890875ea0aebe51f0?sid=3dcf18c6-27b6-45e5-a734-30cc36ff9587

packages/playground/src/dashboard/components/create_farm.vue
```diff
@@ -71,18 +71,20 @@ export default {
...
     const isCreating = ref(false);
     const gridStore = useGrid();
     const valid = ref(false);
+
     async function createFarm() {
       try {
         isCreating.value = true;
         await gridStore.grid.farms.create({ name: props.name });
         createCustomToast("Farm created successfully.", ToastType.success);
-        notifyDelaying();
```
0oM4R, line 84: why do you remove it? i think it will be better to notify the user that it will reload in a while

amiraabouhadid: added back

0oM4R (approved):

samaradel (approved):

Issue #3520, opened 2024-10-09
🐞 [Bug]: Domains page is broken when the profile isn't active


samaradel:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### which package/s did you face the problem with?

Dashboard

### What happened?

The domains page works when the profile is not active. Does the domains page require login??

### Steps To Reproduce

- Navigate to the Dashboard.
- Close the profile manager without logging in.
- The sidebar is disabled for the pages that require login and the user shouldn't be able to access any deployment page without logging in.

### which network/s did you face the problem on?

Dev

### version

77e194

### Relevant screenshots/screen records

![image](https://github.com/user-attachments/assets/f71fdd17-3bb1-43a3-a9a9-1f057ea8eaa6)

xmonader:
It should have the same behavior as the other pages e.g Virtual Machines and Orchestrators  

![image](https://github.com/user-attachments/assets/9774792a-0331-43ae-bc2c-1b17cfc6ef79)

samaradel:
### Work Completed:
- Fix the Domains page view on the before-login view
- Check if there client in gridstore

khaledyoussef24:
verified on devent
Version d757d29

![Image](https://github.com/user-attachments/assets/0eb98faf-4b2a-4d51-ad71-258a0058cb04)

[testcase](https://app.testlodge.com/a/26076/projects/40893/suites/234374?expand_section=442285#case_4150325)

PR #3532, opened 2024-10-16
Fix the Domains page view on the before-login view

samaradel:
### Description

Fix the Domains page view on the before-login view
### Changes

[Screencast from 16-10-24 14:45:32.webm](https://github.com/user-attachments/assets/88b5fe45-6fbe-4292-8ea1-92ae5c5ec05c)

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/3520

### Tested Scenarios

- Try to reach the domains page before logging in.
- After login go to the domains page and logout from there.

### Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation

### DIFF
--- a/packages/playground/src/components/node_selector/TfDomainName.vue
+++ b/packages/playground/src/components/node_selector/TfDomainName.vue
@@ -147,7 +147,6 @@ export default {
       size: window.env.PAGE_SIZE,
       page: Math.max(1, pagination.value.page),
       farmId: enableCustomDomain.value ? props.farm?.farmId : undefined,
-      availableFor: gridStore.client.twinId,
     }));
     const selectedDomain = ref<NodeInfo | null>(null);
     const loadDomains = () => domainsTask.value.run(gridStore, filters.value);
--- a/packages/playground/src/components/node_selector/TfDomainName.vue
+++ b/packages/playground/src/components/node_selector/TfDomainName.vue
@@ -147,6 +147,7 @@ export default {
       size: window.env.PAGE_SIZE,
       page: Math.max(1, pagination.value.page),
       farmId: enableCustomDomain.value ? props.farm?.farmId : undefined,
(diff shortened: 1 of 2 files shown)


0oM4R (approved):

packages/playground/src/components/node_selector/TfDomainName.vue
```diff
@@ -147,7 +147,6 @@ export default {
       size: window.env.PAGE_SIZE,
       page: Math.max(1, pagination.value.page),
       farmId: enableCustomDomain.value ? props.farm?.farmId : undefined,
-      availableFor: gridStore.client.twinId,
```
0oM4R, line 150: why did you remove this ?

0oM4R (changes requested):

packages/playground/src/components/node_selector/TfDomainName.vue
samaradel: It's seeking for the twinId when you logout from the dashboard when you are already on the domains page and the page blocks on that error so you have to refresh the page to apply the logout method.

0oM4R: but i think this will break the filter logic!, available for is needed while listing the nodes in the domains

samaradel: maybe it needs to be optional to avoid breaking on logout, let me check

amiraabouhadid (changes requested):
breaks when you do the following:
- log in
- go to domains page
- press log out button
- look at console error below

image

amiraabouhadid (approved):
my bad was testing wrong branch lgtm

maayarosama (approved):

0oM4R (approved):

PR #3544, opened 2024-10-21
Update farm reload after farm creation

amiraabouhadid:
### Changes

- added previously removed event emitter in create farm
- added prop change listener and changer in parent that emits reload farms in child

### Related Issues

#3471

### Tested Scenarios

- created a farm, waited for reload then created another farm and waited for reload
- need to wait 30seconds after farm creation for reload
- tested it after page refresh

### Checklist

https://www.loom.com/share/1a0590819e504839b53fb95f59b4a758?sid=ef2a05d8-fd91-4688-b000-e80dd8263b0e

### DIFF
--- a/packages/playground/src/dashboard/components/create_farm.vue
+++ b/packages/playground/src/dashboard/components/create_farm.vue
@@ -59,7 +59,7 @@ import { createCustomToast, ToastType } from "../../utils/custom_toast";
 
 export default {
   name: "CreateFarm",
-  setup() {
+  setup(_, context) {
     const showDialogue = ref(false);
     const isCreating = ref(false);
     const gridStore = useGrid();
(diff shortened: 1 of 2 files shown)


maayarosama (approved):

samaradel:
So, there's no reload bar? I think we still need it, as the user doesn't know how much time he should wait.
@AhmedHanafy725 what do you think ?

amiraabouhadid:
> So, there's no reload bar? I think we still need it, as the user doesn't know how much time he should wait. @AhmedHanafy725 what do you think ?

if we add a reload bar it wont be consistent with the behavior of other tables in the dashboard. so best to leave it as it is.

samaradel (approved):

Issue #2364, opened 2024-03-06
Validation error when listing deployments in the dashboard from other clients


A-Harby:
### Description

A clear description of the bug is:

Validation error from deployments made outside of the TS client, either go or terraform (name validation is different from go compared with the TS).
![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/78362868-8f01-4edd-82ac-db6f5ffb357f)

Must include any relevant identifiers like:

- **Network**: Devnet

- **Version**: ea66380

### Steps to reproduce

- Deploy a VM or K8s from a Go client, CLI, TFRobot, or even Terraform.
- Navigate to the dashboard.
- Connect your wallet with the same account.
- Navigate to the deployment page and check the logs.

ramezsaeed:
@A-Harby 
this is only for logs, or the deployment is not shown if the validation is different ?

A-Harby:
No, The deployments were listed, but I was only testing listing from other clients so maybe it could cause a problem with adding a domain in vms or works in the kubernetes from ts client

Mahmoud-Emad:
What is the status here?

amiraabouhadid:
Sorry for the delay took some time to do investigative work done, was able to duplicate the error. Identifying source of error in client.

amiraabouhadid:
Might be caused by inconsistent type casting in client.

amiraabouhadid:
error is due to missing validations for number of characters in name in go client which is inconsistent with the validation implemented in the ts client.  @A-Harby please list any other missing validations in go client. Please also let us know if validations are missing in terraform

A-Harby:
The naming validation is different for every client (TS, GO, and Terraform). Only the TS client has an alphanumeric validation naming for any deployment name, disk name, and network name with a minimum and maximum length.

while the terraform doesn't have any length validation and can accept underscores only beside alphanumeric, and the same goes for the Go client so far(there could be a difference in some cases).

![image](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/109473873/26a6da90-2732-4496-81ca-716aa677a99c)

amiraabouhadid:
opened an issue for it in the go sdk [here ](https://github.com/threefoldtech/tfgrid-sdk-go/issues/1055)

AhmedHanafy725:
check https://github.com/threefoldtech/tfgrid-sdk-go/issues/1055#issuecomment-2419558915

[...]

PR #3549, opened 2024-10-23
update max name length in solutions

amiraabouhadid:
### Changes

- changed max name length in deployment model in client
- changed max name length in all deployment dashboard validation
- in solutions that require domain, max name length was changed to 15 chars
- in all other solutions, max name length was changed to 35 chars
### Related Issues

#2364

### Tested Scenarios

- entered a name longer than 35 chars in dashboard deployments

image

### DIFF
--- a/packages/playground/src/weblets/tf_domains.vue
+++ b/packages/playground/src/weblets/tf_domains.vue
@@ -1,7 +1,9 @@
 <template>
   <weblet-layout ref="layout">
-    
-    <template #title><span><v-icon  class="pr-3">mdi-web-box</v-icon></span>Deploy Domains Instance </template>
+    <template #title
+      ><span><v-icon class="pr-3">mdi-web-box</v-icon></span
+      >Deploy Domains Instance
+    </template>
 
     <d-tabs :tabs="[{ title: 'Config', value: 'config' }]">
       <template #config>
@@ -14,7 +16,7 @@
               validators.isAlphanumeric('Subdomain should consist of letters and numbers only.'),
               subdomain => validators.isAlpha('Subdomain must start with alphabet char.')(subdomain[0]),
               validators.minLength('Subdomain must be at least 4 characters.', 4),
-              subdomain => validators.maxLength('Subdomain cannot exceed 50 characters.', 50)(subdomain),
+              subdomain => validators.maxLength('Subdomain cannot exceed 35 characters.', 35)(subdomain),
             ]"
             :async-rules="[validateSubdomain]"
             #="{ props }"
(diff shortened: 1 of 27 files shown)


samaradel (approved):

0oM4R (commented):
image

is there a reason for gitea to be only 15 char ?

amiraabouhadid:
> image
> 
> is there a reason for gitea to be only 15 char ?

good question, for node pilot too @zaelgohary ?

zaelgohary:
> 
> > is there a reason for gitea to be only 15 char ?
> 
> 
> 
> good question, for node pilot too @zaelgohary ?

I'm not sure if node pilot has another restriction rule on its own but you may try to deploy and check if it works.

amiraabouhadid:
@0oM4R  changing gitea and node pilot name max length to 35 and deploying throws this error 

image

image

so won't be changing gitea and node pilot max length from 15 to 35

0oM4R:
i think we need to handle gatewaynameModal length, i see it got changed [here](https://github.com/threefoldtech/tfgrid-sdk-ts/commit/c5f1bad1c09934f276e88c53980a82e17b3ce3fa#diff-13f46ac7ceefb9f9d1ca1a680d587fccdb952ec2a5c94eea64c2cce762236927R237) 

it is happening with matermost

0oM4R (changes requested):
i think we need to handle gatewaynameModal length, i see it got changed [here](https://github.com/threefoldtech/tfgrid-sdk-ts/commit/c5f1bad1c09934f276e88c53980a82e17b3ce3fa#diff-13f46ac7ceefb9f9d1ca1a680d587fccdb952ec2a5c94eea64c2cce762236927R237) 

it is happening with matermost 
Screenshot from 2024-10-29 13-32-50

amiraabouhadid:
@0oM4R good catch, got validation error for more than 36 chars indeed
image
it has been set to 35 char max now

0oM4R (commented):
i think we have to update gatewaynameModel to accept the namelength + length of the domain @AhmedHanafy725  can you please provide numbers here ? 

this was already exists but I don't know why it got removed
after manage that please update gitea and node pilot domain names

packages/playground/src/weblets/tf_domains.vue
0oM4R, line 1: no need to be 15 here as the name here is the gateway name lets make it 35

image

amiraabouhadid: done 
image

0oM4R (approved):

samaradel (approved):

Issue #3509, opened 2024-10-07
🐞 [Bug]: Can't use `add_worker` in the same script as deploy k8s with Mycelium option is added


A-Harby:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### which package/s did you face the problem with?

grid_client

### What happened?

I can deploy a K8s master and worker with Mycelium, but it will fail if I add a worker in the same script with the `add_worker` function and then run them together. 

### Steps To Reproduce

If you want to try, you can use the k8s test file; just change all the Mycelium options to turn true and run it; it will fail. 

### which network/s did you face the problem on?

Dev

### version

.

### Relevant screenshots/screen records

![image](https://github.com/user-attachments/assets/ee100699-1876-467e-8c70-bc593c3e2986)

### Relevant log output

```console
ValidationError: Another mycelium seed is used for this network vh71o5bpbu9q34g on this 167
```

samaradel:
Work in Progress (WIP):
Checking ```myceliumSeeds``` and how script works

samaradel:
### Work Completed:
- Check if ```myceliumNetworkSeed.seed``` is existed.
- Remove ```.skip``` from ```add worker``` test case.

atefnazmy:
- A similar issue happens when adding a worker via the Dashboard

![add_worker](https://github.com/user-attachments/assets/7e6bfd27-4fc0-4790-b827-3a9b053ea4cb)

samaradel:
Blocked on [3680](https://github.com/threefoldtech/tfgrid-sdk-ts/issues/3680) as now I got the same error 
![image](https://github.com/user-attachments/assets/5b52e472-f723-4a3c-a592-ef2b893c9540)

0oM4R:
![image](https://github.com/user-attachments/assets/996ace80-bbd7-4529-b498-44e8ef9ac267)
faced this issue again on dashboard

samaradel:
it's blocked on https://github.com/threefoldtech/tfgrid-sdk-ts/issues/3509

khaledyoussef24:
the deployment works fine but the test keeps failing 
may be we shall check the validation or assertions 
<img width="865" alt="Screenshot 2024-12-30 at 8 50 01 AM" src="https://github.com/user-attachments/assets/3f0a08a8-97e6-42a1-a6d3-b2bf86bd0935" />
[...]

PR #3565, opened 2024-10-30
Fix adding worker with mycelium to k8s

samaradel:
### Description

- Check if ```myceliumNetworkSeed.seed``` is existed.
- Remove ```.skip``` from ```add worker``` test case.

### Changes

image

### Tested Scenarios

- Go to the Grid Client package
- Run `yarn test tests/modules/kubernetes.test.ts`
- All cases should pass

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/3509

### Checklist

- [x] Tests included
- [x] Build pass
- [ ] Documentation

### DIFF
--- a/packages/grid_client/src/primitives/network.ts
+++ b/packages/grid_client/src/primitives/network.ts
@@ -176,7 +176,7 @@ class Network {
       return network["node_id"] === nodeId;
     });
     const myceliumNetworkSeed = myceliumSeeds.find(item => item.nodeId == nodeId);
-    if (network && network.mycelium && network.mycelium?.hex_key) {
+    if (network && network.mycelium && network.mycelium?.hex_key && myceliumNetworkSeed?.seed) {
       if (myceliumSeeds && myceliumSeeds.length > 0 && myceliumNetworkSeed?.seed !== network.mycelium.hex_key) {
         throw new ValidationError(`Another mycelium seed is used for this network ${this.name} on this ${nodeId}`);
       }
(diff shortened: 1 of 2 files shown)


0oM4R (changes requested):
please update the PR description to include tested scenarios.
and brief  about the original couase of the issue will be great

ramezsaeed (approved):

0oM4R (commented):
tests are still failing 
Screenshot from 2024-11-21 13-19-42

samaradel:
image
works now

maayarosama:
I get the following error while trying to run the test, I suggest that it might be a good idea to to read the network, mnemonics and ssh key straight from the config file 
Screenshot from 2024-12-05 14-57-52

samaradel:
> I get the following error while trying to run the test, I suggest that it might be a good idea to to read the network, mnemonics and ssh key straight from the config file Screenshot from 2024-12-05 14-57-52

I'm not sure if we should do that, I see it implemented that way in many scripts.. @AhmedHanafy725 what do you think?

AhmedHanafy725:
> > I get the following error while trying to run the test, I suggest that it might be a good idea to to read the network, mnemonics and ssh key straight from the config file Screenshot from 2024-12-05 14-57-52
> 
> I'm not sure if we should do that, I see it implemented that way in many scripts.. @AhmedHanafy725 what do you think?

it's required for the ssh package, but it should be mentioned clearly in the docs that we should generate an ed25519 sshkey in the correct path. Please create an issue for it

ramezsaeed:
https://github.com/threefoldtech/tfgrid-sdk-ts/pull/3565#issuecomment-2525673735

its mentioned in the tests here to use ed25519 sshkey:

- https://github.com/threefoldtech/tfgrid-sdk-ts/blob/development/packages/grid_client/tests/README.md
- https://github.com/threefoldtech/tfgrid-sdk-ts/blob/development/packages/grid_client/docs/tests.md

Issue #3523, opened 2024-10-13
🐞 [Bug]: zmount tests are failing on the grid tests workflow


A-Harby:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### which package/s did you face the problem with?

Dashboard

### What happened?

The zmount tests are keep failing.

### which network/s did you face the problem on?

Dev

### version

dev branch

### Relevant screenshots/screen records

https://github.com/threefoldtech/tfgrid-sdk-ts/runs/31240774792#r0s30

### Relevant log output

```console
.
```

0oM4R:
@ramezsaeed please check if this test related issue or not.

A-Harby:
> @ramezsaeed please check if this test related issue or not.

I checked the tests, and all its doing is just calling the zmount module function and asserting the result, so the changes are coming from the module, not the tests.

samaradel:
### Work in Progress (WIP):
checking jest return values

samaradel:
### Work Completed:
Fix ```zmount``` test cases expected errors.

samaradel:
### Work Completed:
Refactor test cases

samaradel:
### Work Completed:

Add function to get random number to validate z-mount size

samaradel:
### Work Completed:
Add random ranges to check the min, max, and valid sizes.

khaledyoussef24:
verified and tests are now passing and working  fine
<img width="740" alt="Image" src="https://github.com/user-attachments/assets/cf54b1a6-ac24-4b71-b8bb-088843341ef0" />
[testcase](https://app.testlodge.com/a/26076/projects/40893/suites/194096?expand_section=366955#case_4431580)

PR #3608, opened 2024-11-10
Fix zmount test cases

samaradel:
### Description

Fix zmount test suite expected errors.

### Changes
image

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/3523

### Tested Scenarios

- Update the value of size to get errors

### Checklist

- [x] Tests included
- [x] Build pass
- [ ] Documentation


packages/grid_client/tests/modules/zmount.test.ts
```diff
@@ -15,29 +15,43 @@ describe("Zmount module", () => {
 
     zmount.size = size;
 
-    const result = () => zmount.challenge();
+    const result = zmount.challenge();
 
-    expect(result).toThrow();
+    expect(result).toBe(size.toString());
```
xmonader: can directly do expect(zmount.challenge()).toBe....

packages/grid_client/tests/modules/zmount.test.ts
```diff
@@ -15,29 +15,43 @@ describe("Zmount module", () => {
...
   });
 
   test("Max value for size.", () => {
     const size = 100 * 1025 ** 4;
```
xmonader: What do these numbers mean? try extracting to a const

packages/grid_client/tests/modules/zmount.test.ts
```diff
@@ -15,29 +15,43 @@ describe("Zmount module", () => {
...
+      expect.objectContaining({
+        constraints: {
+          max: "size must not be greater than 10995116277760",
+        },
+      }),
+    );
   });
 
   test("Size doesn't accept decimal value.", () => {
     const size = 1.5;
```
xmonader: is that 1.5 byte?

packages/grid_client/tests/modules/zmount.test.ts
```diff
@@ -15,29 +15,43 @@ describe("Zmount module", () => {
...
+      expect.objectContaining({
+        constraints: {
+          isInt: "size must be an integer number",
+          min: "size must not be less than 104857600",
+        },
+      }),
+    );
   });
 
   test("Size empty value.", () => {
```
xmonader: While we are at it, check against null, undefined, NaN, {} , and []

packages/grid_client/tests/modules/zmount.test.ts
```diff
@@ -15,29 +15,43 @@ describe("Zmount module", () => {
...
     const result = () => zmount.challenge();
 
-    expect(result).toThrow();
+    expect(() => {
+      zmount.size = size;
+      result;
+    }).toThrow(
+      expect.objectContaining({
+        constraints: {
+          max: "size must not be greater than 10995116277760",
```
xmonader: check against the boundaries max, and min

and off by 1 min - 1 and max+1

xmonader (changes requested):
We need to use better descriptions to describe the cases better

0oM4R (approved):
passed locally, not sure about the tests description 
image

packages/grid_client/tests/modules/zmount.test.ts
```diff
@@ -1,58 +1,39 @@
...
+  beforeEach(() => {
+    zmount = new Zmount();
   });
 
-  test("Size doesn't accept decimal value.", () => {
-    const size = 1.5;
-
-    zmount.size = size;
+  describe("size property validation", () => {
+    it("should fail validation if size is less than the minimum", async () => {
```
amiraabouhadid, line 36: shouldn't we have a test to test if input size is int first? before testing min and max

packages/grid_client/tests/modules/zmount.test.ts
```diff
@@ -1,58 +1,39 @@
...
-    zmount.size = negative_size;
-
-    const result = () => zmount.challenge();
+  describe("challenge method", () => {
+    it("should return the size as a string", () => {
+      zmount.size = 5 * 1024 ** 3;
 
-    expect(result).toThrow();
+      const result = zmount.challenge();
+      expect(result).toBe("5368709120");
```
amiraabouhadid: without checking if size is int , the challenge can return any special chars in the string, no?

packages/grid_client/tests/modules/zmount.test.ts
```diff
@@ -1,58 +1,39 @@
...
+  describe("size property validation", () => {
+    it("should fail validation if size is less than the minimum", async () => {
+      // Less than 100 MB
+      expect(() => (zmount.size = 50 * 1024 ** 2)).toThrow();
+    });
 
-    const result = () => zmount.challenge();
+    it("should fail validation if size is greater than the maximum", async () => {
+      // Greater than 10 TB
+      expect(() => (zmount.size = 15 * 1024 ** 4)).toThrow();
```
amiraabouhadid: define the min and max as constants before any of the tests

ramezsaeed (approved):

amiraabouhadid:
please place the min and max values in a constant before any of the tests. Please move tests that check if size is null, Nan or not an int before the min and max tests

maayarosama (approved):

amiraabouhadid (approved):

packages/grid_client/tests/modules/zmount.test.ts
```diff
@@ -1,58 +1,58 @@
+import { validate } from "class-validator";
+
 import { Zmount } from "../../src";
 
-let zmount: Zmount;
+describe("Zmount Class", () => {
+  let zmount: Zmount;
+  const minSize = 50 * 1024 ** 2; // Less than 100 MB
+  const maxSize = 15 * 1024 ** 4; // Greater than 10 TB
+  const validSize = 5 * 1024 ** 3;
```
AhmedHanafy725: it's better to have a random size (within a certain range) to test with

you can use this method to generate the random number https://github.com/threefoldtech/tfgrid-sdk-ts/blob/f9a5b06796073bcd61ef729240308d122beb3646/packages/grid_client/src/helpers/utils.ts#L18

AhmedHanafy725 (changes requested):
you should add a test for setting a floating number

ramezsaeed (approved):

packages/grid_client/tests/modules/zmount.test.ts
```diff
@@ -1,58 +1,60 @@
...
-    zmount.size = size;
+import { Zmount } from "../../src";
+import { getRandomNumber } from "../../src/helpers/utils";
 
-    const result = () => zmount.challenge();
+describe("Zmount Class", () => {
+  let zmount: Zmount;
+  const minSize = 50 * 1024 ** 2; // Less than 100 MB
+  const maxSize = 15 * 1024 ** 4; // Greater than 10 TB
+  const validSize = 5 * 1024 ** 3;
```
maayarosama: Why are we initializing this with a value, if we're using getRandomNumber

packages/grid_client/tests/modules/zmount.test.ts
```diff
@@ -1,58 +1,59 @@
...
-    zmount.size = negative_size;
-
-    const result = () => zmount.challenge();
+  describe("challenge method", () => {
+    it("should return the size as a string", () => {
+      zmount.size = validSize;
 
-    expect(result).toThrow();
+      const result = zmount.challenge();
+      expect(typeof result).toBe("string");
```
AhmedHanafy725: you can cast `validSize` to string and check against its value

AhmedHanafy725 (approved):

maayarosama (approved):

PR #3646, opened 2024-11-17
Handle rented node in grace period

samaradel:
### Description

Check if the rented node contract is in the grace period when listing the node.

### Changes

image

### Related Issues

### Tested Scenarios

- Navigate to the Node finder page.
- Rent a dedicated node.
- Leave it till it goes to the grace period.
- Navigate to any solution.
- Try to deploy an instance on the rented node.

### Checklist

- [x] Tests included
- [x] Build pass
- [ ] Documentation

### DIFF
--- a/packages/playground/src/components/node_selector/TfAutoNodeSelector.vue
+++ b/packages/playground/src/components/node_selector/TfAutoNodeSelector.vue
@@ -307,10 +307,7 @@ export default {
               id: node?.rentContractId,
             });
             if (state.gracePeriod) {
-              createCustomToast(
-                `You can't deploy on node ${node.nodeId}, its rent contract is in grace period.`,
-                ToastType.danger,
-              );
+              throw `You can't deploy on node ${node.nodeId}, its rent contract is in grace period.`;
             }
           }
         } catch (error) {
--- a/packages/playground/src/components/node_selector/TfAutoNodeSelector.vue
+++ b/packages/playground/src/components/node_selector/TfAutoNodeSelector.vue
@@ -151,6 +151,9 @@ import { computed, nextTick, onMounted, onUnmounted, type PropType, ref } from "
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
 import type { VCard } from "vuetify/components/VCard";
 
+import { createCustomToast, ToastType } from "@/utils/custom_toast";
+import { normalizeError } from "@/utils/helpers";
+
 import { useAsync, usePagination, useWatchDeep } from "../../hooks";
 import { ValidatorStatus } from "../../hooks/form_validator";
 import { useGrid } from "../../stores";
@@ -167,7 +170,6 @@ import {
   validateRentContract,
 } from "../../utils/nodeSelector";
 import TfNodeDetailsCard from "./TfNodeDetailsCard.vue";
-
 export default {
   name: "TfAutoNodeSelector",
   components: { TfNodeDetailsCard },
@@ -299,14 +301,26 @@ export default {
 
     const nodeInputValidateTask = useAsync<boolean, string, [NodeInfo | undefined]>(
       async node => {
-        if (node && node?.rentContractId !== 0) {
-          const { state } = await gridStore.grid.contracts.get({
-            id: node?.rentContractId,
-          });
-          if (state.gracePeriod) {
-            return false;
+        try {
+          if (node && node.rentContractId !== 0) {
+            const { state } = await gridStore.grid.contracts.get({
+              id: node?.rentContractId,
+            });
+            if (state.gracePeriod) {
+              createCustomToast(
+                `You can't deploy on node ${node.nodeId}, its rent contract is in grace period.`,
+                ToastType.danger,
+              );
+            }
           }
+        } catch (error) {
+          const err = normalizeError(
+            error,
+            "Something went wrong while checking status of the node. Please check your connection and try again.",
+          );
+          throw err;
         }
+
         const nodeCapacityValid = await checkNodeCapacityPool(gridStore, node, props.filters);
         const rentContractValid = props.filters.dedicated ? await validateRentContract(gridStore, node) : true;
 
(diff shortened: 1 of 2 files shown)


0oM4R (commented):
image


i see two toasts appears, should be only on and should be inside the node finder

0oM4R (commented):
still having toast 

[Screencast from 2024-11-20 16-27-08.webm](https://github.com/user-attachments/assets/6b72b11b-4781-420b-b1bf-20993a39cd84)

also the inside node selector the error is shown as info in  first

ramezsaeed (approved):

packages/playground/src/components/node_selector/TfAutoNodeSelector.vue
```diff
@@ -151,6 +151,9 @@ import { computed, nextTick, onMounted, onUnmounted, type PropType, ref } from "
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
 import type { VCard } from "vuetify/components/VCard";
 
+import { createCustomToast, ToastType } from "@/utils/custom_toast";
```
0oM4R: are those used anywhere?

packages/playground/src/components/node_selector/TfAutoNodeSelector.vue
```diff
@@ -299,14 +301,23 @@ export default {
...
-          });
-          if (state.gracePeriod) {
-            return false;
+        try {
+          if (node && node.rentContractId !== 0) {
+            const { state } = await gridStore.grid.contracts.get({
+              id: node?.rentContractId,
+            });
+            if (state.gracePeriod) {
+              throw `You can't deploy on node ${node.nodeId}, its rent contract is in grace period.`;
```
0oM4R: i suggest adding time out before throwing the error, for more friendly UX.

packages/playground/src/utils/nodeSelector.ts
```diff
@@ -229,10 +229,7 @@ export async function validateRentContract(
         id: node.rentContractId,
       });
       if (contractInfo.state.gracePeriod) {
-        createCustomToast(
-          `You can't deploy on node ${node.nodeId}, its rent contract is in grace period.`,
-          ToastType.danger,
-        );
+        throw `You can't deploy on node ${node.nodeId}, its rent contract is in grace period.`;
```
0oM4R: i don't think we should throw an error here, we should update the field message. 
image

0oM4R: image

packages/playground/src/components/node_selector/TfAutoNodeSelector.vue
samaradel: checked the associated PRs and I see it's not anymore.

packages/playground/src/components/node_selector/TfAutoNodeSelector.vue
samaradel: why do we need timeout here? we check the state object only.

packages/playground/src/utils/nodeSelector.ts
samaradel: I don't think so as it's doing the double toast issue and it all should take the same behavior (throwing an error or firing a toast)

packages/playground/src/components/node_selector/TfAutoNodeSelector.vue
0oM4R: https://github.com/threefoldtech/tfgrid-sdk-ts/pull/3646#pullrequestreview-2448787416
in this video, the switching between info and error toast is to fast, if you have any other solutoin to hide the info toast then show the error will be much better imo

[Screencast from 2024-12-01 11-27-36.webm](https://github.com/user-attachments/assets/25278603-d7a4-4848-9299-9cf5b2d7d6d0)

packages/playground/src/utils/nodeSelector.ts
0oM4R: Sorry, I disagree. I don't see any validation error in the manual selector that appears on toast, all errors are shown in the field info, As you said, we need to make the behavior consistent. @AhmedHanafy725  what do you think?

AhmedHanafy725: It should be consistent. The error shouldn't be in the toast, it should be a field error

packages/playground/src/components/node_selector/TfAutoNodeSelector.vue
samaradel: fixed

amiraabouhadid:
no toast error, only console err
image
not sure if resources issue is from this pr changes or in dashboard overall
field error takes a while until it shows
image

packages/playground/src/components/node_selector/TfAutoNodeSelector.vue
```diff
@@ -299,14 +298,27 @@ export default {
...
-            return false;
+        try {
+          if (node && node.rentContractId !== 0) {
+            const { state } = await gridStore.grid.contracts.get({
+              id: node?.rentContractId,
+            });
+            if (state.gracePeriod) {
+              await new Promise((_, reject) => {
+                setTimeout(() => {
+                  reject(Error(`You can't deploy on node ${node.nodeId}, its rent contract is in grace period.`));
```
amiraabouhadid, line 309: shouldn't we use console.err?

packages/playground/src/components/node_selector/TfAutoNodeSelector.vue
```diff
@@ -299,14 +298,27 @@ export default {
...
+                }, 2000);
+              });
+            }
           }
+        } catch (error) {
+          const err = normalizeError(
```
amiraabouhadid, line 315: shouldnt we use console.err instead ?

samaradel: Instead? I don't think so. It's a connection error, which the user should be aware of in toast.

0oM4R (commented):
on selecting the node with the automatic node selector, i got this error in console

image

Issue #3575, opened 2024-11-03
🐞 [Bug]: Unreserving dedicated node from solution node selector not reflected


samaradel:
### which package/s did you face the problem with?

Dashboard

### What happened?

Unreserving the dedicated node from the solution node selector requires a refresh to reflect.

### Steps To Reproduce

- Go to any solution.
- switch dedicated nodes to view the reserved nodes.
- click on unreserve button.

### which network/s did you face the problem on?

Dev

### version

b8181e

### Relevant screenshots/screen records

[Screencast from 03-11-24 18:28:23.webm](https://github.com/user-attachments/assets/4b36e276-c750-4c86-a40a-c6738c4fa1b0)

### Relevant log output

```console
N/A
```

ramezsaeed:
Update the button to fitch the status from grid proxy and update the button to show reserve iso unreserv.

samaradel:
### Work Completed:
Fix the reserve action button.

samaradel:
### Work Completed:
- Change founding active contract on node alert msg to error 
- Fix reserving node button loading

samaradel:
#blocked on https://github.com/threefoldtech/tfgrid-sdk-ts/pull/3587

amiraabouhadid:
faced when reviewing #3748
https://www.loom.com/share/9fc9a7898e294466af53da5bfc33e364?sid=b346b48f-10b2-4c6a-ab04-93b3d6785a00

samaradel:
> faced when reviewing [#3748](https://github.com/threefoldtech/tfgrid-sdk-ts/pull/3748) https://www.loom.com/share/9fc9a7898e294466af53da5bfc33e364?sid=b346b48f-10b2-4c6a-ab04-93b3d6785a00

it takes from 17 to 20 seconds to reflect

samaradel:
there's no dedicated node to check

samaradel:
### Work Completed: 
Fix action btn after reserve

0oM4R:
Work completed:
in grid client node id added to FilterOptions to be able to filter nodes based on its id.
refactor the current reserve and unreserve post actions to update the node status and emit event to update it in its parents
 
pr merged

atefnazmy:
- Verified on Dev,
- Version: bdf1c49
- The solution node selector Unreseve button works as expected and reflects correctly.

https://github.com/user-attachments/assets/4828bd61-f140-4e54-a384-5e39b41f3592

[...]

PR #3587, opened 2024-11-04
Add rentableOrRentedBy query in dedicated filter

samaradel:
### Description

- Add ```rentableOrRentedBy``` query in the dedicated filter.
- merging development branch.

### Changes

[Screencast from 04-11-24 16:30:07.webm](https://github.com/user-attachments/assets/c5001aec-fd6d-4afc-87bd-0191c51fab53)

### Tested Scenarios

- Go to any solution page
- Toggle on the Dedicated node switch
- Load Nodes and you'll find the nodes you already rented and the nodes that are available for you to rent

### Scenarios Update

- Rent a node and make sure it's not listed in the free dedicated nodes filter.
- Unreserve the node and ensure it is back in the free dedicated node filter.
- Try to deploy on a node not rented by you (should fail with proper error message)

### Documentation PR
- https://github.com/threefoldtech/info_grid/issues/705

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/2775

### Checklist

- [x] Tests included
- [x] Build pass
- [x] Documentation
- [x] Code format and docstrings
- [x] Screenshots/Video attached (needed for UI changes)

PR #3650, opened 2024-11-18
Fix reserve action button

samaradel:
### Description

Fix the reserve action button.

### Changes

- Reduce check time of Un/Reserved buttons
- Check the twin ID of the rented node when the component is mounted
-  Reset values after reserve / unreserve process

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/3575

### Tested Scenarios

- Navigate to any solution
- Switch on dedicated nodes
- Reserve any node and unreserve it

### Updated Scenarios
- Reserve from the node page and un-reserve any from the solution page.
- Reserve from a solution page, then un-reserve from the node page.

### Checklist

- [x] Tests included
- [x] Build pass
- [ ] Documentation

### DIFF
--- a/packages/playground/src/dashboard/components/reserve_action_btn.vue
+++ b/packages/playground/src/dashboard/components/reserve_action_btn.vue
@@ -18,14 +18,14 @@
     </v-dialog>
     <v-btn
       :loading="loadingReserveNode"
-      v-if="node.rentedByTwinId === 0"
+      v-if="!reserved"
       :disabled="disableButton || hasInsufficientBalance"
       @click.stop="reserveNode"
       color="primary"
     >
       Reserve
     </v-btn>
-    <span v-if="node.rentedByTwinId === 0" class="ml-2">
+    <span v-if="!reserved" class="ml-2">
       <v-tooltip text="You must acquire a minimum of 2 TFTs in order to reserve any node." location="bottom">
         <template #activator="{ props }">
           <VIcon icon="mdi-information-outline" v-bind="props" />
@@ -37,7 +37,7 @@
       color="error"
       :loading="loadingUnreserveBtn"
       :disabled="disableButton"
-      v-if="node.rentedByTwinId === profile?.twinId"
+      v-if="reserved"
       @click.stop="removeReserve"
     >
       Unreserve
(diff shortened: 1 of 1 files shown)


maayarosama (approved):

ramezsaeed (commented):
- Try reserve a dedicated node from solution X, and un-reserve it from solution Y.
- Reserve from node page and un-reserve any from solution page.
- Reserve from a solution page, then un-reserve from node page.

packages/playground/src/dashboard/components/reserve_action_btn.vue
```diff
@@ -111,57 +115,59 @@ export default {
           createCustomToast(`node ${props.node.nodeId} has active contracts`, ToastType.info);
```
0oM4R: i suggest changing this to an error, not info, as the process of unreserve is failed

0oM4R: also small comment line before the request will make our life easier in the future :D, to mention that we are checking if there is contract on the rented node

0oM4R (commented):
I think we should add await on the time out to keep loading untill the val;ue got updated 

[Screencast from 2024-11-25 13-24-48.webm](https://github.com/user-attachments/assets/8260b064-b60d-4133-b62e-bfa0ec1bbc26)

ramezsaeed:
Please confirm these scenarios had been tested; and add them to the PR body.
https://github.com/threefoldtech/tfgrid-sdk-ts/pull/3650#pullrequestreview-2456695150

samaradel:
@ramezsaeed I'm waiting for https://github.com/threefoldtech/tfgrid-sdk-ts/pull/3587 as I can't see any dedicated nodes

samaradel:
@ramezsaeed the scenarios are tested except for the first case which is blocked on https://github.com/threefoldtech/tfgrid-sdk-ts/pull/3587

maayarosama:
> @ramezsaeed the scenarios are tested except for the first case which is blocked on #3587

All scenarios passed, except the first one as it's blocked on another issue, Once the its pr is merged will approve

Issue #3632, opened 2024-11-12
🐞 [Bug]: Save button not disabled after setting extra fee


zaelgohary:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### which package/s did you face the problem with?

Dashboard

### What happened?

Save button not disabled after setting extra fee.

https://app.testlodge.com/a/26076/projects/40893/runs/757631/run?selected=33242679&user_id=170529

### Steps To Reproduce

- Click on Your Farms from the sidebar then Your Farms
- In your nodes table, click on the set extra fee icon
- Enter your fee and Save
- Close the dialogue & open it again

### which network/s did you face the problem on?

Dev

### version

96e183a

### Relevant screenshots/screen records

![image](https://github.com/user-attachments/assets/028fc113-e3d3-4f2d-887c-63a68538058b)

### Relevant log output

```console
-
```

atefnazmy:
-Verified on Dev,
-Version: 4e2fc0b
-The "Save" button is now properly disabled after the modifications are successfully executed.


https://github.com/user-attachments/assets/4e4027c2-b16f-4b72-a431-61cc07c5d87a

#### Related Link:
https://app.testlodge.com/a/26076/projects/40893/runs/757631/run?selected=33242679&user_id=170529

PR #3657, opened 2024-11-19
fix save additional fees btn disable/enable

amiraabouhadid:
### Changes

- just converted fee in text field to number first before comparison with fee fetched from grid 
- fixed passing of node id to dialog to avoid requesting extra fee from grid for other nodes when you only need to fetch current node's extra fee

### Related Issues

[#3632](https://github.com/threefoldtech/tfgrid-sdk-ts/issues/3632)

### Tested Scenarios

- opened up set additional fees dialog, observe save btn is disabled
- changed fee, observed save become active when fee is different from original fee
- change back to original fee watch save btn become disabled again 

### Documentation PR

### Checklist

https://www.loom.com/share/61bc3d16573e45249a0df9499e0d3158?sid=8fff94fb-2de8-45a6-899d-c20a77b465ae

### DIFF
--- a/packages/playground/src/dashboard/components/set_extra_fee.vue
+++ b/packages/playground/src/dashboard/components/set_extra_fee.vue
@@ -24,8 +24,7 @@
             e.g. GPUs) while renting.
           </v-card-text>
           <v-card-text>
-            <VProgressLinear indeterminate color="primary" height="2" v-if="loading" />
-            <form-validator v-model="valid" v-else>
+            <form-validator v-model="valid">
               <input-validator
                 :value="inputFee"
                 :rules="[
@@ -43,11 +42,13 @@
                     suffix="USD/month"
                     outlined
                     label="Additional Fees"
-                    :disabled="isSetting"
+                    :disabled="isSetting || loading"
+                    :loading="loading"
                   ></v-text-field>
                 </input-tooltip>
               </input-validator>
             </form-validator>
+
             <v-divider></v-divider>
           </v-card-text>
           <v-card-actions class="justify-end my-1 mr-2">
@@ -102,7 +103,7 @@ export default {
       currentNodeId.value = props.nodeId;
       currentFee.value = (await getExtraFee()) ?? 0;
       inputFee.value = currentFee.value;
-      loading.value = false;
+      // loading.value = false;
     }
     async function getExtraFee() {
       try {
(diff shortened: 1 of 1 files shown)


maayarosama:
Doesn't work
[Screencast from 2024-11-19 13-28-06.webm](https://github.com/user-attachments/assets/9fc959ec-8aee-45bf-b14b-d2d51075623f)

amiraabouhadid:
> Doesn't work [Screencast from 2024-11-19 13-28-06.webm](https://github.com/user-attachments/assets/9fc959ec-8aee-45bf-b14b-d2d51075623f)

\



thanks, 
issue was actually in how node id is passed to dialogue and multiple unnecessary requests to get additional fees. should be fixed now, please check

maayarosama (approved):

ramezsaeed (commented):
I can see all test trying higher values; so lets try to change it to a lower value or zero and see whats happened in the save btn.

packages/playground/src/dashboard/components/set_extra_fee.vue
```diff
@@ -87,40 +87,39 @@ export default {
...
 
-    watch(fee, async (value, _) => {
-      const currentFee = await getExtraFee();
-      if (value == currentFee || value === null) {
-        isDisabled.value = true;
-      } else {
-        isDisabled.value = false;
-      }
+    watch(inputFee, async () => {
+      isDisabled.value = currentFee.value == inputFee.value;
```
0oM4R: I think we can replace this with computed

packages/playground/src/dashboard/components/set_extra_fee.vue
```diff
@@ -87,40 +87,39 @@ export default {
...
     const isDisabled = ref(false);
+    const currentFee = ref(0);
+    const currentNodeId = ref(0);
 
-    onMounted(async () => {
-      fee.value = (await getExtraFee()) as unknown as number;
-    });
-
+    async function setupDialog() {
+      showDialogue.value = true;
```
0oM4R, line 102: should we have loading to handle if the extra fee take some time to return?

amiraabouhadid:
> I can see all test trying higher values; so lets try to change it to a lower value or zero and see whats happened in the save btn.

tested decreasing fee actually, just wasnt included in screencast. setting to 0 works too.

0oM4R (approved):

ramezsaeed (approved):

maayarosama (approved):

Issue #3648, opened 2024-11-18
🐞 [Bug]: Google Play Button Style in Deposit Dialog is Distorted


zaelgohary:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### which package/s did you face the problem with?

Dashboard

### What happened?

![image](https://github.com/user-attachments/assets/a7d14062-07e8-4c3f-aa9d-111f9d29ca9f)

### Steps To Reproduce

1. Login
2. Go to Bridge then open deposit dialogue

### which network/s did you face the problem on?

Dev

### version

bf66d67

### Relevant screenshots/screen records

-

### Relevant log output

```console
-
```

ramezsaeed:
Need to be fixed in light ode too.

![Screenshot from 2024-11-19 11-18-31](https://github.com/user-attachments/assets/50b60b26-a4ac-4e7b-9cc4-44770f1c8604)



check if we can unify the qr code and buttons in one place.

amiraabouhadid:
fixed google play btn in dark mode, added btns in light mode, extracted qr and btns into a separate component. will investigate why qr code shrinks in light mode

amiraabouhadid:
qr code was fine, it was a margin from light mode btn which has been removed now. No adequate google play store light btn was found so reverted back to using dark mode btns for both themes

amiraabouhadid:
used slot instead of passed msg prop to QRPlayStore component. integrated component in twin view

khaledyoussef24:
verified on devnet
Version c2d6060

<img width="1680" alt="Image" src="https://github.com/user-attachments/assets/1ca8930b-90a5-472c-b0f4-05c65579badd" />
<img width="1680" alt="Image" src="https://github.com/user-attachments/assets/2f7e2f52-f2e0-4f0a-9e2d-049d934dd6f3" />

[testcase](https://app.testlodge.com/a/26076/projects/40893/suites/234374?expand_section=422339#case_4150411)

PR #3681, opened 2024-11-25
Update app btns & QR code 

amiraabouhadid:
### Changes
- added white icons in light mode
- extracted qr and playstore btns in a component 
- used component in deposit dialog and in profile manager 

### Related Issues

 #3648

### Tested Scenarios
- in light and dark mode do the following: 
         - clicked on playstore and app store btns, see that they redirect to correct threefold connect play store and app store pages respectively 
         - open up threefold connect app, go to wallet, send page, scan qr code see that it adds correct twin id 

### Checklist

image
image


samaradel (approved):

packages/playground/public/images/icons/g_play.png
0oM4R, line 1: image


i don't think this image is the best option, to avoid that can we have both in dark ?
as in the prfile manager 
image

maayarosama (approved):

0oM4R (changes requested):
image

execuse my ocd but i think we should make both of them on same size

0oM4R (approved):

0oM4R:
not sure if we should include the twin page qr code as well,

maayarosama (approved):

packages/playground/src/components/deposit_dialog.vue
```diff
@@ -94,6 +94,8 @@ const props = defineProps({
   openDepositDialog: Boolean,
   twinId: Number,
 });
+const qrHeader = `<b> OR </b>
+      <p class="mb-3">Use ThreeFold Connect to scan this QRcode:</p>`;
```
0oM4R: i think the best practice is to use [slot](https://vuejs.org/guide/components/slots),

packages/playground/src/components/qr_play_store.vue
```diff
@@ -0,0 +1,49 @@
...
+        v-html="app.src"
+        :href="app.url"
+        target="_blank"
+      />
+    </div>
+  </div>
+</template>
+<script setup lang="ts">
+const props = defineProps({
+  qr: String,
```
0oM4R, line 23: Great! we can use this component to have the login QR code as well!

packages/playground/src/weblets/profile_manager.vue
```diff
@@ -385,26 +385,13 @@
...
-                  :style="{ cursor: 'pointer', width: '9rem' }"
-                  class="app-btn mr-2"
-                  :title="app.alt"
-                  v-html="app.src"
-                  :href="app.url"
-                  target="_blank"
-                />
-              </div>
+                <a class="app-link" :href="manual.tf_connect_app" target="_blank">Threefold Connect</a> to fund your
+                account.
```
0oM4R: please wrap those lines with a `p`  tag

packages/playground/src/dashboard/twin_view.vue
```diff
@@ -185,31 +185,14 @@
...
-                target="_blank"
-              />
-            </div>
+          <v-col lg="4" md="12" sm="12" class="my-4" v-if="profileManager.profile">
+            <QRPlayStore
+              :qr="'TFT:' + bridge + '?message=twin_' + profileManager.profile.twinId + '&sender=me&amount=100'"
+            >
+              Scan the QR code using
+              <a class="app-link" :href="manual.tf_connect_app" target="_blank">Threefold Connect</a> to fund your
+              account.</QRPlayStore
```
0oM4R: please wrap this text with p tag as well

0oM4R (commented):
great work
i notice that we changed the style

packages/playground/src/components/qr_play_store.vue
```diff
@@ -1,6 +1,6 @@
 <template>
   <div class="d-flex flex-column text-center align-center">
-    <slot></slot>
+    <p><slot></slot></p>
```
0oM4R: Why are we wrapping the slot itself? we should put the tag element on the slot content right ?

amiraabouhadid: it's the same

0oM4R: my idea was if we need to reuse this component with another content inside the slot whatever the content is will be inside a p tag, anyways thank you and great work

0oM4R (approved):

Issue #3556, opened 2024-10-28
🐞 [Bug]: logger can't read properties of value


amiraabouhadid:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### which package/s did you face the problem with?

Dashboard

### What happened?

console error from logger accessing undefined.value

### Steps To Reproduce

1. pull development branch
2. yarn lerna run build
3. make run project=playground
4. look for error in console

### which network/s did you face the problem on?

Dev

### version

current

### Relevant screenshots/screen records

![image](https://github.com/user-attachments/assets/de130547-d316-412b-9b56-1d10b3be795a)

### Relevant log output

```console
[Vue warn]: Error in setup function: "TypeError: Cannot read properties of undefined (reading 'value')"

found in

---> <VBtn>... (16 recursive calls)
       <App> at /home/amira/tfgrid-sdk-ts/packages/playground/src/App.vue
         <AsyncComponentWrapper>
```

ramezsaeed:
cannot reproduce.
please re-verify

amiraabouhadid:
was not able to reproduce on development_2.7 nor on development branches. nevertheless added conditional to check if logs.value is defined before pushing item to it

khaledyoussef24:
issue is no longer reproduicable; tried to produce the same issue many times but with different contracts but didn't pop out.

PR #3734, opened 2024-12-15
 logger fix undefined err

amiraabouhadid:
### Changes

- added a conditional to check if logs.value exists before pushing item to list 

### Related Issues

#3556

### Tested Scenarios

- start up playground, observed logs load correctly
- did some transactions (transfer by twinID) observed logs get updated correctly, no console err from logger 

image


0oM4R (approved):
tried to rent a node, deploy vm ,and transfer tfts, all works fine without any weird logs

samaradel (approved):

samaradel:
Works fine with me

PR #3742, opened 2024-12-15
Handle rented node in grace period 

samaradel:
### Description

Check if the rented node contract is in the grace period when listing the node.

### Changes

image

### Related Issues

### Tested Scenarios

- Navigate to the Node finder page.
- Rent a dedicated node.
- Leave it till it goes to the grace period.
- Navigate to any solution.
- Try to deploy an instance on the rented node.

### Checklist

- [x] Tests included
- [x] Build pass
- [ ] Documentation


samaradel:
cpd from https://github.com/threefoldtech/tfgrid-sdk-ts/pull/3646

0oM4R (commented):
image


https://github.com/threefoldtech/tfgrid-sdk-ts/pull/3646#pullrequestreview-2504471530

samaradel:
@0oM4R check it, it already fixed

amiraabouhadid (approved):
no console err 
image

0oM4R (commented):
on first if the node is selected; 
we won't see any error but if you select another node then click on the graceperiod node will see the error


also the error must be handled 

- the node is not a valid node; so we shouldn't auto select it; need to handle this case as well 




image

packages/playground/src/utils/nodeSelector.ts
```diff
@@ -235,10 +234,13 @@ export async function validateRentContract(
...
-        createCustomToast(
-          `You can't deploy on node ${node.nodeId}, its rent contract is in grace period.`,
-          ToastType.danger,
-        );
+        const err = `You can't deploy on node ${node.nodeId}, its rent contract is in grace period.`;
+        await new Promise((_, reject) => {
+          setTimeout(() => {
+            reject(Error(err));
+          }, 2000);
+        });
```
0oM4R: i don't think we need to add timeout here, not the auto one there is no toast overlap what do you think?

samaradel: totally agree, removed

0oM4R: I see you removed timeout in the automatic on to, can you explain why I think ti is a better UX even if we set it to only one sec

0oM4R (changes requested):
graceperiod


error still exsist if i select the grace period node

samaradel:
@0oM4R works with me, please make sure to have the latest update and build

0oM4R:
https://github.com/user-attachments/assets/f8faaada-8999-4d4d-8585-6b940c0b4958

packages/playground/src/components/node_selector/TfAutoNodeSelector.vue
```diff
@@ -304,14 +303,24 @@ export default {
...
+        } catch (error) {
+          const err = normalizeError(
+            error,
+            "Something went wrong while checking status of the node. Please check your connection and try again.",
+          );
+          throw err;
         }
+
         const nodeCapacityValid = await checkNodeCapacityPool(gridStore, node, props.filters);
         const rentContractValid = props.filters.dedicated ? await validateRentContract(gridStore, node) : true;
```
0oM4R: we already do the same in validateRentContract is there any extra logic added in the new added lines above

packages/playground/src/components/node_selector/TfNodeDetailsCard.vue
```diff
@@ -364,6 +356,18 @@ export default {
...
+        if (props.status === "Init" && props.node) {
+          ctx.emit("node:select", props.node as NodeInfo);
+        }
+        if (props.node?.dedicated && props.node.rentContractId === 0) return false;
+        await validateRentContract(gridStore, props.node as NodeInfo);
+      } catch (error) {
+        return error;
+      }
+    }
+
```
0oM4R: why we are calling `validateRentContract` here while it is already on the manual and auto node finder?
image
i think we need to make it just emit the select event and the parent component will handle the validation

packages/playground/src/utils/nodeSelector.ts
```diff
@@ -231,14 +230,11 @@ export async function validateRentContract(
       throw `Node ${node.nodeId} is not rented`;
     }
     if (node.rentContractId !== 0) {
```
0oM4R, line 233: I suggest having early return here if the node is not rented and not in a dedicated farm, so there is no need to continue, right?

just a reminder that the dedicated node in shared farm can be used as shared node

0oM4R (changes requested):
I think we just need the changes in utils only; no need for the other changes

packages/playground/src/components/node_selector/TfAutoNodeSelector.vue
samaradel: I see it only fires when `props.filters.dedicated` is on and with our latest changes in the dashboard we can have a  dedicated node in the normal list, not sure if we should remove it

packages/playground/src/components/node_selector/TfNodeDetailsCard.vue
samaradel: same reason https://github.com/threefoldtech/tfgrid-sdk-ts/pull/3742/files#r1966744835

packages/playground/src/components/node_selector/TfAutoNodeSelector.vue
0oM4R: so we should change this condition to be applicable with the latest changes; My point is we shouldn't do any validation inside the details card; in some cases we will do the same validation twice! on inside the details card and the other in the parent component

samaradel: agreed

0oM4R (approved):
dedicated nodes in shared farms are valid nodes 
image

validation works fine on automatic finder 

image


all works fine thanks.


image

ramezsaeed (approved):

Issue #3463, opened 2024-09-25
Reduce using watchers and use computed instead


zaelgohary:
### Which package/s are you suggesting this feature for?

Dashboard

### Is your feature request related to a problem? Please describe

We currently use more than 80 watchers in the Dashboard which can be expensive performance-wise. Some of them are necessary and some are not.

![image](https://github.com/user-attachments/assets/f22e1ae4-b10f-4de3-b2ab-a3d7c414e107)

### Describe the solution you'd like

Reduce watchers with computed properties to enhance performance.

amiraabouhadid:
replaced these watched variables with computed: 
- sidebar toolbarExtended 
-  logger collapsed watcher 
-  node details dialog 
- VM image name, flist and entyPoint

amiraabouhadid:
found a bug introduced by changing sidebar toolextended var

amiraabouhadid:
fixed bug, tested other changes for any new bugs

amiraabouhadid:
Rewrote sidebar handling to reduce watchers

amiraabouhadid:
Went through remaining watchers in pricing calculator and algorand capacity

amiraabouhadid:
replaced 3 watchers in algorand, removed unnecessary code in node details, tested changes in node details, logger and algorand

amiraabouhadid:
tried out replacing watchers in app_theme, dynamic_tabs and form_validator. Will test any negative side effects next.

amiraabouhadid:
remove watcher in app_theme, changes in dynamic tabs and form validator introduced bugs so reverted changes. Examined watchers in input validator and filter, replacing watchers is not feasible

amiraabouhadid:
replaced iswireguard watcher to computed in manage gateway dialog, tested with domain add and domain list. looked at watchers in 

- networks
-  node details
-  root fs
-  simulator
-  tf navigation loader

none can be replaced with computed

amiraabouhadid:
replaced watcher in 

- userfarms
- node filters
- public config  
- reserve btn. 

Looking into watchers in transfer view and full_vm filters.

amiraabouhadid:
replaced/reduced watchers in the following files and tested for any negative side effects:
- transfer_view
- full_vm
- micro_vm
- tf_algorand
- tfrobot
- algorand_capcity
removed no longer used watch imports to reduce overhead.
went over all files remaining with watch to make sure replacing them is not feasible.
[...]

PR #3841, opened 2025-01-29
Replace watchers

amiraabouhadid:
### Description

reduced number of watchers used throughout playground rendering

### Changes

replaced watchers whenever possible with computed or combined multiple watchers into one or removed unnecessary watchers in order to enhance performance

### Related Issues

#3463

### Tested Scenarios

- reran all edited solutions deployments after reducing watchers
- tested normal login/logout/connect wallet flow
- tested logger for correct log updates & collapse/expand
- tested node selector correct toggles and results with reserve/unreserve testing 
- tested farms correct rendering /reload after farm create 
- correct transfer flow and amount validation
- correct theme change and toggle with refresh 
- correct public config update

### Documentation PR

#### Preliminary Checks:

- [x] Preliminary Checks
  - What about edge cases?

#### UI Checks:

- [x] UI Checks
  - Does every button work?
  - Does every validation work?

#### Code Quality Checks:

- [x] Code Quality Checks

### Testing Checklist

- [x] Tested the un-happy path and rollback scenarios.
- [x] Documentation updated to meet the latest changes.
- [x] Can be covered by automated tests.

### General Checklist

- [x] Tests included
- [x] Build pass
- [ ] Documentation
- [ ] Code format and docstring


packages/playground/src/weblets/tf_algorand.vue
```diff
@@ -157,8 +157,12 @@ const selectedSSHKeys = ref("");
...
 const grid = gridStore.client as GridClient;
 
-watch(firstRound, () => lastRoundInput.value.validate(lastRound.value.toString()));
-
+const isValid = computed(() => {
+  if (firstRound.value) {
+    return lastRoundInput.value.validate(lastRound.value.toString());
+  }
+  return false;
+});
```
zaelgohary: isValid is not used anywhere.

packages/playground/src/dashboard/transfer_view.vue
```diff
@@ -146,13 +146,13 @@ const receptTwinFromTwinID = ref<Twin>();
...
+const computedValidation = computed(() => {
   if (transferAmount.value) {
-    await amountRef.value?.reset();
+    amountRef.value?.reset();
     amountRef.value?.validate();
   }
+  return null; // or some meaningful return value if needed
 });
-
 const tick = ref(0);
```
zaelgohary: computedValidation is not used anywhere in the code.

packages/playground/src/dashboard/components/public_config.vue
```diff
@@ -193,13 +193,15 @@ export default {
...
     const grid = useGrid();
 
     const defualtNodeConfig = ref<PublicConfig>(publicConfigInitializer());
     const config = ref<PublicConfig>(publicConfigInitializer());
-
+    const isConfigChanged = computed(() => {
+      formRef.value?.validate();
+      return !isEqual(defualtNodeConfig.value, config.value);
+    });
     function publicConfigInitializer(): PublicConfig {
```
zaelgohary: Computed properties should be pure functions. Also, validation should only run if condition is true. This should be applied on all changed computed properties.

packages/playground/src/components/select_vm_image.vue
```diff
@@ -58,22 +58,23 @@ const props = defineProps({
...
+  }
+  return "";
+});
+
+const entryPoint = computed(() => {
+  if (name.value != "Other") {
+    return image.value.entryPoint;
+  }
+  return "";
+});
```
zaelgohary: flist & entrypoint always returns an empty string if other is chosen.

https://github.com/user-attachments/assets/a185e117-55b1-4bd1-869b-7e63eed6a1ab

packages/playground/src/components/app_theme.vue
```diff
@@ -4,31 +4,47 @@
...
+function setTheme(themeName: string) {
+  theme.global.name.value = themeName;
+  updateLocalStorage(themeName);
+  isLight.value = themeName === "light";
+}
+
+function changeTheme() {
+  const newTheme = isLight.value ? "dark" : "light";
+  setTheme(newTheme);
+}
```
zaelgohary: I get this err.

Image

packages/playground/src/components/node_details.vue
```diff
@@ -1,6 +1,6 @@
 <template>
   <v-dialog
-    v-model="dialog"
+    v-model="$props.openDialog"
     @update:modelValue="(val:boolean) => closeDialog(val)"
```
zaelgohary, line 4: I don't think this is a good approach as v-model uses 2-way binding and props are immutable so there might be some issue when we try to mutate it.

packages/playground/src/weblets/tf_deployment_list.vue
```diff
@@ -14,6 +14,7 @@
     >
       <VmDeploymentTable
         :projectName="tabs[activeTab].value"
+        :projectTitle="tabs[activeTab].title"
```
zaelgohary: where is this used?

packages/playground/src/components/app_theme.vue
amiraabouhadid: please include steps to reproduce as no err is shown in my console when i toggle theme

amiraabouhadid: did find a bug when toggling theme from settings page when trying to reproduce now so thanks haha

packages/playground/src/weblets/tf_algorand.vue
amiraabouhadid: correct, it isn't, it is used to replace a watcher.

packages/playground/src/dashboard/transfer_view.vue
amiraabouhadid: same here correct, it isn't used, but it is used to replace a watcher.

packages/playground/src/dashboard/components/public_config.vue
amiraabouhadid: I will have disregard this in favor of improved performance and less watchers tbh, lemme know if you have a better suggestion though

packages/playground/src/weblets/tf_deployment_list.vue
amiraabouhadid: irrelevant change from another branch

packages/playground/src/components/node_details.vue
amiraabouhadid: it is used in many other components safely 
Screenshot 2025-05-04 at 15 30 38

packages/playground/src/components/algorand_capacity.vue
```diff
@@ -100,9 +100,19 @@ const cpu = ref(8);
...
+  emits("update:storage", storage.value);
+};
+
+watch(
+  () => [cpu.value, memory.value, storage.value],
+  () => {
+    emitResourceUpdates();
+  },
+  { immediate: true },
+);
```
0oM4R: We shouldn't use a single call for three emits, if the CPU only got changed, we will have another 2 unnecessary emits

packages/playground/src/components/app_theme.vue
```diff
@@ -4,31 +4,43 @@
...
+import { LocalStorageSettingsKey, updateLocalStorage } from "@/utils/settings";
 
 const theme = useTheme();
-const light = ref(false);
 
-watch(light, light => (theme.global.name.value = light ? "light" : "dark"));
-watch(theme.global.name, theme => {
-  localStorage.setItem(LocalStorageSettingsKey.THEME_KEY, theme);
-  light.value = theme === "light";
+const isLight = ref(false);
```
0oM4R: why do we neeed islight i think the computed light can do the same

packages/playground/src/dashboard/components/public_config.vue
0oM4R: Replacing watchers, I think should replace the replaceable watchers, in this case this is not a replaceable watcher

packages/playground/src/dashboard/transfer_view.vue
0oM4R: i think this is not a replacable watcher

packages/playground/src/utils/settings.ts
```diff
@@ -11,3 +11,6 @@ export enum LocalStorageSettingsKey {
 export enum SessionStorageSettingsKey {
   PASSWORD_KEY = "password",
 }
+export const updateLocalStorage = (newTheme: string) => {
```
0oM4R:
```suggestion
export const updateThemeInLocalStorage = (newTheme: string) => {
```

0oM4R (changes requested):
Great effort ya Amira 
Partial review; please consider that the watches are important, we can't replace them, the watch and the computed are different, and each has a specific job.Many changes do not follow the best practice of using computed.

packages/playground/src/components/app_theme.vue
amiraabouhadid: you need it for the onMount, otherwise you get this heinous err if you rely solely on the computed light var 
Screenshot 2025-05-12 at 13 15 07

0oM4R: Not sure how you applied that, but I did that and totally 

image

packages/playground/src/dashboard/farms_view.vue
```diff
@@ -29,13 +29,11 @@ export default {
     const userFarms = ref();
     function handleFarmCreated() {
       farmsReload.value = !farmsReload.value;
-    }
-    watch(
-      () => farmsReload.value,
-      () => {
+      if (userFarms.value) {
         userFarms.value.reloadFarms = farmsReload.value;
```
0oM4R, line 33: what is the idea of having two values got updated on reload, if that is not part of your issue, I will open one to revisit this logic

packages/playground/src/views/nodes.vue
```diff
@@ -546,18 +547,21 @@ export default {
...
-          oldNodeStatus.value = filters.value.status;
-          filters.value.status = UnifiedNodeStatus.UpStandby;
-        } else {
-          filters.value.status = oldNodeStatus.value;
-        }
-      },
-    );
+    const oldNodeStatus = computed(() => {
+      return filters.value.status;
+    });
```
0oM4R: Doesn't truly preserve the previous value of status.

oldNodeStatus just mirrors filters.value.status, so when rentable is true and status is overridden, oldNodeStatus.value becomes UpStandby, and thus you lose the original.

this can be tested by selecting rentable you will see that the value of status field will be 
image

and will be disabled, and if you toggled off the rentable should go back to the previous state

This behavior got changed in this pr; after toggling off the rentable, the old status doesn't go back

packages/playground/src/weblets/tf_algorand.vue
0oM4R: so there is no need to replace it with computed, this is the perfect scenario to have watcher imo

packages/playground/src/dashboard/farms_view.vue
amiraabouhadid: this logic is removed in this pr, so comment is irrelevant

packages/playground/src/weblets/tf_algorand.vue
amiraabouhadid: there is a need for efficiency purposes, please compare the efficiency burden of an unused computed variable to that of a watcher. If it is more, I will revert back to watcher

0oM4R: computed is more efficient for tracking derived state, but we are not tracking that data at all 
and the watch is more efficient for triggering side-effects, which is the case here

packages/playground/src/components/LogMessage.vue
```diff
@@ -80,16 +80,7 @@ export default {
...
-
-    watch(
-      () => props.value.log.id,
-      () => {
-        if (!collapsed.value) {
-          collapsed.value = true;
-        }
-      },
-    );
+    const collapsed = computed(() => !!props.value.log?.id);
```
samaradel: While testing from a private window, I got many errors, and when I clicked on show more I got  
image

image

samaradel: [Screencast from 18-05-25 16:50:30.webm](https://github.com/user-attachments/assets/3c582250-30a9-42ec-9483-1ec15d869f4d)

packages/playground/src/weblets/tf_full_vm.vue
```diff
@@ -185,7 +185,11 @@ const dedicated = ref(false);
 const rentedByMe = ref(false);
 const certified = ref(false);
 const disks = ref<Disk[]>([]);
-const hasGPU = ref(false);
+const GPUToggled = ref(false);
+const hasGPU = computed(() => {
+  return (!dedicated.value && !rentedByMe.value) || GPUToggled.value;
```
0oM4R: image
this case is not valid, i can't select GPU without selecting rented by me `or` dedicated
hasgpu should be true if the gputoggled and one of the dedicated and rented by is on

0oM4R (approved):

samaradel (approved):

Issue #3837, opened 2025-01-27
🐞 [Bug]: dedicated flag shows nodes that I didn't rent myself


xmonader:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### which package/s did you face the problem with?

Dashboard

### What happened?

![Image](https://github.com/user-attachments/assets/cda1ed82-4e71-46c8-9b57-4ac2b052b27d)

### which network/s did you face the problem on?

Main

### version

latest

### Relevant screenshots/screen records

![Image](https://github.com/user-attachments/assets/6e43ad7b-bf10-4388-81d2-3bf4db7e5698)

### Relevant log output

```console
-
```

What I'd like to have is to remove the ambiguity of the Dedicated flag to be
- New flag `Rented by me` 
- New flag `Include rentable nodes` , that's basically the behavior of the current dedicated (so just a rename in the UI)

samaradel:
### Work in Progress (WIP):
Add rented by me and include rentable nodes filter (only on micro VM), on switching on both, filters conflicts each other

samaradel:
### Work Completed:
Fix rentable and rented by when both filters are fired

samaradel:
### Work Completed:
Remove `rentedByMe` filter

samaradel:
### Work Completed:
- Handle `renredBy` filter
- Add filters to `freeflow` app

samaradel:
### Work Completed:
Rename filters

samaradel:
### Work Completed:
- Remove unwanted contract validation
- Early exit for check it node is rented and not in dedicated farm
- Remove the check only if the dedicated filter is ON

khaledyoussef24:
verified ondevnet
Version 96da187

the New filter works fine after renting a node while deploying and shows only nodes rented by me

<img width="1680" alt="Image" src="https://github.com/user-attachments/assets/a3e7db0b-81ce-48f6-aa16-901195710186" />

[testcase](https://app.testlodge.com/a/26076/projects/40893/suites/234374?expand_section=426661#case_4467830)

PR #3852, opened 2025-01-30
Add rented by me and include rentable nodes filter

samaradel:
### Description

Add rented by me and include rentable nodes filter

### Changes

[Screencast from 17-02-25 20:42:05.webm](https://github.com/user-attachments/assets/beb4543a-8155-4631-af03-0e6dc8d9087d)

#### Updated filters
image

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/3837
### Tested Scenarios

- Switch Rented by me (will show only the rented nodes)
- Switch Include rentable nodes (will show only the rentable nodes)
- Switch both (will show the rentable nodes and rented nodes )

### Documentation PR

- https://github.com/threefoldtech/info_grid/issues/742

#### Preliminary Checks:

- [x] Preliminary Checks
  - What about edge cases?

#### UI Checks:

- [x] UI Checks
  - Does every button work?
  - Does every validation work?

#### Code Quality Checks:

- [x] Code Quality Checks

### Testing Checklist

- [ ] Does it Meet the specified acceptance criteria.
- [ ] Test if changes can affect any other functionality.

### General Checklist

- [x] Tests included
- [x] Build pass
- [x] Documentation


packages/playground/src/components/k8s_worker.vue
```diff
@@ -76,8 +76,17 @@
       v-model.number="$props.modelValue.rootFsSize"
     />
 
+    <!-- <input-tooltip inline tooltip="" :href="manual"> -->
```
0oM4R, line 79: Im assuming this is commented waiting for the documentation, right? if yes, please open an issue to remove it

packages/grid_client/src/modules/models.ts
```diff
@@ -634,6 +634,7 @@ class FilterOptions {
   @Expose() @IsOptional() @IsBoolean() hasGPU?: boolean;
   @Expose() @IsOptional() @IsBoolean() rentable?: boolean;
   @Expose() @IsOptional() @IsInt() @Min(1) rentedBy?: number;
+  @Expose() @IsOptional() @IsBoolean() rentedByMe?: boolean;
```
0oM4R: is this needed?

0oM4R (changes requested):
I don't think we need to have a rentedbyMe to be passed to the filters; we can have computed value that returns the twinid or undifinded based on the rentedByMe toggle value and pass it to the rentedby

packages/playground/src/components/k8s_worker.vue
samaradel: opened and mentioned in the pr

packages/grid_client/src/modules/models.ts
samaradel: I think it would be more clear, I'm fine with both 
@AhmedHanafy725 what do you think?

samaradel: filter removed

packages/playground/src/utils/nodeSelector.ts
```diff
@@ -207,7 +209,9 @@ export function normalizeNodeFilters(
     country: options?.location.country,
     gateway: options?.gateway,
     healthy: true,
-    rentableOrRentedBy: filters.dedicated ? options?.twinId : undefined,
+    rentable: bothRentalFiltersActive ? undefined : filters.dedicated || undefined,
+    rentedBy: bothRentalFiltersActive ? undefined : filters.rentedBy ? options?.twinId : undefined,
```
0oM4R: `filters.rentedBy` is containing the twin id already, right? i

packages/playground/src/types/index.ts
```diff
@@ -71,6 +72,8 @@ export interface CaproverWorker {
   name: string;
   solution?: solutionFlavor;
   dedicated?: boolean;
+  rentedBy?: number;
+  rentedByMe?: boolean;
```
0oM4R, line 85: do we still need rentedByMe?

packages/playground/src/types/index.ts
```diff
@@ -63,6 +63,7 @@ export interface K8SWorker {
   rootFsSize: number;
   rentedBy?: number;
   dedicated: boolean;
+  rentedByMe?: boolean;
```
0oM4R, line 74: do we still need rentedByMe?

0oM4R (commented):
freeflow is not updated with new changes, please check

packages/playground/src/utils/nodeSelector.ts
samaradel: yes or undefined

packages/playground/src/types/index.ts
samaradel: it's for the interface as it is used for the switch of the workers only

packages/playground/src/types/index.ts
samaradel: same usage https://github.com/threefoldtech/tfgrid-sdk-ts/pull/3852#discussion_r1963724208

packages/playground/src/utils/nodeSelector.ts
0oM4R: so i suggest 
```suggestion
    rentedBy: bothRentalFiltersActive ? undefined : filters.rentedBy || undefined,
```

samaradel:
It's not a part of the dashboard applications  anymore

0oM4R:
> It's not a part of the dashboard applications anymore

It should be added; we may reuse the application again; then the component should be updated.

packages/playground/src/types/nodeSelector.ts
```diff
@@ -22,6 +22,8 @@ export interface SelectionDetailsFilters {
   solutionDisk?: number;
   certified?: boolean;
   dedicated?: boolean;
+  rentedBy?: number;
+  rentable?: boolean;
   exclusiveFor?: string;
   rentable_or_rented_by?: number;
```
0oM4R, line 28: rentable_or_rented_by?: number;
this should be removed as we are not using it any where

packages/playground/src/weblets/tf_staticwebsite.vue
```diff
@@ -144,6 +149,10 @@ const flist: Flist = {
   entryPoint: "/sbin/zinit init",
 };
 const dedicated = ref(false);
+const rentedByMe = ref(false);
+const gridStore = useGrid();
+const grid = gridStore.client as GridClient;
```
0oM4R, line 154: we need to use this variable inside the deploy method as the other solutions; __can be handled in separate issue__

0oM4R (commented):
will test caprover worker later when we have node that have ipv4 rentable

packages/playground/src/types/nodeSelector.ts
samaradel: we still use it when the both filters are active

0oM4R:
I suggest changing the `include rentable nodes`  title to be `Renatble`
what if the use only toggle on the `include rentable nodes` so what are we including it on? it will list only rentable nodes 
also `Rented by me` should be on by default, as @xmonader ask



should we update the tooltip to be learn more about rentable nodes?

image

make sure to pull the development branch

samaradel:
I agree with renaming but for turning on my rented nodes filter will list only my nodes so if it doesn't satisfy the requirements it will show no nodes found.
The current behavior "without turning on any filter" will show the rented nodes on the top of nodes, I think it's the behavior you want.

0oM4R (commented):
I dont think we need `nodes` word here as we do in certified 

image

Rented by me (only)
Rentable
Certified 

also the tooltip i think should changed as i mentioned 
image



@ehab-hassan what do you think?

xmonader (approved):

PR #3687, opened 2024-11-28
change url for terms and conditions from threefold library to manual

mik-tf:
### Description

- Changed the URL for the Terms and Conditions from Library to Manual

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/3686

### Preliminary Checks:

- Does it completely address the issue linked?
  - IMO, it should work but please test. 

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation
- [ ] Code format and docstrings
- [ ] Screenshots/Video attached (needed for UI changes)

Issue #3686, opened 2024-11-28
Update Links for T&C on Dashboard


mik-tf:
### Which package/s are you suggesting this feature for?

Dashboard

### Is your feature request related to a problem? Please describe

Kristof wants us to remove the library.threefold.me. But as @sabrinasadik  pointed out here: https://github.com/threefoldtech/tf_operations/issues/2958, we need to make sure the links on the dashboard for the T&C are still valid.

### Describe the solution you'd like

We could point to the Manual T&C links instead, which are the content copied from the library.threefold.me

- Links
  - https://manual.grid.tf/knowledge_base/legal/terms_conditions_all3.html

atefnazmy:
- Verified on Dev,
- Version: 85b7dcf
- Terms and Conditions url has been updated from threefold library to manual

https://github.com/user-attachments/assets/8bdfcd16-fe47-4219-a2cc-38275b7d39fe

#### Related Link:
- [Testcase](https://app.testlodge.com/a/26076/projects/40893/suites/234374?expand_section=442285#case_4150401)

PR #3866, opened 2025-02-05
Set terms and conditions page without library link

mik-tf:
### Description

- Changed the URL for the Terms and Conditions from Library to Manual
- Added a markdown file with pictures with direct links to manual
- Tested locally and it works.

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/3686

### Preliminary Checks:

  - Yes. The goal is to not use library.threefold.me

### Notes

- This replaces PR #3687 

@0oM4R I think this should be better than the previous PR.


maayarosama:
Codacy check is failing, please fix

0oM4R:
@AhmedHanafy725 is it okay to clone the content into our project, or it should be from a single source of it?

mik-tf:
Thanks @maayarosama! I didn't know we could not inject HTML into Markdown!

The page has less aesthetic features, but now all tests pass.

@0oM4R waiting for a confirmation on your question. Thanks.

AhmedHanafy725:
> @AhmedHanafy725 is it okay to clone the content into our project, or it should be from a single source of it?

it shouldn't be on the dashboard, it should be in one place

mik-tf:
@0oM4R @AhmedHanafy725 OK good. Thus I rewrote it so now it takes the markdown from the github repo of the manual, and parses it. Basically, it now does what we had before for library but now for the manual. All tests pass too and it works on a local version of the Dashboard. Should be ok. LMK! And thanks for bearing with me through this never-ending PR+issue. :D

packages/playground/src/weblets/profile_manager.vue
```diff
@@ -841,36 +841,37 @@ function parseAcceptTermsImage(tempDiv: HTMLDivElement, url: string) {
...
-}
 watch(openAcceptTerms, async () => {
   if (openAcceptTerms.value) {
     try {
-      const url = "https://library.threefold.me/info/legal/";
-      const response = await fetch(url + "readme.md");
-      const mdContent = await response.text();
-      const parsedContent = marked.parse(mdContent);
+      const response = await fetch(
+        "https://raw.githubusercontent.com/threefoldtech/info_grid/master/src/knowledge_base/legal/terms_conditions_all3.md",
```
maayarosama: let's add the link to manual.ts file, import the manual and use it from there

packages/playground/src/weblets/profile_manager.vue
```diff
@@ -841,36 +841,37 @@ function parseAcceptTermsImage(tempDiv: HTMLDivElement, url: string) {
...
+        const absoluteUrl = `https://manual.grid.tf/knowledge_base/legal/${relativePath}.html`;
+        return `[${linkText}](${absoluteUrl})`;
+      });
+
+      // Handle links in subdirectories
+      mdContent = mdContent.replace(/\[([^\]]+)\]\((\.\/[^)]+\/[^)]+)\.md\)/g, (match, linkText, path) => {
+        const relativePath = path.replace("./", "");
+        const absoluteUrl = `https://manual.grid.tf/knowledge_base/legal/${relativePath}.html`;
+        return `[${linkText}](${absoluteUrl})`;
+      });
```
maayarosama: let's extract this part into an externalparseAcceptTermsLink  function, so the watch is less cluttered.

packages/playground/src/weblets/profile_manager.vue
maayarosama, line 1: parseAcceptTermsImage function isn't used anymore, please remove it

maayarosama (changes requested):

packages/playground/src/weblets/profile_manager.vue
mik-tf: Done in commit 3e46425e6aafce89ed4c6862dbe78d3c1d95a17b

packages/playground/src/weblets/profile_manager.vue
mik-tf: Done in commit 40fe3c7c3a28da5b878aeecf9cf73bde29982975

packages/playground/src/weblets/profile_manager.vue
mik-tf: Done in commit 0c20c48b3ee43e7b792485f7ce399dd00ae92e36.
Made 3 functions instead of 1 big.

Thanks for all the review + tips.

mik-tf:
@maayarosama lmk what you think! I made 3 commits with the review you made. Thanks again.

packages/grid_client/src/modules/tfchain.ts
```diff
@@ -477,7 +477,9 @@ class TFChain implements blockchainInterface {
       });
     }
     await (
-      await client.termsAndConditions.accept({ documentLink: "https://library.threefold.me/info/legal/#/" })
+      await client.termsAndConditions.accept({
+        documentLink: "https://manual.grid.tf/knowledge_base/legal/terms_conditions_all3.html",
```
0oM4R, line 481: please use the exported url from the utils/manaul

maayarosama:
Works fine with me
[Screencast from 2025-02-11 13-13-58.webm](https://github.com/user-attachments/assets/31bae3dc-b708-4672-9a28-c36836066015)

packages/grid_client/src/modules/tfchain.ts
mik-tf: Done in commit 7724a374daad49f1984d4519de641ef5486757b6 and quick fix in 8caedc013a1c69398dc41a510e234a9e365ef758. Thanks. 

Update: still have to fix something. Will ping when ready.

mik-tf: @0oM4R You can see I tried different ways to make it work, but it is not working. As I understand, it means we need to change the playground code to accommodate being in a browser or being run in node.

As the link was hard coded before my PR, I think fixing this would be the case of a separate issue. If you have suggestions I can explore too but I think it will change more code than what is the scope of this PR. So before moving forward, either devs provide some insight into how they would prefer to do it, and/or we do another issue for this specific issue.

Update: Btw sorry for all those failed workflow tests. I will run the test locally from now on before pushing to the branch within a PR.

0oM4R (dismissed):

packages/playground/src/weblets/profile_manager.vue
```diff
@@ -832,45 +833,33 @@ function validateConfirmPassword(value: string) {
...
+const replaceMarkdownLinks = (content: string, pattern: RegExp, baseUrl: string) => {
+  return content.replace(pattern, (_, linkText, path) => {
+    const relativePath = path.replace("./", "");
+    return `[${linkText}](${urlJoin(baseUrl, `${relativePath}.html`)})`;
   });
-}
+};
+
+const processMarkdownContent = (content: string, baseUrl: string) => {
+  let processedContent = content.replace("./" + manual.legal_header_img, manual.manual_raw_legal_img);
```
0oM4R: legal image is not as before: 
image

before was 
image

the class info-legal-image must be added to apply the style
as in

[tfgrid-sdk-ts/packages/playground/src/weblets/profile_manager.vue](https://github.com/threefoldtech/tfgrid-sdk-ts/blob/f9cfcc4e9cc6bdf4d3e3bfc4bcc8f64c25b65e38/packages/playground/src/weblets/profile_manager.vue#L839-L840)

Lines 839 to 840 in [f9cfcc4](https://github.com/threefoldtech/tfgrid-sdk-ts/commit/f9cfcc4e9cc6bdf4d3e3bfc4bcc8f64c25b65e38)

     // Update the style of the image. 
     imgElement.setAttribute("class", "info-legal-image");

0oM4R (changes requested):

packages/grid_client/src/modules/tfchain.ts
0oM4R: nvm we will handle it later thank you

packages/playground/src/weblets/profile_manager.vue
mik-tf: Good observation. I tried to do it before by injecting html but I see it was the wrong way. Thanks!

Done in b04c29c9ccd581254c2bed099900026677073228

mik-tf:
# Update

- All changes made
- Lmk if there's anything else. Thanks!

0oM4R (approved):
telegram-cloud-photo-size-4-5868265073120757905-y


functional now, If there are any further changes needed in style, I suggest opening a new issue.

@mik-tf  thank you!

amiraabouhadid (commented):
all links work but when I change the screen size, the terms and conditions dialog disappears and I'm redirected to the login tab which can be confusing
https://www.loom.com/share/18e8b96df62646f6898760846ed5034f?sid=8bdd61d8-1107-40a0-a425-c570e0a4b684
Also, can we please reduce the title font size, seems too big for me 
image

mik-tf:
> all links work but when I change the screen size, the terms and conditions dialog disappears and I'm redirected to the login tab which can be confusing

What you describe here is also happening with the current Legal page of the current Dashboard. So it should be a separate issue (it's an issue in itself if we want to fix this).

mik-tf:
> Also, can we please reduce the title font size, seems too big for me

@amiraabouhadid this is a style issue, and I quote @0oM4R here: "functional now, If there are any further changes needed in style, I suggest opening a new issue."

If you want to adjust the style, we can simply merge this PR, and create a new one with "style" as the focus. Lmk. Thanks for all the review + help.

mik-tf:
@maayarosama I think all the changes you requested have been resolved. If that is the case, can you approve the PR or approve all the changes? It still says there is 1 change requested. Thanks!

Issue #4027, opened 2025-04-07
🐞 [Bug]: Total Cost Changes When Pagination Settings Are Modified


atefnazmy:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### which package/s did you face the problem with?

Dashboard

### What happened?

The total cost of contracts displayed on the dashboard changes when the pagination item number (items per page) is updated. the total cost should be calculated based on all contracts, regardless of pagination.

### Steps To Reproduce

1. Navigate to the Dashboard and connect your wallet.
2. From the Sidebar select Deploy then Your Contracts.
3. Observe the total cost displayed at the top-left of the table.
4. Change the pagination setting and revise the total cost again.

### which network/s did you face the problem on?

Dev

### version

6f50738

### Relevant screenshots/screen records

https://github.com/user-attachments/assets/7b3958a9-289e-4d15-854f-fcb071b55fba

### Relevant log output

```console
n/a
```

ramezsaeed:
Acceptance testing:
- Test all pagination page are listed in total cost
- add/delete contracts should reflect in the total cost
- the query should be the same whenever the page pagination changed.

samaradel:
### Work Completed:
The contracts list component is modified to calculate the total cost for all contracts instead of just the paginated ones.

samaradel:
### Work Completed:
Prevent load the total cost on pagination

samaradel:
### Work Completed:
Checking the PR comments

samaradel:
### Work Completed:
Reduce get the total cost function calls

samaradel:
### Work Completed:
- Use the gridproxy consumption endpoint
- Calculate the total cost after loading contracts
- Double check the total cost in case no contracts and on init new contracts
- Fix get total cost stuck loading

samaradel:
### Work Completed:
- Revert custom events
- Call the get total cost function only from inside the load contracts function

khaledyoussef24:
verified on devnet
Version e0ea734
now total cost is consistent
it shows the cost of all deployment not only per page

https://github.com/user-attachments/assets/09df0627-a17d-438b-ab00-74a22a1dc0e4
[testcase](https://app.testlodge.com/a/26076/projects/40893/suites/234374?expand_section=430129#case_4523161)

PR #4072, opened 2025-04-24
Handle the total cost for all contracts

samaradel:
### Description

I've modified the contracts list component to calculate the total cost for all contracts instead of just the paginated ones. 

### Changes

[Screencast from 24-04-25 15:34:42.webm](https://github.com/user-attachments/assets/28e5ff0c-d01a-4ea6-8a59-68b918022673)

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/4027

### Tested Scenarios

Scenarios provided by [Ramez](https://github.com/threefoldtech/tfgrid-sdk-ts/issues/4027#issuecomment-2820820110):

- Test that all pagination pages are listed in the total cost
- Add/delete contracts should reflect in the total cost
- The query should be the same whenever the page pagination changes.

### Documentation PR

#### Preliminary Checks:

- [x] Preliminary Checks
  - What about edge cases?

#### UI Checks:

- [x] UI Checks
  - Does every button work?
  - Does every validation work?

#### Code Quality Checks:

- [x] Code Quality Checks

### Testing Checklist

- [x] Can be covered by automated tests.

### General Checklist

- [x] Tests included
- [x] Build pass
- [ ] Documentation


amiraabouhadid (commented):
I don't think the total cost should load everytime pagination is changed 
https://www.loom.com/share/d9d1d5cddb804cef9e508139767657d4?sid=2511c8cb-debc-4968-ae32-f8ab86580af5

packages/playground/src/weblets/tf_contracts_list.vue
```diff
@@ -465,12 +469,47 @@ const nodeStatus = computed(() => {
...
+    const allNodeContracts = await _normalizeContracts(nodeResponse.data, ContractType.Node);
+    const allNameContracts = await _normalizeContracts(nameResponse.data, ContractType.Name);
+    const allRentContracts = await _normalizeContracts(rentResponse.data, ContractType.Rent);
+
+    const allContracts = [...allNodeContracts, ...allNameContracts, ...allRentContracts];
+    for (const contract of allContracts) {
+      const consumption =
+        contract.consumption !== undefined && contract.consumption !== null ? contract.consumption.valueOf() : 0;
+      totalCost.value = +new Decimal(totalCost.value).add(consumption);
+    }
```
0oM4R, line 504: The same logic of listing contracts is already used in the  grid client 
https://github.com/threefoldtech/tfgrid-sdk-ts/blob/b65f6b43b809181aebc274c5576b4c06726b9ead/packages/grid_client/src/modules/contracts.ts#L300-L312

packages/playground/src/weblets/tf_contracts_list.vue
```diff
@@ -386,7 +388,9 @@ async function loadContracts(type?: ContractType) {
     contracts.value = [...nodeContracts.value, ...nameContracts.value, ...rentContracts.value];
 
     // Update the total cost of the contracts.
-    await getTotalCost();
+    if (!type) {
+      await getTotalCost();
+    }
```
0oM4R, line 393: can you explain this change, please? I think the get total cost should be triggered on mount, and on refresh after delete contract, and we should ignore all other actions

samaradel: To avoid calling the `getTotalCost()` on unwanted cases like using pagination or sorting, I see the type isn't passed in several cases like the `refresh` button and `onDeletedContracts()` so I used this condition

packages/playground/src/weblets/tf_contracts_list.vue
samaradel: Please explain what's the problem with using `gridProxyClient`?

0oM4R: code redundance, the old function is already tested, and we are using it in many places so why do we need to introduce a new method that does the same exact thing?

packages/playground/src/weblets/tf_contracts_list.vue
0oM4R: IMO we need to unattach the getTotalCost function from loadContracts as the use case of getTotalCost now changed, even if the loadContracts failed we need to show the total cost, whatever happen in loadContracts whatever contracts got listed we need to show the Total cost of all contracts; 
we can call the getTotal cost in 
- onmounted 
- onclick refresh
- ondeletedContracts function after the timeout

0oM4R:
I see nothing changed please let me know if you have any question, or what do you think about the comments

amiraabouhadid:
cost is no longer loading when pagination changes, great job ya samar. Please also address omar's comments or provide explanation as to why you choose to not address them, thanks

samaradel:
guys, please check the time of my comments and yours, I didn't ask for review after your comments :D

PR #4174, opened 2025-05-27
OI app, set high specs for greater UX

### LINKED ISSUE #4173: 🐞 [Bug]: Define the small instance config to match the minimum requirement for  Open WebUI solution
### Is there an existing issue for this?

- [x] I have searched the existing issues

### which package/s did you face the problem with?

Dashboard

### What happened?

as user can deploy an open webUi instance with the min requirements in the deployment page; this suppose to work as expected.
but because the space is too small the solution always hangs; so we need to set the minimum requirement for a successful deployment and block the deploy button otherwise.

### Steps To Reproduce

  - go to Labs
  - select open WebUi
  - don't change any value, only select the domain
  - click deploy
  - the deploy succeeded and user can ssh to the machine
Then
  - the user cannot access the url at all and always get 500
  - if the user was lucky and can open the webui url; the create account is not responsive.

### which network/s did you face the problem on?

Test

### version

2.8.0-rc4

### Twin ID/s

_No response_

### Node ID/s

_No response_

### Farm ID/s

_No response_

### Contract ID/s

_No response_

### Relevant screenshots/screen records

Image

Image
Image

### Relevant log output

```console
--
```

mik-tf:
# Related Issue

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/4173

# Work Done

- Updated the specs so we are sure it provides a nice UX for the user
- In the docs, we will state the "minimum" specs so users can use the "cheapest" deployment if needed.

# References

- Closes #4173

### DIFF
--- a/packages/playground/src/weblets/tf_openwebui.vue
+++ b/packages/playground/src/weblets/tf_openwebui.vue
@@ -32,9 +32,9 @@
         </input-validator>
 
         <SelectSolutionFlavor
-          :small="{ cpu: 2, memory: 8, disk: 25 }"
-          :medium="{ cpu: 4, memory: 16, disk: 50 }"
-          :large="{ cpu: 8, memory: 32, disk: 100 }"
+          :small="{ cpu: 4, memory: 16, disk: 125 }"
+          :medium="{ cpu: 8, memory: 32, disk: 250 }"
+          :large="{ cpu: 16, memory: 64, disk: 500 }"
           v-model="solution"
         />
 


0oM4R (changes requested):
Please make sure to handle the custom flavor
This should be done by adding validation on the input fields in custom case

0oM4R:
> Please make sure to handle the custom flavor This should be done by adding validation on the input fields in custom case

handled in https://github.com/threefoldtech/tfgrid-sdk-ts/issues/4188

0oM4R (approved):
works fine on small package, installed gemma3

Issue #3391, opened 2024-09-09
🐞 [Bug]: fqdn gw workload doesn't create a name contract


Omarabdul3ziz:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### which package/s did you face the problem with?

grid_client

### What happened?

Each gateway workload should create two contracts. a node contract deployed on the node represents the workload deployment itself. and another name contract reserving the name on the chain.

this happens on the name gateway. but not in the fqdn

checking the database records, there is no single contract with fqdn 

### which network/s did you face the problem on?

Dev

### version

de0a62b

### Relevant screenshots/screen records

![image](https://github.com/user-attachments/assets/fefe7572-5686-4d1e-bf9c-f020c62389f8)

### Relevant log output

```console
nothing
```

zaelgohary:
Store FQDN in name contracts.

samaradel:
update:
emit and create FQDN contract

samaradel:
update:
Create a function to handle the gateway workloads

samaradel:
update:
- Use the actual name from workload data
- Changing the reassignment operations to push operations

samaradel:
update:
Create a function to handle the name of contracts

0oM4R:
we need to consider this 
> If we can't store the FQDN as is, there will be many variations of domains that will end up with the same string after removing `.`, `-`, and `_`. So, no unique records, which is the target of having the name contract.

_Originally posted by @AhmedHanafy725 in https://github.com/threefoldtech/tfgrid-sdk-ts/pull/4317#discussion_r2265177153_

PR #4317, opened 2025-07-29
Emit and create contract for FQDN

samaradel:
### Description

FQDN gateways will create name contracts using the workload name (which is alphanumeric) instead of the FQDN (which contains invalid characters)

### Changes

image

image

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/3391

### Tested Scenarios

- Deploy FQDN and check the FQDN contract if it existed

### Documentation PR

#### Preliminary Checks:

- [x] Preliminary Checks
  - What about edge cases?

#### UI Checks:

- [x] UI Checks
  - Does every button work?
  - Does every validation work?

#### Code Quality Checks:

- [x] Code Quality Checks

### Testing Checklist

### General Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation

### DIFF
--- a/packages/grid_client/src/high_level/twinDeploymentHandler.ts
+++ b/packages/grid_client/src/high_level/twinDeploymentHandler.ts
@@ -412,6 +412,42 @@ class TwinDeploymentHandler {
     return await this.deploymentFactory.fromObj(deployment);
   }
 
+  private async handleGatewayWorkloads(
+    workloads: Workload[],
+    operation: Operations,
+  ): Promise<{
+    nameExtrinsics: ExtrinsicResult<Contract>[];
+    deletedExtrinsics: ExtrinsicResult<number>[];
+  }> {
+    const nameExtrinsics: ExtrinsicResult<Contract>[] = [];
+    const deletedExtrinsics: ExtrinsicResult<number>[] = [];
+
+    for (const workload of workloads) {
+      if (workload.type === WorkloadTypes.gatewaynameproxy) {
+        events.emit("logs", `Check the name contract for the workload with name: ${workload.name}`);
+        if (operation === Operations.delete) {
+          const extrinsic = await this.deleteNameContract(workload.data["name"]);
+          if (extrinsic) deletedExtrinsics.push(extrinsic);
+        } else {
+          const extrinsic = await this.createNameContract(workload.data["name"]);
+          nameExtrinsics.push(extrinsic);
+        }
+      } else if (workload.type === WorkloadTypes.gatewayfqdnproxy) {
+        events.emit("logs", `Check the name contract for the FQDN workload with name: ${workload.name}`);
+        const contractName = workload.name.replace(/[^a-zA-Z0-9]/g, "").toLowerCase();
+        if (operation === Operations.delete) {
+          const extrinsic = await this.deleteNameContract(contractName);
+          if (extrinsic) deletedExtrinsics.push(extrinsic);
+        } else {
+          const extrinsic = await this.createNameContract(contractName);
+          nameExtrinsics.push(extrinsic);
+        }
+      }
+    }
+
+    return { nameExtrinsics, deletedExtrinsics };
+  }
+
   async rollback(contracts) {
     // cancel all created contracts and leave the updated ones.
     events.emit("logs", "Rolling back deployments");
@@ -465,23 +501,14 @@ class TwinDeploymentHandler {
 
   async PrepareExtrinsic(twinDeployment: TwinDeployment, contracts) {
     const nodeExtrinsics: ExtrinsicResult<Contract>[] = [];
-    const nameExtrinsics: ExtrinsicResult<Contract>[] = [];
-    const deletedExtrinsics: ExtrinsicResult<number>[] = [];
+    let nameExtrinsics: ExtrinsicResult<Contract>[] = [];
+    let deletedExtrinsics: ExtrinsicResult<number>[] = [];
+
     if (twinDeployment.operation === Operations.deploy) {
       events.emit("logs", `Deploying on node_id: ${twinDeployment.nodeId}`);
-      for (const workload of twinDeployment.deployment.workloads) {
-        // check if the deployment need name contract
-        if (workload.type === WorkloadTypes.gatewaynameproxy) {
-          events.emit("logs", `Check the name contract for the workload with name: ${workload.name}`);
-          const extrinsic = await this.createNameContract(workload.data["name"]);
-          nameExtrinsics.push(extrinsic);
-        } else if (workload.type === WorkloadTypes.gatewayfqdnproxy) {
-          events.emit("logs", `Check the name contract for the FQDN workload with name: ${workload.name}`);
-          const contractName = workload.name.replace(/[^a-zA-Z0-9]/g, "").toLowerCase();
-          const extrinsic = await this.createNameContract(contractName);
-          nameExtrinsics.push(extrinsic);
-        }
-      }
+      const gatewayResults = await this.handleGatewayWorkloads(twinDeployment.deployment.workloads, Operations.deploy);
+      nameExtrinsics = gatewayResults.nameExtrinsics;
+
       const extrinsic = await this.tfclient.contracts.createNode({
         hash: twinDeployment.deployment.challenge_hash(),
         data: twinDeployment.metadata,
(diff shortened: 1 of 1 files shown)


packages/grid_client/src/high_level/twinDeploymentHandler.ts
```diff
@@ -492,6 +497,11 @@ class TwinDeploymentHandler {
           events.emit("logs", `Check the name contract for the workload with name: ${workload.name}`);
           const extrinsic = await this.createNameContract(workload.data["name"]);
           nameExtrinsics.push(extrinsic);
+        } else if (workload.type === WorkloadTypes.gatewayfqdnproxy) {
+          events.emit("logs", `Check the name contract for the FQDN workload with name: ${workload.name}`);
+          const contractName = workload.name.replace(/[^a-zA-Z0-9]/g, "").toLowerCase();
```
0oM4R: i suggest having this logic in a function and reusing it 
to avoid code redundancy, andit  will be easier to maintain

packages/grid_client/src/high_level/twinDeploymentHandler.ts
```diff
@@ -491,19 +518,9 @@ class TwinDeploymentHandler {
...
-          nameExtrinsics.push(extrinsic);
-        } else if (workload.type === WorkloadTypes.gatewayfqdnproxy) {
-          events.emit("logs", `Check the name contract for the FQDN workload with name: ${workload.name}`);
-          const contractName = workload.name.replace(/[^a-zA-Z0-9]/g, "").toLowerCase();
-          const extrinsic = await this.createNameContract(contractName);
-          nameExtrinsics.push(extrinsic);
-        }
-      }
+      const gatewayResults = await this.handleGatewayWorkloads(twinDeployment.deployment.workloads, Operations.update);
+      nameExtrinsics = gatewayResults.nameExtrinsics;
```
0oM4R: We should push the result, not reassign it. 
also we can have 
```ts
const gatewayResults = await this.handleGatewayWorkloads(twinDeployment.deployment.workloads, twinDeployment.operation)
```
then insdie the operation check block:
```ts

else if (twinDeployment.operation === Operations.update) {
...
nameExtrinsics.push(gatewayResults.nameExtrinsics)
}
```
same on delete

0oM4R (changes requested):

packages/grid_client/src/high_level/twinDeploymentHandler.ts
```diff
@@ -412,6 +412,42 @@ class TwinDeploymentHandler {
...
+        if (operation === Operations.delete) {
+          const extrinsic = await this.deleteNameContract(workload.data["name"]);
+          if (extrinsic) deletedExtrinsics.push(extrinsic);
+        } else {
+          const extrinsic = await this.createNameContract(workload.data["name"]);
+          nameExtrinsics.push(extrinsic);
+        }
+      } else if (workload.type === WorkloadTypes.gatewayfqdnproxy) {
+        events.emit("logs", `Check the name contract for the FQDN workload with name: ${workload.name}`);
+        const contractName = workload.name.replace(/[^a-zA-Z0-9]/g, "").toLowerCase();
```
AhmedHanafy725: That's wrong. Why should the workload name be a unique identifier?! The workload name should be unique on the deployment object only.  The unique identifier that should be on the name contract is the FQDN.

AhmedHanafy725 (changes requested):

ramezsaeed:
Test scenarios:
1. Workload name with alphanumeric characters
    - Given a workload named testGateway42
    - When a gateway is created
    - Then the name contract should exactly match testGateway42
    - And contain no special characters from the FQDN
    
2. FQDN with invalid characters
    -  Given a FQDN api.dev-company.org
    - And a workload name apiDevCompany
    - When an FQDN gateway is created
    - Then the contract name should be apiDevCompany
    - And not contain ., -, or any other invalid characters
    
3. Contract name uniqueness
    - Given two gateways with same workload name which have speciall characters my_domain$1
    - When both are deployed
    - Then the system should reject the second contract
    - Or append a unique suffix to prevent a name conflict

packages/grid_client/src/high_level/twinDeploymentHandler.ts
```diff
@@ -412,6 +412,42 @@ class TwinDeploymentHandler {
...
+  private async handleGatewayWorkloads(
+    workloads: Workload[],
+    operation: Operations,
+  ): Promise<{
+    nameExtrinsics: ExtrinsicResult<Contract>[];
+    deletedExtrinsics: ExtrinsicResult<number>[];
+  }> {
+    const nameExtrinsics: ExtrinsicResult<Contract>[] = [];
+    const deletedExtrinsics: ExtrinsicResult<number>[] = [];
+
```
0oM4R, line 424: We can add a function here to reduce code redundancy, so you can define the contract name, based on workload type, then pass it to this function, whatever the operation is 
```ts
  const processContract = async (contractName: string) => {
    if (operation === Operations.delete) {
      const extrinsic = await this.deleteNameContract(contractName);
      if (extrinsic) deletedExtrinsics.push(extrinsic);
    } else {
      const extrinsic = await this.createNameContract(contractName);
      nameExtrinsics.push(extrinsic);
    }
  };
```

0oM4R (changes requested):
created a fqdn with fqdn `kassem-test.elgrid.xyz`
this created 
gateway contract 
details 

```json
{
  "version": 0,
  "twin_id": 800,
  "contract_id": 225610,
  "metadata": "",
  "description": "",
  "expiration": 0,
  "signature_requirement": {
    "requests": [
      {
        "twin_id": 800,
        "required": false,
        "weight": 1
      }
    ],
    "weight_required": 1,
    "signatures": [
      {
        "twin_id": 800,
        "signature": "1aa73655d7bc6fdb6e6488fbac42365d6cc9155af79abf668adc1d9d4ccffb529419859e37cd52375511959af546a2c471ec3bca8578a9776d37094ccdf5d881",
        "signature_type": "sr25519"
      }
    ],
    "signature_style": ""
  },
  "workloads": [
    {
      "version": 0,
      "name": "kassem8",
      "type": "gateway-fqdn-proxy",
      "data": {
        "__type": "gateway-fqdn-proxy",
        "fqdn": "kassem-test.elgrid.xyz",
        "tls_passthrough": false,
        "backends": [
          "http://10.20.2.2:80"
        ],
        "network": "nwcnn2q"
      },
      "metadata": "",
      "description": "",
      "result": {
        "created": 1754566136,
        "state": "ok",
        "message": "",
        "data": {}
      }
    }
  ]
}
```

and name contract 
```json
{
  "contract_id": 225611,
  "twin_id": 800,
  "type": "name",
  "state": "Created",
  "created_at": 1754566128,
  "details": {
    "nodeId": "-",
    "farm_id": "-"
  },
  "solutionName": "kassemtestelgridxyz",
  "solutionType": "-",
  "expiration": "-",
  "consumption": 0,
  "discountPackage": "None"
}
```
then tried to create new fqdn pointing to `kassem.test.elgrid.xyz`
image


image

packages/grid_client/src/high_level/twinDeploymentHandler.ts
```diff
@@ -412,6 +412,42 @@ class TwinDeploymentHandler {
...
+        if (operation === Operations.delete) {
+          const extrinsic = await this.deleteNameContract(workload.data["name"]);
+          if (extrinsic) deletedExtrinsics.push(extrinsic);
+        } else {
+          const extrinsic = await this.createNameContract(workload.data["name"]);
+          nameExtrinsics.push(extrinsic);
+        }
+      } else if (workload.type === WorkloadTypes.gatewayfqdnproxy) {
+        events.emit("logs", `Check the name contract for the FQDN workload with name: ${workload.name}`);
+        const contractName = workload.data["fqdn"].replace(/[^a-zA-Z0-9]/g, "").toLowerCase();
```
0oM4R, line 432: we should store the full fqdn name and do not remove anything

samaradel: I think we shouldn't as tf chain doesn't support the characters like dots for example

AhmedHanafy725: If we can't store the FQDN as is, there will be many variations of domains that will end up with the same string after removing `.`, `-`, and `_`. So, no unique records, which is the target of having the name contract.

samaradel: so, any recommended ideas?

0oM4R:
> If we can't store the FQDN as is, there will be many variations of domains that will end up with the same string after removing ., -, and _. So, no unique records, which is the target of having the name contract.

will close this pr as we need to decide how to fix this blocker

Issue #4109, opened 2025-05-11
🐞 [Bug]: Email field not cleared or enabled/disabled correctly when mnemonic is removed


khaledyoussef24:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### which package/s did you face the problem with?

Dashboard

### What happened?

When you paste a valid mnemonic into the TFChain “Connect Your Wallet” flow the dashboard auto-loads your email. If you then delete the mnemonic:

The email list remains visible even though the input is disabled.

If you paste a new (invalid or empty) mnemonic, the old emails are cleared, the email field becomes enabled and immediately shows “Email is required” before you even touch it.

https://github.com/threefoldtech/tfgrid-sdk-ts/issues/3805

### Steps To Reproduce

1. Open staging.dashboard.dev.grid.tf → Connect Your TFChain Wallet
2. Paste a valid mnemonic 
3. emails will load below
4. Delete the mnemonic entirely 
5. loaded emails remain, but the email input is disabled
6. Paste a new (empty/invalid) mnemonic or click “Create Account” 
7. emails clear, email input becomes enabled, and “Email is required” immediately shows

### which network/s did you face the problem on?

Dev

### version

Version 2.8.0-rc2

### Relevant screenshots/screen records

![Image](https://github.com/user-attachments/assets/127195db-10aa-40f1-97b7-1845d6ef94ff)
![Image](https://github.com/user-attachments/assets/7b9ceb65-4d6d-4e48-acf9-83e96f6219d8)

### Relevant log output

```console
.
```

samaradel:
update:
fix email input behavior

samaradel:
update:
- Fix Email input validation
- Disable all input if there is no mnemonic
- Prevent email input loading if it is a new account

samaradel:
update:
Enable email field if the mnemonic is valid

samaradel:
update:
- Remove email checks 
- Handle the validation of the mnemonic input directly on the mnemonic field

0oM4R:
Introduced a simpler pr that will handle the mentioned core issue only, if the mnemonic changed ,the email should be removed

PR #4327, opened 2025-08-03
Fix email loading

samaradel:
### Description

- The email input was disabled whenever `mnemonicInput?.status !== ValidatorStatus.Valid`, which included when there was no mnemonic at all
- The validation rules were always applied, even when the field should be disabled
- When the mnemonic was cleared, the email state wasn't properly reset

### Changes

[Screencast from 03-08-25 14:49:20.webm](https://github.com/user-attachments/assets/3e3742b7-1acd-4e48-b732-00ac091402f4)

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/4109 

### Tested Scenarios

- When a valid mnemonic: email populated, field disabled, validation disabled
- When mnemonic is deleted: Email cleared, field enabled, validation enabled but not triggered
- When an invalid mnemonic is pasted: Email cleared, field enabled, validation enabled but not triggered
- Check if the user interacts with the enabled email field: Normal validation behavior resumes

### Documentation PR

#### Preliminary Checks:

- [x] Preliminary Checks
  - What about edge cases?

#### UI Checks:

- [x] UI Checks
  - Does every button work?
  - Does every validation work?

#### Code Quality Checks:

- [x] Code Quality Checks

### Testing Checklist

### General Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation

### DIFF
--- a/packages/playground/src/components/profile_manager/ConnectWallet.vue
+++ b/packages/playground/src/components/profile_manager/ConnectWallet.vue
@@ -109,11 +109,14 @@
       <!-- Email -->
       <input-validator
         id="email-validator"
+        ref="emailInput"
         :value="email"
-        :rules="[
-          validators.required('Email is required.'),
-          validators.isEmail('Please provide a valid email address.'),
-        ]"
+        :rules="
+          mnemonic && mnemonicInput?.status === ValidatorStatus.Valid
+            ? [validators.required('Email is required.'), validators.isEmail('Please provide a valid email address.')]
+            : []
+        "
+        :disable-validation="!!isEmailDisabled"
         #="{ props }"
       >
         <v-text-field
@@ -124,7 +127,7 @@
           label="Email"
           placeholder="email@example.com"
           :loading="loadEmail"
-          :disabled="creatingAccount || connecting || loadEmail || mnemonicInput?.status !== ValidatorStatus.Valid"
+          :disabled="isEmailDisabled"
           autocomplete="off"
         />
       </input-validator>
@@ -191,7 +194,7 @@ import { TwinNotExistError } from "@threefold/types";
 import { validateMnemonic } from "bip39";
 import Cryptr from "cryptr";
 import md5 from "md5";
-import { ref, watch } from "vue";
+import { ref, watch, nextTick, computed } from "vue";
 
 import { ValidatorStatus } from "@/hooks/form_validator";
 import { useInputRef } from "@/hooks/input_validator";
@@ -209,6 +212,17 @@ const keypairType = ref(KeypairType.sr25519);
 const keyType = ["sr25519", "ed25519"];
 const confirmPasswordInput = useInputRef();
 const mnemonicInput = useInputRef();
+const emailInput = useInputRef();
+
+const isEmailDisabled = computed(() => {
+  return (
+    creatingAccount.value ||
+    connecting.value ||
+    loadEmail.value ||
+    (mnemonic.value && mnemonicInput.value?.status !== ValidatorStatus.Valid)
+  );
+});
+
 // loading
 const loadEmail = ref(false);
 const creatingAccount = ref(false);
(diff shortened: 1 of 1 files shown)


0oM4R (commented):
after creating a new account without touching the email account the email validation got triggered, also after adding mnemonic the email will try to load stored email, if none will show the error, 
image

packages/playground/src/components/profile_manager/ConnectWallet.vue
```diff
@@ -318,7 +318,29 @@ const validateMnemonicInput = async (input: string) => {
...
+    // Check if twin exists first
+    try {
+      await grid!.twins.get_my_twin_id();
+      // Twin exists, so get email with loading state
+      loadEmail.value = true;
+      try {
+        getEmail(grid!);
+      } finally {
+        loadEmail.value = false;
+      }
```
0oM4R, line 322: why did you remove the load email handler from getEmail? 
also the old logic was handle getEmail errors, so no need to handle it here in nested try

0oM4R (changes requested):
After adding a non-active mnemonic, the email field is disabled and can't proceed
image

packages/playground/src/components/profile_manager/ConnectWallet.vue
```diff
@@ -124,16 +125,18 @@
           label="Email"
           placeholder="email@example.com"
           :loading="loadEmail"
-          :disabled="creatingAccount || connecting || loadEmail || mnemonicInput?.status !== ValidatorStatus.Valid"
+          :disabled="isEmailDisabled"
           autocomplete="off"
+          @blur="() => emailInput?.validate()"
+          @update:model-value="() => emailInput?.validate()"
```
0oM4R, line 138: I think this is the actual behavior of the input validator. Why do we need to specify that?

samaradel: After handling the disabled part, it doesn't work, so I handled it this way

packages/playground/src/components/profile_manager/ConnectWallet.vue
samaradel: Well, it was fire even though the mnemonic is new (new account), which is useless

0oM4R: image

i don't think it was firing any errors. 
also in your new code `await grid!.twins.get_my_twin_id();` if this failed so it will not trigger load email as it will go to the catch block

packages/playground/src/components/profile_manager/ConnectWallet.vue
```diff
@@ -124,16 +125,18 @@
...
         />
       </input-validator>
       <!-- Passwords -->
       <WalletPassword
         id="password-input"
         v-model="password"
         mode="Create"
-        :disabled="creatingAccount || connecting"
+        :disabled="creatingAccount || connecting || !isMnemonicValid"
```
0oM4R, line 146: Why do we disable those if the mnemonic is empty? 
The idea behind disabling the email field is to avoid overwriting the entered email after getting the stored email. This case is on password fields, so I don't think we should disable them

packages/playground/src/components/profile_manager/ConnectWallet.vue
```diff
@@ -152,7 +155,7 @@
               ...confirmPasswordInputProps,
               ...validationProps,
             }"
-            :disabled="creatingAccount || connecting"
+            :disabled="creatingAccount || connecting || !isMnemonicValid"
```
0oM4R, line 165: please check https://github.com/threefoldtech/tfgrid-sdk-ts/pull/4327/files#r2265191703

packages/playground/src/components/profile_manager/ConnectWallet.vue
```diff
@@ -263,27 +290,65 @@ function reloadValidation() {
...
+
+  if (!input) {
+    email.value = "";
+    if (emailInput.value) {
+      emailInput.value.reset();
+    }
+    return {
+      message: "Mnemonic or Hex Seed is required.",
+    };
+  }
```
0oM4R: Checking if the mnemonic is empty is done by the required validator, I think we will never reach this

0oM4R (commented):
I think there are many unnecessary changes, we can add to mnemonic field somthing like 
` @update:model-value="() => email ? (email = '') : null"` 
and will reset the email if the mnemonic changed 
we only may need to handle the validation of email in this case

packages/playground/src/components/profile_manager/ConnectWallet.vue
samaradel: if we don't have a valid mnemonic, why do we need it enabled? It's useless

ramezsaeed:
You can also have a look here for more testing scenarios about email.
https://github.com/threefoldtech/tfgrid-sdk-ts/blob/development/packages/playground/tests/frontend_selenium/tests/TFChain/test_homepage.py

0oM4R:
[fix: clear email input and validation when mnemonic changes #4382](https://github.com/threefoldtech/tfgrid-sdk-ts/pull/4382)
handled here

Issue #4246, opened 2025-06-22
🐞 [Bug]: casperlabs stucks after  deploying


khaledyoussef24:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### which package/s did you face the problem with?

Dashboard

### What happened?

After deploying Casperlabs, the services are all telling  "checking" and "processing."
produced with 3 deployments
followed the docs

<img width="1265" alt="Image" src="https://github.com/user-attachments/assets/74eaf438-2f89-4eca-8695-ec12338e6016" />

### which network/s did you face the problem on?

Main

### version

2.8

### Relevant screenshots/screen records

.

### Relevant log output

```console
.
```

zaelgohary:
Investigation:

I found that Casperlabs service in the VM is not running, and the logs showed that it had an error while key gen step. Fixed by added a condition before key gen, but faced other issues regarding Casperlabs version & missing dependencies.

zaelgohary:
I could fix the dependency issues by upgrading Ubuntu & Casper version within the VM.

<img width="1779" height="849" alt="Image" src="https://github.com/user-attachments/assets/c3168a4d-eb7f-46aa-bdda-207168a4c6e4" />

Currently working on fixing the flist.

zaelgohary:
Work Completed:

Updated the flist with above changes and created a [PR](https://github.com/threefoldtech/tf-images/pull/367) in tf-images. Applied [hardeware requirements](https://docs.casper.network/operators/setup/hardware) in solution.

zaelgohary:
I fixed the node log file issue as the file path was not created on the machine, and the logs were empty in the dashboard. Additionally, I worked on resolving the node sync issue; however, I couldn't test it properly as there were no available nodes on the dev or QA environments due to the new hardware requirements. Tested a bit on the mainnet but faced connection issues as seen in the screenshot below. Will continue testing today. Opened an [issue](https://git.ourworld.tf/tfgrid_internal/circle_tfgrid_ops/issues/540) to fund main acc.

<img width="930" height="244" alt="Image" src="https://github.com/user-attachments/assets/ea7eb20f-c99d-49d7-8d29-385c1645482c" />

zaelgohary:
The fast node sync is not working due to an empty trust hash. In this case, node sync will be from genesis, which is very slow and could take days. I tried multiple solutions to get a working, trusted hash, but all failed. I also tried multiple validator IPs but all were unreachable.

[...]

PR #4394, opened 2025-10-19
Update flist & casper solution flavours

zaelgohary:
### Description

- Update Casper solution falcours as per [Hardware Requirements](https://docs.casper.network/operators/setup/hardware)
- Edit flist

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/4246

### DIFF
--- a/packages/grid_client/src/helpers/flists.ts
+++ b/packages/grid_client/src/helpers/flists.ts
@@ -85,7 +85,7 @@ export const FLISTS = {
     entryPoint: "/sbin/zinit init",
   },
   CASPERLABS: {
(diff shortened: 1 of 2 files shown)


packages/grid_client/src/helpers/flists.ts
```diff
@@ -85,7 +85,7 @@ export const FLISTS = {
     entryPoint: "/sbin/zinit init",
   },
   CASPERLABS: {
-    value: "https://hub.grid.tf/tf-official-apps/casperlabs-latest.flist",
+    value: "https://hub.grid.tf/zaelgohary.3bot/zaelgohary-casper-ubuntu22-latest.flist",
```
0oM4R, line 88: we should have the official image here

PR #4395, opened 2025-10-19
Fix deps optimization warnings and include terser compressor

0oM4R:
On a fresh build, we may have some runtime warnings; it works after a refresh. This is due to the new Vite version. It is a known issue on Vite, so we exclude the complex deps to avoid this error 

also added terser copressor

### DIFF
--- a/packages/playground/package.json
+++ b/packages/playground/package.json
@@ -70,6 +70,7 @@
     "npm-run-all": "^4.1.5",
     "sass": "^1.87.0",
     "start-server-and-test": "^2.0.11",
+    "terser": "^5.44.0",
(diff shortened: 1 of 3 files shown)


zaelgohary (changes requested):
Should Update the lock file too.

zaelgohary:
Also, I still get the warnings.

image

0oM4R:
> Also, I still get the warnings.
> 
> image

This is not related, the pr is for the yarn dev command, the warnings are related to eslint not the build

zaelgohary:
@0oM4R Please fix the security & codacy checks.

Issue #4381, opened 2025-10-13
🐞 [Bug]: Workers contracts are listed with the cluster name


0oM4R:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### which package/s did you face the problem with?

Dashboard

### What happened?

<img width="866" height="144" alt="Image" src="https://github.com/user-attachments/assets/e32d02a0-1935-46e6-a922-512f966b6771" />

<img width="1567" height="802" alt="Image" src="https://github.com/user-attachments/assets/6ab1bc18-ce6d-4934-85d6-275b8abb9bd0" />

### which network/s did you face the problem on?

Dev

### version

2.8.2

### Relevant screenshots/screen records

.

### Relevant log output

```console
.
```

PR #4397, opened 2025-10-20
Fix worker name metadata in create deployment & add worker

zaelgohary:
### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/4381

image


0oM4R:
tried to deploy a cluster with two workers, but the workers never got listed, even on the deployemnt data dialog after deploy

AhmedHanafy725 (changes requested):
We use the cluster's name on the metadata, not the workers' or masters' names. This will break the compatibility with the other clients.

Issue #2269, opened 2024-02-27
inconsistency in adding disks parameters for FullVM and MiroVM


ramezsaeed:
### Description
When trying to add disk for FullVm or MIcroVM, the parameters are inconsistent.

A clear description of the bug:

### 1- MicroVM
    - when add disk it shows only two fields (Disk name, Disk size)
    - as per discussion with Hanafy we are missing here the disk mount point too (as in micro vm case the ZOS can us and mount this mount point)
    - when I update the field for disk name it appears after deployments as the disk mount point and the name i updated doesn't show at all.

![Screenshot from 2024-02-27 13-38-31](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/16083592/bac6d6b0-cbe0-499e-ab79-1f1359dfe8ee)
![Screenshot from 2024-02-27 13-18-23](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/16083592/eda5704b-5e9e-42f7-b50d-ce4ba4904cf5)

### 2- FullVM
    - when I update the field for disk name it appears after deployments correctly
![Screenshot from 2024-02-27 14-01-44](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/16083592/ed0c9164-c26c-4574-9a16-cce0f4b089c1)

Must include any relevant identifiers like:

- **Network**: QAnet

- **Version**: 2.3.0-alpha12

- **Twin ID**: 463

zaelgohary:
Work Completed:

Micro VM had the mounting point as seen in the screenshot below.

<img width="1495" height="515" alt="Image" src="https://github.com/user-attachments/assets/93b45489-b00f-4b54-b3eb-df678b01f12f" />

I just unified the disk label to show disk name only. User can always check for mounting point in the json tab.

PR #4398, opened 2025-10-20
Unify disk label in full & micro vm

zaelgohary:
### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/2269

image

### DIFF
--- a/packages/playground/src/components/deployment_data_dialog.vue
+++ b/packages/playground/src/components/deployment_data_dialog.vue
@@ -71,7 +71,7 @@
               <CopyReadonlyInput
                 v-for="disk of contract.mounts"
                 :key="disk.name"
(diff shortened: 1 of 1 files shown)


packages/playground/src/components/deployment_data_dialog.vue
```diff
@@ -71,7 +71,7 @@
               <CopyReadonlyInput
                 v-for="disk of contract.mounts"
                 :key="disk.name"
-                :label="getDiskLabel(contract, disk)"
+                :label="getDiskLabel(disk)"
```
0oM4R:
```suggestion
                :label="`Disk ${disk.name}`"
```
no need to have ea function for this

Issue #4326, opened 2025-08-03
🐞 [Bug]: Farm Name Filter Field Resets and Placeholder Persists When Typing


atefnazmy:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### which package/s did you face the problem with?

Dashboard

### What happened?

- When attempting to filter nodes by farm name in the Nodes Finder, the input field resets after each character is entered.
- The placeholder text reappears and cannot be removed when trying to type or clear the field

### Steps To Reproduce

-  Navigate to https://staging.dashboard.dev.grid.tf/
- Click on the Deploy icon in the sidebar
- Expand the Deploy option and click on Nodes Finder
- Select the Farm Name filter
- Try to type any farm name (each character input causes the field to reset)

### which network/s did you face the problem on?

Dev

### version

721cde5

### Relevant screenshots/screen records

https://github.com/user-attachments/assets/c7df1b41-e2c7-47fb-b4e2-746b861dda0e

### Relevant log output

```console
n/a
```

PR #4399, opened 2025-10-22
Fix farm filter

zaelgohary:
### Description

VAutocomplete's v-model:search two-way binding was causing circular updates. VAutocomplete internally syncs the search input with the selected item's name ("All Farms"), which overwrites user input after each keystroke.

### Changes

- Changed from 2-way binding to controlled search binding
- Disabled built-in filtering which was interfering w our search
- Added isClearing flag
- Added new handleSearchUpdate w guards
- Edited onClear
- Updated farms computed property to match country filter behavior

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/4326

### DIFF
--- a/packages/playground/src/components/node_selector/TfSelectFarm.vue
+++ b/packages/playground/src/components/node_selector/TfSelectFarm.vue
@@ -2,7 +2,7 @@
   <input-tooltip :tooltip="$props.tooltip as string" :disabled="insetTooltip">
     <VAutocomplete
       v-model:menu="menuOpened"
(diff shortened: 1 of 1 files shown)


0oM4R (dismissed):

0oM4R:
[Screencast from 2025-10-22 14-22-59.webm](https://github.com/user-attachments/assets/8b42c1b1-e3ef-4917-ba04-8dd1619c8f37)

On clear, the old value is kept I think the value should be cleared as well

xmonader (approved):

Issue #4017, opened 2025-04-03
🐞 [Bug]: sometimes FullVm instance is listed in the microVm deployment page


ramezsaeed:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### which package/s did you face the problem with?

Dashboard

### What happened?

sometimes the fullVM instances is listed in MicroVm deployment list and vise versa.

### Steps To Reproduce

1- go to FullVM page
2- deploy a full VM
3- check its listed in deployment table
4- go to microVM page
5- will find the full VM listed in deployment table

### which network/s did you face the problem on?

Main

### version

2.7.0

### Relevant screenshots/screen records

![Image](https://github.com/user-attachments/assets/144b06eb-e8d5-4075-8596-4dac97963819)
![Image](https://github.com/user-attachments/assets/0f37ee17-3f7e-45d7-9f72-33e705fc2b09)

### Relevant log output

```console
--
```

amiraabouhadid:
was able to reproduce 

<img width="1202" alt="Image" src="https://github.com/user-attachments/assets/f1aa34e5-8bed-4add-a555-7cbb936ddf5d" />

<img width="1183" alt="Image" src="https://github.com/user-attachments/assets/275beb42-4dfd-433c-b125-7ebda4b09724" />

amiraabouhadid:
it only happens on vm creation, if you reload the micro vm page, you'll find that the vm deployment is no longer listed. Investigating if issue is from the grid client 
issue isn't in deploymentListEnvironment, nor in projectName setting in vm_deployment_table but in the chunks returned and merged in loadDeployments

amiraabouhadid:
Revised loadVms logic, looking for concise alternative to chunks approach

amiraabouhadid:
Tried rewriting loadVms and loadDeployments logic but failed, a more straightforward option would be to change Micro Vm's deployment name prefix to 'mvm' or 'mm'  instead of the current 'vm', which is the main cause of the bug as it loads deployments based on prefix, and both Full Vm and Micro Vm share the 'vm' prefix. What do you think @AhmedHanafy725 ?

amiraabouhadid:
two issues: loadDeployment is called multiple times somewhere and the show Deployments btn is active even when it has not been toggled

amiraabouhadid:
[...]

PR #4420, opened 2025-11-03
fix: prevent fullvm deployments from appearing in micro vm list

zaelgohary:
### Changes

- Use OR pattern matching for "projectName":"vm/" or "projectName":"vm"}"
instead of substring contains to prevent false matches where "fullvm"
was incorrectly matched when filtering for "vm" deployments.

### Related Issues
- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/4017

### DIFF
--- a/packages/grid_client/src/clients/tf-grid/contracts.ts
+++ b/packages/grid_client/src/clients/tf-grid/contracts.ts
@@ -216,7 +216,12 @@ class TFContracts extends Contracts {
       }
 
       if (options.projectName) {
-        filterQuery += `{ deploymentData_contains: \"\\\"projectName\\\":\\\"${options.projectName}\" }`;
+        // Match projectName with slash (new format) OR closing brace (old format)
+        // Prevents substring false positives: "vm" won't match "fullvm" or misnamed contracts
+        filterQuery += `{ OR: [
+          { deploymentData_contains: \"\\\"projectName\\\":\\\"${options.projectName}/\" },
+          { deploymentData_contains: \"\\\"projectName\\\":\\\"${options.projectName}\\\"}\" }
+        ]}`;
       }
 
       filterQuery += "]";
--- a/packages/grid_client/src/clients/tf-grid/contracts.ts
+++ b/packages/grid_client/src/clients/tf-grid/contracts.ts
@@ -216,12 +216,7 @@ class TFContracts extends Contracts {
       }
 
       if (options.projectName) {
-        // Match projectName with slash (new format) OR closing brace (old format)
-        // Prevents substring false positives: "vm" won't match "fullvm" or misnamed contracts
-        filterQuery += `{ OR: [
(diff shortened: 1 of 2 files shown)


packages/grid_client/src/clients/tf-grid/contracts.ts
```diff
@@ -216,7 +216,12 @@ class TFContracts extends Contracts {
       }
 
       if (options.projectName) {
-        filterQuery += `{ deploymentData_contains: \"\\\"projectName\\\":\\\"${options.projectName}\" }`;
+        // Match projectName with slash (new format) OR closing brace (old format)
+        // Prevents substring false positives: "vm" won't match "fullvm" or misnamed contracts
+        filterQuery += `{ OR: [
+          { deploymentData_contains: \"\\\"projectName\\\":\\\"${options.projectName}/\" },
+          { deploymentData_contains: \"\\\"projectName\\\":\\\"${options.projectName}\\\"}\" }
+        ]}`;
```
AhmedHanafy725: no need for this change as it match with `"projectName": "vm"` not just `vm`

packages/grid_client/src/modules/base.ts
```diff
@@ -270,8 +270,12 @@ class BaseModule {
       const alreadyFetchedContracts: GqlNodeContract[] = [];
 
       for (const contract of BaseModule.newContracts) {
-        if (!contract.parsedDeploymentData?.projectName.includes(this.projectName)) continue;
-        if (contract.parsedDeploymentData.type !== moduleName) continue;
+        if (this.projectName) {
+          const pName = contract.parsedDeploymentData?.projectName;
+          // Match exact OR prefix with slash (prevents "vm" from matching "fullvm")
+          if (pName !== this.projectName && !pName?.startsWith(this.projectName + "/")) continue;
```
AhmedHanafy725: the `pName` on the contract will be "vm/{name}" so the exact match won't list it at all.

AhmedHanafy725 (changes requested):

packages/grid_client/src/modules/base.ts
```diff
@@ -270,8 +270,10 @@ class BaseModule {
       const alreadyFetchedContracts: GqlNodeContract[] = [];
 
       for (const contract of BaseModule.newContracts) {
-        if (!contract.parsedDeploymentData?.projectName.includes(this.projectName)) continue;
-        if (contract.parsedDeploymentData.type !== moduleName) continue;
+        if (this.projectName) {
+          if (!contract.parsedDeploymentData?.projectName?.startsWith(this.projectName + "/")) continue;
```
AhmedHanafy725:
```suggestion
          if (!contract.parsedDeploymentData?.projectName?.startsWith(this.projectName)) continue;
```

AhmedHanafy725 (approved):

Issue #4340, opened 2025-08-13
TFT price inconsistencies


sameh-farouk:
https://github.com/threefoldtech/tfgrid-sdk-ts/blob/4b0f59bda264e5396548f81d936f81611be7ad4d/packages/grid_client/src/clients/tf-grid/contracts.ts#L383-L396

- The conversion process utilizes the tfchain-client to obtain the current, rounded market price.  
- This information is used for cost estimation purposes.  
- However, actual on-chain billing applies a different TFT price, as it takes into account both minimum and maximum configured values (currently set at 10 mUSD to 1000 mUSD), which can be queried from the tfchain. 

The conversion should instead query `MinTftPrice`, `MaxTftPrice`, and `AverageTftPrice`, then clamp `AverageTftPrice` to the configured bounds.

For more context, see also:
- https://github.com/threefoldtech/tfgrid-sdk-go/issues/1410

zaelgohary:
Work Completed:

Implemented `getTFTBillingRate` in tfchain client and applied changes in grid client & playground. However, I've been notified that changes shouldn't be applied everywhere. Do you have any idea where and where not `getTFTBillingRate` should be used? And if not used everywhere, wouldn't this create another inconsistency if users noticed the different rates? @AhmedHanafy725

sameh-farouk:
> Do you have any idea where and where not getTFTBillingRate should be used?

TFT uses two different rates, each serving its own purpose. By understanding each rate's purpose, it would be easier to reason about where to use each one:  

1. Market Rate
 • This is the real trading value of TFT on exchanges (determined primarily by the interaction of supply and demand).
 • It’s used for buying, selling, swapping, or showing TFT’s value in fiat.
 • Anything financial or market-oriented relies on this rate.

2. On-Chain Billing Rate (Bounded Rate)
 • Used internally inside TFChain to calculate the cost of resources and contracts.
 • It’s still derived from the market rate, but it’s kept within council-defined minimum and maximum bounds to soften the impact of volatility and keep TFT’s purchasing power (on TF grid) stable  
• This ensures predictable, stable billing regardless of market extreme swings.

Now, when each rate is used:
 • Market rate -> for trading, valuations, user buy/sell decisions.
 • Billing rate -> for resource costs, contract billing, workload cost estimations.


> And if not used everywhere, wouldn't this create another inconsistency if users noticed the different rates?

[...]

PR #4430, opened 2025-11-17
Implement getTFTBillingRate

zaelgohary:
- Implement getTFTBillingRate n tfchain client
- Replace get tft price w getTFTBillingRate in grid client & playground

https://github.com/threefoldtech/tfgrid-sdk-ts/issues/4340

### DIFF
--- a/packages/grid_client/src/clients/tf-grid/contracts.ts
+++ b/packages/grid_client/src/clients/tf-grid/contracts.ts
@@ -387,7 +387,7 @@ class TFContracts extends Contracts {
    */
   private async convertToTFT(USD: Decimal) {
     try {
-      const tftPrice = (await this.client.tftPrice.get()) ?? 0;
+      const tftPrice = (await this.client.tftPrice.getTFTBillingRate()) ?? 0;
       const tft = new currency(tftPrice, 15).convertUSDtoTFT({ amount: USD.toNumber() });
       return new Decimal(tft);
     } catch (error) {
--- a/packages/grid_client/src/clients/tf-grid/contracts.ts
+++ b/packages/grid_client/src/clients/tf-grid/contracts.ts
@@ -20,6 +20,7 @@ import { Decimal } from "decimal.js";
 import { bytesToGB, formatErrorMessage } from "../../helpers";
 import { calculator, ContractStates, currency } from "../../modules";
 import { Graphql } from "../graphql/client";
+import { TFTPrice } from "./tftPrice";
 
 export type DiscountLevel = "None" | "Default" | "Bronze" | "Silver" | "Gold";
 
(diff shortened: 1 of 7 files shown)


packages/tfchain_client/src/tft_price.ts
```diff
@@ -11,6 +11,36 @@ class QueryTFTPrice {
...
+  async getTFTBillingRate(): Promise<number> {
+    const [minRes, maxRes, avgRes] = await Promise.all([
+      this.client.api.query.tftPriceModule.minTftPrice(),
+      this.client.api.query.tftPriceModule.maxTftPrice(),
+      this.client.api.query.tftPriceModule.averageTftPrice(),
+    ]);
+
+    const minPrice = (minRes.toPrimitive() as number) ?? 0;
+    const maxPrice = (maxRes.toPrimitive() as number) ?? 0;
+    const averagePrice = (avgRes.toPrimitive() as number) ?? 0;
```
sameh-farouk: I’m not deep into TypeScript, but this raises a question: is there a reason we’re masking the failure rather than throwing?
I think continuing with a missing or invalid value is risky, as it will produce incorrect billing while giving no indication that something went wrong, and it completely obscures the real issue, making debugging much harder.

packages/grid_client/src/clients/tf-grid/contracts.ts
```diff
@@ -387,7 +387,7 @@ class TFContracts extends Contracts {
    */
   private async convertToTFT(USD: Decimal) {
     try {
-      const tftPrice = (await this.client.tftPrice.get()) ?? 0;
+      const tftPrice = (await this.client.tftPrice.getTFTBillingRate()) ?? 0;
```
sameh-farouk: The same error masking happens here

packages/grid_client/src/clients/tf-grid/tftPrice.ts
```diff
@@ -5,6 +5,11 @@ class TFTPrice extends QueryTFTPrice {
     const priceInMili = await super.get();
     return priceInMili / 1000;
   }
+
+  async getTFTBillingRate(): Promise<number> {
```
sameh-farouk: The same name `getTFTBillingRate()`  is used at the tfchain_client and tfgrid_client, but they return different units. In the call sites, some divide by 1000 after getting the value, some don't, based on the place/client where they import the method. For example, Callers using the grid client should not divide again.

I would suggest changing the tfgrid client method name to `getTFTBillingRateUSD()` to avoid potential confusion about the unit returned at call sites.  Optionally, the tfchain client method name can also reflect the unit (`getTFTBillingRateMUSD`)

packages/tfchain_client/src/tft_price.ts
```diff
@@ -11,6 +11,46 @@ class QueryTFTPrice {
     const res = await this.client.api.query.tftPriceModule.tftPrice();
     return res.toPrimitive() as number;
   }
+
+  @checkConnection
+  async getTFTBillingRateMUSD(): Promise<number> {
```
AhmedHanafy725, line 16: The uppercase M sounds like it's millions not milli
```suggestion
  async getTFTBillingRatemUSD(): Promise<number> {

```

zaelgohary: The capital 'M' in MUSD was used strictly to adhere to camel case convention. Within this repo, we exclusively deal with standard USD and milli-USD; we do not use the Mega prefix anywhere. Therefore, the limited scope of our units ensures there is no ambiguity for the reader.

Issue #3458, opened 2024-09-25
Reduce unused JS from Dashboard


zaelgohary:
### Which package/s are you suggesting this feature for?

Dashboard

### Is your feature request related to a problem? Please describe

Related: https://pagespeed.web.dev/analysis/https-staging-dashboard-dev-grid-tf/6eb48ptgon?form_factor=desktop

### Describe the solution you'd like

-

PR #4432, opened 2025-11-19
Optimize bundle size: lazy load heavy dependencies and improve tree-shaking

zaelgohary:
### Changes

- Lazy load jspdf and jspdf-autotable (load only when PDF download requested)
- Lazy load Sentry (load only when telemetry enabled)
- Optimize lodash imports to enable tree-shaking (use specific paths)
- Change static route imports to dynamic imports for better code splitting
- Configure Vite optimizeDeps to prevent HMR reload issues with lodash sub-modules
- Remove duplicate ProjectName.Openwebui entry in delete_deployment utils
- Make internal utility functions private where appropriate

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/3458

### DIFF
--- a/packages/playground/src/utils/delete_deployment.ts
+++ b/packages/playground/src/utils/delete_deployment.ts
@@ -101,7 +101,6 @@ export function solutionHasGateway(projectName: ProjectName) {
     ProjectName.Nostr,
     ProjectName.StaticWebsite,
     ProjectName.NodePilot,
-    ProjectName.Openwebui,
   ];
 
   for (const solution of solutions) {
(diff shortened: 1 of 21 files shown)


packages/playground/src/utils/delete_deployment.ts
```diff
@@ -101,7 +101,6 @@ export function solutionHasGateway(projectName: ProjectName) {
     ProjectName.Nostr,
     ProjectName.StaticWebsite,
     ProjectName.NodePilot,
-    ProjectName.Openwebui,
```
AhmedHanafy725, line 104: Why do we delete that?

zaelgohary: duplicate.

AhmedHanafy725 (approved):

Issue #4320, opened 2025-07-31
🐞 [Bug]: Error TS2339: Property 'GATEWAY' does not exist


samaradel:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### which package/s did you face the problem with?

grid_client

### What happened?

Grid client gateway test fails

### Steps To Reproduce

- Go grid client package
- Run gateway test

### which network/s did you face the problem on?

Dev

### version

02bfe9463999ab7b48ac15a4380445bbc2bd117e

### Relevant screenshots/screen records

<img width="1875" height="440" alt="Image" src="https://github.com/user-attachments/assets/f84f214d-c8cb-4649-955c-68d7a3e3f53f" />

### Relevant log output

```console
N/A
```

PR #4457, opened 2025-11-24
Fix gateway test

zaelgohary:
- Fix flist & entrypoint value in gateway test

https://github.com/threefoldtech/tfgrid-sdk-ts/issues/4320


AhmedHanafy725 (commented):
Did you try this fix? I believe the test will fail

AhmedHanafy725 (approved):

PR #4464, opened 2025-11-26
Add BTC Node release date

zaelgohary:
https://github.com/threefoldtech/tfgrid-sdk-ts/issues/4392


AhmedHanafy725 (approved):
Next, we should set the date to the release date right before making the release tag

Issue #4451, opened 2025-11-24
Console Interception Overhead


zaelgohary:
#### Issue: Logger Intercepts All Console Calls
**Location:** `components/logger.vue:143`

**Problem:**
- `LoggerInterceptor` wraps all console methods
- Every `console.log/error/warn` call triggers IndexedDB write
- Serialization happens synchronously
- No filtering of noisy logs (e.g., Vite HMR logs)

**Impact:**
- Overhead on every console call
- Blocks main thread during IndexedDB writes
- Performance degradation in development

**Fix:**
```typescript
// Use async batching for log writes
const logQueue: LI[] = [];
let flushTimeout: ReturnType<typeof setTimeout> | null = null;

async function interceptMessage(instance: LI) {
  logQueue.push(instance);
  
  // Batch writes every 500ms or when queue reaches 50 items
  if (logQueue.length >= 50 || !flushTimeout) {
    flushTimeout = setTimeout(async () => {
      await flushLogQueue();
      flushTimeout = null;
    }, 500);
  }
}

async function flushLogQueue() {
  const batch = logQueue.splice(0, 50);
  await Promise.all(batch.map(log => logsDBClient.write(log)));
}
```

Issue #4448, opened 2025-11-24
IndexedDB Logger Without Size Limits


zaelgohary:
#### Issue: Console Logger Storage Growth
**Location:** `components/logger.vue:143-250`

**Problem:**
- Logger intercepts ALL console calls (`console.log`, `console.error`, etc.)
- Stores every log message in IndexedDB without size limits
- No cleanup/rotation of old logs
- Serialization overhead on every log call
- `readAll()` loads ALL logs into memory for download

**Impact:**
- IndexedDB storage fills up over time
- Performance degradation as log count grows
- Memory issues when downloading logs
- Overhead on every console call

**Fix:**
```typescript
// Add size limits and rotation
const MAX_LOGS = 10000;
const MAX_DB_SIZE = 50 * 1024 * 1024; // 50MB

async function interceptMessage(instance: LI) {
  // ... existing code ...
  
  const count = await logsDBClient.count();
  if (count > MAX_LOGS) {
    // Delete oldest logs
    const toDelete = count - MAX_LOGS + 1000; // Keep some buffer
    await logsDBClient.deleteRange(0, toDelete);
  }
  
  // Check DB size periodically
  if (count % 1000 === 0) {
    const size = await logsDBClient.getSize();
    if (size > MAX_DB_SIZE) {
      await logsDBClient.deleteRange(0, Math.floor(count * 0.3)); // Delete 30% oldest
    }
  }
}
```

PR #4488, opened 2025-12-02
Optimize dashboard logger performance and console interception

zaelgohary:
- Batch async log writes (50 items, 500ms delay) to eliminate blocking
- Filter dev tooling noise (Vite/HMR/Vue warnings) before queuing
- Disable interception on DB failure to eliminate overhead
- Cap in-memory logs (2000) and IndexedDB storage (10000 with rotation)
- Filter Vue warnings to show only user-relevant logs
- Prevent recursive interception and ensure cleanup on unmount

https://github.com/threefoldtech/tfgrid-sdk-ts/issues/4451
https://github.com/threefoldtech/tfgrid-sdk-ts/issues/4448


packages/playground/src/clients/indexedDB/index.ts
```diff
@@ -140,6 +140,32 @@ export class IndexedDBClient {
...
+    await this._lock.acquireAsync();
+
+    return new Promise((res, rej) => {
+      const store = this._createStore();
+      const range = IDBKeyRange.bound(startId, endId);
+      const query = store.openCursor(range);
+
+      query.onsuccess = () => {
+        const cursor = query.result;
+        if (cursor) {
```
AhmedHanafy725: I tried to see if it's actually deleting the logs but it reached here and the cursor was null, so nothing was deleted.

Image

zaelgohary: after the fix: 
image

